Issue #49, opened 2026-05-12
/mcp accepts invalid Bearer token as HTTP 200 unauthenticated initialize


markmhendrickson:
## Summary
`POST /mcp` treats the mere presence of an `Authorization: Bearer ...` header as enough to avoid a 401 challenge before the token is validated. In encryption-off/shared-bearer mode, a key-derived `neotoma auth mcp-token` value fails correctly on `POST /entities/query` with `AUTH_INVALID`, but `POST /mcp` returns HTTP 200 with an unauthenticated initialize response.

## Why it matters
This makes a mode-mismatched or invalid Bearer token appear to work through the MCP route, while REST rejects it. That matches a user report from the v0.11.1/GHSA-5cvp-p7p4-mcx9 follow-up and can send operators down the wrong path when diagnosing auth configuration.

## Reproduction
1. Run server with `NEOTOMA_ENCRYPTION_ENABLED=false` and `NEOTOMA_BEARER_TOKEN=shared-secret-token`.
2. Send `POST /entities/query` with `Authorization: Bearer <64-char key-derived-token-shape>`.
3. Observe `401 AUTH_INVALID`.
4. Send MCP initialize to `POST /mcp` with the same Bearer header.
5. Prior behavior: HTTP 200 with unauthenticated initialize payload.

## Local fix in progress
A local patch now validates non-shared Bearer tokens in `src/actions.ts` by resolving them via MCP OAuth before the request reaches the MCP transport. Invalid tokens return `401 invalid_token`. Regression coverage was added in `tests/integration/mcp_invalid_bearer_auth.test.ts`.

## Verification
- `npx vitest run tests/integration/mcp_invalid_bearer_auth.test.ts` passed.
- `npx vitest run tests/integration/tunnel_auth.test.ts` passed.

markmhendrickson:
Implemented local patch and regression coverage.

Changed files:
- `src/actions.ts`: validates non-shared Bearer tokens on `/mcp` by resolving MCP OAuth before handing off to transport; invalid tokens now return `401 invalid_token`.
- `tests/integration/mcp_invalid_bearer_auth.test.ts`: covers encryption-off/shared-bearer mode with a key-derived-token-shaped invalid Bearer value and asserts JSON-RPC 401.

Verification:
- `npx vitest run tests/integration/mcp_invalid_bearer_auth.test.ts` passed.
- `npx vitest run tests/integration/tunnel_auth.test.ts` passed.
[...]

Issue #78, opened 2026-05-12
Inspector SPA missing from npm tarball in v0.11.0/v0.11.1 (fixed in v0.12.1 packaging fail-closed)


hoffjero:
## Summary

Public post-mortem for a packaging regression that affected v0.11.0 and v0.11.1: the bundled Inspector SPA (`dist/inspector/index.html` + assets) was not included in the npm tarball when the `inspector/` submodule was not initialized at publish time. `scripts/build_inspector.js` silently skipped the build step instead of failing. Fixed in v0.12.1 by making prepublish exit non-zero when the submodule is missing.

Filing as a post-mortem so a future operator hitting `404 Not Found` on `/inspector` (or `null` for `server_info.inspector_url`) on v0.11.x can find this issue and the fix path.

## Symptom (observed by us on v0.11.0 → v0.11.1)

- `GET /inspector` via tunnel → `404 Not Found` (not `308 Permanent Redirect`)
- `GET /server-info` returned `"inspector_url": null` instead of a path
- `find $(npm prefix -g)/lib/node_modules/neotoma/dist/inspector` → no such directory
- The npm tarball at `https://www.npmjs.com/package/neotoma/v/0.11.1` did not contain `dist/inspector/`

## Reproduction (against pre-fix versions)

1. Fresh clone of `neotoma` at `v0.11.0` or `v0.11.1`
2. **Skip** `git submodule update --init inspector`
3. `npm run pack:local` or `npm publish` (dry-run)
4. Inspect tarball: no `dist/inspector/` directory
5. Install the tarball globally: `GET /inspector` returns `404`

## Diagnosis trail (private, summarized)

Our internal trail is in our local plan. Key observations that led us to flag this to the maintainer:

- After upgrading mac-mini from v0.10.1 → v0.11.0 → v0.11.1, the Inspector `/inspector` route consistently 404'd via our autossh tunnel
- `server-info`'s `inspector_url` field was `null`
- Confirmed `dist/inspector/index.html` was not present in the installed npm package directory
- Same install procedure on v0.10.1 had a working Inspector → narrowed to a packaging regression between 0.10.x and 0.11.0

## Fix (already shipped in v0.12.1)

Per the v0.12.1 release notes:

[...]

PR #79, opened 2026-05-13
Fix self-hosted auth for key-file-only deployments and Cloudflare Tunnel topology

### LINKED ISSUE #80: Bearer token auth fails when NEOTOMA_KEY_FILE_PATH set but NEOTOMA_ENCRYPTION_ENABLED absent
## Summary

When `NEOTOMA_KEY_FILE_PATH` is configured but `NEOTOMA_ENCRYPTION_ENABLED` is not explicitly set to `true`, the server silently rejects bearer tokens generated by `neotoma auth mcp-token`. The CLI derives the token directly from the key file without consulting the encryption flag, so the token is valid but the server never attempts to validate it.

## Root cause

`getMcpAuthToken()` in `src/crypto/mcp_auth_token.ts` had an early-return guard:

```ts
if (!config.encryption.enabled) return null;
```

`NEOTOMA_ENCRYPTION_ENABLED` defaults to `false`, so any deployment that sets a key file for auth purposes — without also enabling full encryption — gets `null` from the server's token resolver. The bearer check in both the REST and MCP middleware paths was also nested inside `if (config.encryption.enabled)` blocks, compounding the problem.

## Symptoms

- `neotoma auth mcp-token` produces a token
- MCP client configured with that token receives `401 AUTH_INVALID` on REST routes
- `POST /mcp` may return HTTP 200 with an unauthenticated session

## Environment

Reported by @hoffjero in the context of a self-hosted deployment with a key file configured.

## Fix

Tracked in PR #79:
- Remove the `encryption.enabled` guard from `getMcpAuthToken()`
- Move bearer token validation outside the `encryption.enabled` block in both REST and MCP middleware

### LINKED ISSUE #81: isLocalRequest rejects Cloudflare Tunnel requests despite loopback socket
## Summary

When Cloudflare Tunnel (`cloudflared`) runs as a sidecar on the same host as the Neotoma server, MCP and REST requests arrive on a loopback socket but carry `x-forwarded-for` headers injected by cloudflared containing the original client's non-loopback IP. `isLocalRequest()` treats any XFF entry from a non-loopback IP as disqualifying, so the request is denied as non-local even though it physically arrived via loopback.

## Root cause

`isLocalRequest()` in `src/actions.ts` checks `x-forwarded-for` strictly: if any XFF value is not a loopback address, the request is rejected regardless of what socket it arrived on. There was no mechanism to declare that certain IPs in the XFF chain are trusted local proxies rather than remote clients.

## Symptoms

- Server is accessible through Cloudflare Tunnel
- Bearer token is correctly configured (see #80)
- Requests through the tunnel return auth failures despite the token being valid
- Direct loopback connections (bypassing the tunnel) succeed

## Environment

Reported by @hoffjero for a self-hosted deployment behind Cloudflare Tunnel.

## Fix

Tracked in PR #79: new `NEOTOMA_TRUSTED_PROXY_IPS` environment variable accepts comma-separated IPs and IPv4 CIDRs. `isLocalRequest()` now treats XFF entries that match a trusted proxy CIDR as non-disqualifying. Cloudflare Warp/tunnel addresses (`100.64.0.0/10`) can be allowlisted.

```bash
NEOTOMA_TRUSTED_PROXY_IPS=100.64.0.0/10
```

### LINKED ISSUE #82: GitHub issue mirror silently aborts on 422 when repo labels are missing
## Summary

When `createIssue` sends labels to the GitHub Issues API and one or more of those labels do not exist on the target repo, GitHub returns `422 Unprocessable Entity` and the entire issue creation fails. No issue is created, no warning is surfaced, and the error is not actionable.

## Root cause

`createIssue` in `src/services/issues/github_client.ts` passed labels directly to the API without verifying they exist on the repo. A single unknown label — including the canonical `neotoma` label added automatically by `mergeNeotomaToolingIssueLabels` — is enough to abort the entire request with a 422.

## Symptoms

- `submit_issue` or any path that mirrors to GitHub appears to fail
- No GitHub issue is created
- The error is a raw 422 with no indication that the label is the cause

## Environment

Reported by @hoffjero for a repo where the `neotoma` label had not been created in GitHub Settings → Labels.

## Fix

Tracked in PR #79: `createIssue` now catches 422, fetches the repo's actual label list via `listRepoLabelNames`, drops unknown labels with a structured stderr warning, and retries with only valid labels. A new `GitHubApiError` class carries the HTTP status to distinguish label-422s from other errors.

Example warning emitted to stderr:

```
[neotoma] GitHub issue mirror: dropping unknown label(s) [neotoma] from repo owner/repo — add them in GitHub Settings → Labels to include them.
```

### LINKED ISSUE #83: add_issue_message AUTH_REQUIRED gives no actionable guidance on required credentials
## Summary

When a remote agent calls `add_issue_message` without a valid credential, the tool returns `AUTH_REQUIRED` but the error and tool description give no indication of what credential is needed or how to obtain it. Operators have to guess that `guest_access_token` (returned by `submit_issue`) is the designed auth path, or that an `agent_grant` configured via Inspector is the alternative.

## Root cause

The `add_issue_message` tool description in `src/tool_definitions.ts` did not document the two supported remote auth paths, leaving operators with no actionable next step when the error fires.

## Symptoms

- Remote agent calls `add_issue_message`
- Receives `AUTH_REQUIRED`
- No indication in the error or docs of what token to supply or where to get it

## Environment

Reported by @hoffjero for a remote MCP agent attempting to append messages to an existing issue.

## Fix

Tracked in PR #79: tool description updated to explicitly name both auth paths:
- `guest_access_token` returned by `submit_issue` — for agents that filed the issue
- `agent_grant` configured in Inspector → Agents → Grants — for persistent agent access

markmhendrickson:
## Summary

Four auth and integration bugs affecting self-hosted deployments behind Cloudflare Tunnel, each at a distinct layer of the stack.

**#80 — Key-file bearer token rejected by server** (`src/crypto/mcp_auth_token.ts`, `src/actions.ts`): The CLI derives a bearer token from the key file regardless of whether full encryption is enabled. The server's token resolver had an early-return guard that returned null whenever `NEOTOMA_ENCRYPTION_ENABLED` was not explicitly `true`, so the server never attempted to validate a token it should have accepted. Fix: remove the guard; return a token whenever a key source is configured.

**#81 — Cloudflare Tunnel XFF headers disqualifying loopback connections** (`src/actions.ts`): Requests through a tunnel sidecar arrive on a loopback socket but carry `x-forwarded-for` headers with the client's real IP. `isLocalRequest()` rejected any XFF entry that wasn't itself a loopback address, with no way to declare that a proxy IP is local infrastructure. Fix: new `NEOTOMA_TRUSTED_PROXY_IPS` env var (comma-separated IPs/CIDRs); XFF entries matching a trusted CIDR no longer disqualify the request.

**#82 — GitHub issue mirror silently aborts on unknown labels** (`src/services/issues/github_client.ts`): The canonical `neotoma` label is appended to every mirrored issue automatically. If that label (or any other) doesn't exist on the target repo, GitHub returns 422 and the entire issue creation fails with no warning. Fix: catch 422, fetch the repo's actual label list, drop unknowns with a stderr warning, retry with valid labels only.

**#83 — `add_issue_message` AUTH_REQUIRED gives no next step** (`src/tool_definitions.ts`): The error fired but the tool description named neither of the two supported credential paths. Fix: document `guest_access_token` (from `submit_issue`) and `agent_grant` (Inspector → Agents → Grants) explicitly in the tool description.

---

## Changes

- **`getMcpAuthToken()` now returns a token whenever a key source is configured**, regardless of `NEOTOMA_ENCRYPTION_ENABLED`. Closes #80.
- **Bearer token validation moved outside the `encryption.enabled` block** in both REST and MCP middleware. Related to #49 (same surface, inverse failure mode). Closes #80.
- **`NEOTOMA_TRUSTED_PROXY_IPS`** — new env var for declaring trusted proxy CIDRs so tunnel sidecars don't disqualify loopback connections. Closes #81.
- **GitHub label 422 retry** — `createIssue` catches 422, fetches repo labels, drops unknowns with a warning, retries. Closes #82.
- **`add_issue_message` tool description** updated with both remote auth paths. Closes #83.

## Related issues

- #80 — Bearer token rejected when key file set without encryption flag
- #81 — Cloudflare Tunnel XFF blocks loopback auth
- #82 — GitHub mirror 422 on missing labels
- #83 — `add_issue_message` AUTH_REQUIRED uninformative
- #49 — MCP bearer token middleware (same surface, inverse failure mode)

## Test plan

These steps are ordered to match the failure sequence: bearer auth blocks everything else, so verify that first before testing tunnel or issue tooling.

### 0. Install the PR build

```bash
npm install -g neotoma@356ee2c10
# or if running from source:
git fetch origin test/current-branch-build && git checkout test/current-branch-build
npm ci && npm run build
```

Restart your Neotoma server after installing.

---

### 1. Bearer token auth (key file without `NEOTOMA_ENCRYPTION_ENABLED`) — fixes #80

**Prerequisites:** `NEOTOMA_KEY_FILE_PATH` set to your key file. `NEOTOMA_ENCRYPTION_ENABLED` absent or set to `false`.

```bash
# Generate the token the same way your MCP client would
neotoma auth mcp-token

# Test it against the server directly
curl -s -o /dev/null -w "%{http_code}" \
  -H "Authorization: Bearer <token>" \
  http://localhost:<port>/entities/query \
  -X POST -H "Content-Type: application/json" \
  -d '{"entity_type":"contact","limit":1}'
```

**Expected:** `200` (previously `401 AUTH_INVALID`).

---

### 2. Cloudflare Tunnel XFF bypass — fixes #81

**Prerequisites:** Cloudflare Tunnel running as a sidecar on the same host.

**Step 2a — discover the XFF IP your tunnel actually injects.**

The IP cloudflared puts in `x-forwarded-for` depends on its forwarding mode and is not always `100.64.x.x`:
- Warp / CGNAT mode: typically a `100.64.x.x` address
- Passthrough mode (common with Cloudflare Access / Vercel egress): the upstream client IP

Start the server with the PR build (no `NEOTOMA_TRUSTED_PROXY_IPS` set yet) and make one MCP request through the tunnel. Neotoma will print to stderr:

```
[neotoma] isLocalRequest: loopback socket rejected because XFF contains untrusted IP(s): 13.219.225.56. Set NEOTOMA_TRUSTED_PROXY_IPS to trust these addresses.
```

Note the IP(s) printed.

**Step 2b — configure `NEOTOMA_TRUSTED_PROXY_IPS`.**

Add to your server environment, using the IP or CIDR from step 2a:

```bash
# Example for Warp / CGNAT mode
NEOTOMA_TRUSTED_PROXY_IPS=100.64.0.0/10

# Example for passthrough mode (use your actual egress IP or CIDR)
NEOTOMA_TRUSTED_PROXY_IPS=13.219.225.56
```

Comma-separated values and IPv4 CIDRs are both supported.

**Note:** `NEOTOMA_TRUSTED_PROXY_IPS` and `NEOTOMA_TRUST_PROD_LOOPBACK=1` are independent and can coexist. If you are already using `NEOTOMA_TRUST_PROD_LOOPBACK=1` and it works for your setup, there is no obligation to migrate — the PR does not deprecate it.

Restart the server, then connect through the tunnel with the bearer token from step 1.

**Expected:** MCP initialize succeeds through the tunnel. Previously returned a local-auth rejection.

---

### 3. GitHub issue label 422 retry — fixes #82

**Prerequisites:** A GitHub repo configured as your issue mirror that does **not** have the `neotoma` label created yet.

Trigger a mirror write via `submit_issue` or `add_issue_message`.

**Expected:**
- The issue is created successfully with whatever labels do exist.
- Server stderr contains:
  ```
  [neotoma] GitHub issue mirror: dropping unknown label(s) [neotoma] from repo owner/repo — add them in GitHub Settings → Labels to include them.
  ```

Previously this threw 422 and created nothing.

---

### 4. `add_issue_message` auth error message — fixes #83

From an MCP client without a `guest_access_token` or `agent_grant`, call `add_issue_message` on a remote issue.

**Expected:** The error now names both auth paths (`guest_access_token` from `submit_issue`, or an `agent_grant` via Inspector → Agents → Grants). Previously the error gave no indication of what to supply.

### DIFF
--- a/.github/workflows/ci_test_lanes.yml
+++ b/.github/workflows/ci_test_lanes.yml
@@ -145,7 +145,10 @@ jobs:
             BASE_REF="origin/main"
           fi
           git fetch origin "${BASE_REF#origin/}" --depth=200 || true
-          npm run security:classify-diff -- --base "$BASE_REF" --head HEAD --json | tee /tmp/security_classify.json
+          npm run security:classify-diff -- --base "$BASE_REF" --head HEAD --json > /tmp/security_classify_raw.txt 2>&1
+          # Strip npm script header lines before parsing JSON
+          grep -v "^>" /tmp/security_classify_raw.txt | grep -v "^npm" > /tmp/security_classify.json || true
+          cat /tmp/security_classify.json
           SENSITIVE=$(node -e "console.log(require('/tmp/security_classify.json').sensitive)")
           echo "SECURITY_SENSITIVE=$SENSITIVE" >> "$GITHUB_ENV"
 
--- a/docs/developer/tunnels.md
+++ b/docs/developer/tunnels.md
@@ -57,6 +57,28 @@ Do not use a tunnel when:
 
 **Summary:** Tunnel traffic must always authenticate. OAuth is key-gated; non-key users should use `NEOTOMA_BEARER_TOKEN`.
 
+**Caller authentication for cloud-fronted topologies (Vercel, CDN edge, Cloudflare Access):**
+
+When a cloud service (Vercel, a CDN, Cloudflare Access) sits in front of cloudflared, the request that reaches Neotoma carries `x-forwarded-for` with the cloud egress IP — not a loopback address. This matters because:
+
+- `NEOTOMA_TRUST_PROD_LOOPBACK=1` only applies when XFF is absent or all-loopback addresses. A non-loopback XFF entry disqualifies loopback-trust even with that variable set.
+- Callers in this topology — a Vercel app hitting `api.yourdomain.com`, a CDN-fronted webhook, etc. — will receive `401 AUTH_REQUIRED` unless they send a `Bearer` token.
+
+**The fix:** every caller behind a cloud proxy must send `Authorization: Bearer <token>`:
+
+1. Generate the token on the host machine:
+   ```bash
+   neotoma auth mcp-token   # requires NEOTOMA_KEY_FILE_PATH or NEOTOMA_MNEMONIC
+   ```
+2. Set `NEOTOMA_BEARER_TOKEN` in the caller's environment (e.g. `vercel env add NEOTOMA_BEARER_TOKEN production`) with the output from step 1.
+3. Have the caller include `Authorization: Bearer ${NEOTOMA_BEARER_TOKEN}` on every Neotoma request.
+
+To verify which XFF IP your tunnel topology actually injects, start the server and make one request through the tunnel. Neotoma logs to stderr:
+```
+[neotoma] isLocalRequest: loopback socket rejected because XFF contains untrusted IP(s): 13.219.225.56. Set NEOTOMA_TRUSTED_PROXY_IPS to trust these addresses.
+```
+If you want loopback-trust (rather than bearer auth) for these callers, set `NEOTOMA_TRUSTED_PROXY_IPS` to that IP or CIDR. `NEOTOMA_TRUST_PROD_LOOPBACK=1` and `NEOTOMA_TRUSTED_PROXY_IPS` are independent and can coexist.
+
 **Local OAuth over a public tunnel:** With encryption off, OAuth still uses a built-in dev account **after** key-auth preflight succeeds. When the server is reached **via a tunnel** (non-local Host), it requires **explicit approval** (an "Approve this connection" page) before completing OAuth, and it only accepts **allowlisted redirect URIs** (e.g. `cursor://`, `http://localhost`, `http://127.0.0.1`) so the authorization code cannot be sent to an arbitrary third-party site. If users cannot complete key-authenticated OAuth, use bearer token access:
 
 - **`NEOTOMA_BEARER_TOKEN`** in `.env` and send it as `Authorization: Bearer <token>` from the client (no OAuth; good for scripts or single-user).
--- a/docs/releases/in_progress/v0.12.1/github_release_supplement.md
+++ b/docs/releases/in_progress/v0.12.1/github_release_supplement.md
@@ -83,6 +83,11 @@ Headline:
 - Operator-facing hardening knob documentation does not change runtime behavior; it makes already-shipped knobs findable from the security landing page so hosted-mode operators are less likely to leave them at single-tenant defaults.
 - No new advisory under `docs/security/advisories/`. No CVE.
 
+**Amendment — PR #79 (self-hosted auth + Cloudflare Tunnel, commit `356ee2c10`):**
+
+- `src/crypto/mcp_auth_token.ts` — `getMcpAuthToken()` no longer gates token derivation on `NEOTOMA_ENCRYPTION_ENABLED`. The change only widens the set of accepted valid tokens; constant-time comparison (`safeCompareTokens`) is unchanged.
+- `src/actions.ts` — bearer token validation moved above the `encryption.enabled` branch in both REST and MCP middleware; new `NEOTOMA_TRUSTED_PROXY_IPS` opt-in env var for declaring trusted proxy CIDRs. Both changes are off-by-default for existing deployments. Full gate evidence in `security_review.md` § Amendment — PR #79. `test:security:auth-matrix` → 16 passed, 1 skipped.
+
 ## Deferred follow-up
 
 - **Inspector UI for plans on issues** (carried from v0.12.0): the `Plans` tab on `IssueDetailPage`, per-message plan chips, and the related contract test remain deferred. The v0.12.0 supplement (now archived under `docs/releases/completed/v0.12.0/`) describes the scope.
--- a/docs/releases/in_progress/v0.12.1/security_review.md
+++ b/docs/releases/in_progress/v0.12.1/security_review.md
@@ -107,6 +107,61 @@ These are purely documentation; they describe runtime behaviors that already shi
 
 **Verdict:** approved.
 
+## Amendment — PR #79 (self-hosted auth + Cloudflare Tunnel fixes)
+
+PR #79 (`test/current-branch-build` → `main`, commit `356ee2c10`) touches `src/actions.ts` and `src/crypto/mcp_auth_token.ts` — flagged `sensitive=true` by the G1 classifier (`auth-middleware`).
+
+### G1 — `security:classify-diff`
+
+`npm run -s security:classify-diff -- --base origin/main --head test/current-branch-build --json` → `sensitive=true`.
+
+Flagged: `auth-middleware: src/actions.ts`.
+
+### G2 — `security:lint`
+
+`npm run -s security:lint` → **0 errors, 110 warnings across 290 files**. No new errors or suppression directives.
+
+### G3 — `security:manifest:check` and `test:security:auth-matrix`
+
+- `npm run -s security:manifest:check` → `protected_routes_manifest.json: in sync with openapi.yaml (106 routes)`.
+- `npm run -s test:security:auth-matrix` → 16 passed, 1 skipped, 0 failed.
+
+### G4 — Manual review
+
+#### G4.5 Bearer token auth decoupled from `NEOTOMA_ENCRYPTION_ENABLED` (`src/crypto/mcp_auth_token.ts`, `src/actions.ts`)
+
+**What changed:**
+
+`getMcpAuthToken()` previously returned `null` whenever `config.encryption.enabled` was `false`, regardless of whether a key source (`NEOTOMA_KEY_FILE_PATH` or `NEOTOMA_MNEMONIC`) was configured. The early-return guard is removed. The function now returns a derived token whenever a key source is present.
+
+In `src/actions.ts`, the `bearer_mcp_token` validation block in both the REST middleware and the MCP route handler was nested inside `if (config.encryption.enabled)`. Both blocks are now moved above that condition so they execute whenever `getMcpAuthToken()` returns a non-null value.
+
+**Security analysis:**
+
+The change only widens the set of *accepted* valid tokens — it does not relax validation. The token comparison continues to use `safeCompareTokens` (constant-time). An operator must have `NEOTOMA_KEY_FILE_PATH` or `NEOTOMA_MNEMONIC` set for the new path to activate; absent a key source, `getMcpAuthToken()` still returns `null` and the path is inert.
+
+The previous behavior was a misconfiguration trap: `neotoma auth mcp-token` derived a token that the server silently rejected. Fixing this closes a user-facing gap while leaving the authentication model unchanged.
+
+**Risk introduced:** none. The validation logic is identical; only the gating condition changed.
+
+**Verdict:** approved.
+
+#### G4.6 `NEOTOMA_TRUSTED_PROXY_IPS` (`src/actions.ts`)
+
+**What changed:**
+
+New `parseTrustedProxyIPs` and `isTrustedProxyIP` functions read `NEOTOMA_TRUSTED_PROXY_IPS` (comma-separated IPs/CIDRs). `isLocalRequest()` now accepts a loopback-socket request if every XFF entry is either a loopback address or matches a trusted proxy CIDR.
+
+**Security analysis:**
+
+The feature is opt-in and off-by-default — `NEOTOMA_TRUSTED_PROXY_IPS` is unset in all existing deployments, leaving `isLocalRequest()` behavior unchanged.
+
+When set, the operator explicitly declares which proxy IPs are trusted infrastructure. The CIDR matching uses bitwise IPv4 arithmetic with no external dependency and no dynamic code evaluation. The function does not affect any other auth path; it only influences whether `isLocalRequest()` returns `true`.
+
+Misconfiguration risk: an operator who sets `NEOTOMA_TRUSTED_PROXY_IPS=0.0.0.0/0` would trust all XFF IPs, effectively allowing any request on a loopback socket to be treated as local. This is an operator error analogous to misconfiguring `NEOTOMA_TRUST_PROD_LOOPBACK=1`. It is documented as operator responsibility and consistent with how other trust-extension env vars work in this codebase.
+
+**Verdict:** approved.
+
 ## Suggested negative tests (forward-looking)
 
 - Revoked guest token end-to-end (see G4.1).
--- a/docs/security/advisories/2026-05-11-inspector-auth-bypass.md
+++ b/docs/security/advisories/2026-05-11-inspector-auth-bypass.md
@@ -105,6 +105,8 @@ Shipped in [v0.11.1](https://github.com/markmhendrickson/neotoma/releases/tag/v0
 
 If your deployment runs Neotoma behind a single-host reverse proxy and the bearer-fronted UX is too friction-heavy, you may opt back into trusting the loopback chain with `NEOTOMA_TRUST_PROD_LOOPBACK=1`. Document this choice in your runbook; the gates record any new reference to the env var as security-sensitive.
 
+**Operational note — cloud-fronted topologies (added v0.12.1):** `NEOTOMA_TRUST_PROD_LOOPBACK=1` is nullified when callers arrive through a cloud service (Vercel, CDN edge, Cloudflare Access) that injects a non-loopback `x-forwarded-for` entry. In those topologies, loopback-trust never activates regardless of this variable. The symptom is `401 AUTH_REQUIRED` on requests that work fine from CLI or stdio MCP clients (which carry no XFF). The remedy is explicit bearer authentication from the cloud caller — see `docs/developer/tunnels.md` § "Caller authentication for cloud-fronted topologies". As of v0.12.1, Neotoma logs the untrusted XFF IP to stderr when it rejects a loopback-socket request, making this failure mode self-diagnosing.
+
 ## Detection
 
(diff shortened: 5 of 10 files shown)


hoffjero:
## Verification report — PR #79

**Setup:** macOS arm64 mac-mini, node@24, neotoma v0.12.1 dist swapped with PR-build `356ee2c10` (npm-global install was `dist/` → `dist.pre-pr79/`, replaced with locally built `dist/` from the PR commit, retained `inspector/` from the prior build). Pre-test DB backup retained on SSD. Pre-existing config: `NEOTOMA_KEY_FILE_PATH` set, `NEOTOMA_ENCRYPTION_ENABLED` unset, `NEOTOMA_TRUST_PROD_LOOPBACK=1` (v0.11.1 workaround).

### #80 — Bearer token with key file but no encryption flag ✓
Ran `neotoma auth mcp-token` → 64-char hex. Two MCP `initialize` calls:

- valid token → `HTTP 200`
- arbitrary 64-char hex → `HTTP 401`, `Invalid or expired Bearer token. Remove Authorization from mcp.json and click Connect to re-authenticate.`

Token validation now active outside `encryption.enabled`. Fix behaves as advertised.

### #81 — Cloudflare Tunnel XFF / `NEOTOMA_TRUSTED_PROXY_IPS` ✓ (code-level)
Did not migrate off `NEOTOMA_TRUST_PROD_LOOPBACK=1` for this verification — too disruptive to a live tunnel mid-run. Verified in compiled `dist/actions.js`: env-parsing, IPv4 exact + IPv4 CIDR matching, `::ffff:<v4>` mapped addresses, and the docstring with `100.64.0.0/10` example all present. Will migrate to the cleaner env after merge.

### #82 — GitHub mirror label 422 retry ✓ (code-level)
Did not runtime-test from this setup — `submit_issue` writes via the operator at `neotoma.markmhendrickson.com`, so the local PR-build's `github_client.js` patch isn't exercised by our normal flow. Verified in compiled `dist/services/issues/github_client.js`: new `GitHubApiError` class, 422-catch, `listRepoLabelNames()` retry path, stderr warning emit (\`[neotoma] GitHub issue mirror: dropping unknown label(s) [...] from repo X/Y — add them in GitHub Settings → Labels to include them.\`). Reproduction context: yesterday's failure when filing #78 with label \`packaging\` (does not exist on the repo).

### #83 — `add_issue_message` AUTH_REQUIRED uninformative ✓
MCP `tools/list` now returns the expanded description that names both auth paths:

- \`guest_access_token\` from \`submit_issue\` (or read from local issue snapshot)
- \`agent_grant\` configured by the operator (Inspector → Agents → Grants)

The error-path remediation language also lands: *\"If add_issue_message returns AUTH_REQUIRED, it means neither path is satisfied: check that guest_access_token from submit_issue was preserved on the local entity, or ask the operator to configure an agent_grant via Inspector → Agents → Grants.\"*

---

All four fixes look correct. Green from our side — go ahead and merge. Once tagged, I'll move from the manual dist-swap to the official npm release and drop the \`NEOTOMA_TRUST_PROD_LOOPBACK\` env in favor of \`NEOTOMA_TRUSTED_PROXY_IPS=100.64.0.0/10\`.

🤖 Posted by Claude on Jeroen's behalf

hoffjero:
## Migration follow-up — `NEOTOMA_TRUSTED_PROXY_IPS` doesn't fit every tunnel topology

Heads-up before merge: I tried to migrate our setup off `NEOTOMA_TRUST_PROD_LOOPBACK=1` to the cleaner `NEOTOMA_TRUSTED_PROXY_IPS=100.64.0.0/10` and discovered our XFF header doesn't match that CIDR.

**Our topology:** Vercel app → Cloudflare Access → cloudflared tunnel sidecar on mac-mini → loopback Neotoma on 3180.

**Observed XFF:** `13.219.225.56` (AWS / Vercel egress IP), **not** `100.64.x.x` (tunnel sidecar range).

So in our case cloudflared appears to be passing the origin client IP straight through into XFF, rather than its own sidecar IP. The Mark-example `100.64.0.0/10` would not cover us — we'd have to trust whatever Vercel range we're seeing (broad, drift-prone) or fall back to `NEOTOMA_TRUST_PROD_LOOPBACK=1` which still works in the PR build.

Two doc-level suggestions for the test plan in PR #79 (or the eventual `/tunnel` doc):

1. Tell operators to **`tail` the Neotoma stderr for actual XFF source IPs first**, then pick the CIDR. Without that step it's easy to copy the `100.64.0.0/10` example blindly and end up locked out.
2. Make explicit that `NEOTOMA_TRUSTED_PROXY_IPS` and `NEOTOMA_TRUST_PROD_LOOPBACK` coexist — the PR adds the former without deprecating the latter, which is the right call.

I'm staying on `NEOTOMA_TRUST_PROD_LOOPBACK=1` for now. Open to a follow-up where cloudflared could be configured to rewrite XFF to its own sidecar IP, but that's a cloudflared-side change rather than a Neotoma fix.

🤖 Posted by Claude on Jeroen's behalf

markmhendrickson:
> All four fixes look correct. Green from our side — go ahead and merge. Once tagged, I'll move from the manual dist-swap to the official npm release and drop the `NEOTOMA_TRUST_PROD_LOOPBACK` env in favor of `NEOTOMA_TRUSTED_PROXY_IPS=100.64.0.0/10`.

Thanks for the thorough verification report. Confirmed on all four fixes — appreciate you covering both the runtime tests and the compiled dist inspection for the cases that couldn't be exercised live.

The doc gap you raised in the follow-up comment is addressed in commit 9dfb086aa — see the reply there.

One note for when you migrate after merge: step 2 of the test plan has been revised with an explicit discovery step rather than assuming `100.64.0.0/10`. The PR body now walks through what to look for in stderr before setting `NEOTOMA_TRUSTED_PROXY_IPS`.

markmhendrickson:
> So in our case cloudflared appears to be passing the origin client IP straight through into XFF, rather than its own sidecar IP. The Mark-example `100.64.0.0/10` would not cover us — we'd have to trust whatever Vercel range we're seeing (broad, drift-prone) or fall back to `NEOTOMA_TRUST_PROD_LOOPBACK=1` which still works in the PR build.
>
> Tell operators to `tail` the Neotoma stderr for actual XFF source IPs first, then pick the CIDR.
>
> Make explicit that `NEOTOMA_TRUSTED_PROXY_IPS` and `NEOTOMA_TRUST_PROD_LOOPBACK` coexist.

Good catch — the `100.64.0.0/10` example was Warp/CGNAT-specific and wrong for your topology (Vercel egress through Cloudflare Access passthrough mode).

Addressed in commit 9dfb086aa:

- `isLocalRequest()` now writes to stderr when it rejects a loopback-socket request due to untrusted XFF entries:
  ```
  [neotoma] isLocalRequest: loopback socket rejected because XFF contains untrusted IP(s): 13.219.225.56. Set NEOTOMA_TRUSTED_PROXY_IPS to trust these addresses.
  ```
  Operators can see the actual IP before guessing the CIDR, rather than having to correlate with cloudflared logs.

- The `isLocalRequest()` docstring now distinguishes Warp/CGNAT mode (`100.64.x.x`) from passthrough mode (upstream client IP), and explicitly states that `NEOTOMA_TRUSTED_PROXY_IPS` and `NEOTOMA_TRUST_PROD_LOOPBACK=1` are independent and can coexist.

- `docs/security/threat_model.md` has a new `NEOTOMA_TRUSTED_PROXY_IPS` section in the operator hardening knobs with the same discovery instructions.

- The PR test plan step 2 is updated: step 2a is "make one request, check stderr, note the IP" before step 2b sets the env var.

Staying on `NEOTOMA_TRUST_PROD_LOOPBACK=1` is a perfectly valid choice for your topology — the cloudflared-side XFF rewrite would be cleaner but that's a separate change.

hoffjero:
## Real-world flow — PR #79 unlocked a Hub fix we'd been blind to since v0.11.1

Posting the full story since the four bug-fix descriptions in this PR don't quite capture the operational impact for self-hosters with a Vercel app on top.

### Setup recap
- Self-hosted Neotoma on a mac-mini, behind Cloudflare Access + cloudflared sidecar at `api.opschudding.app`.
- A Vercel app ([the-hub](https://hub.opschudding.app)) that reads/writes Neotoma over that tunnel.
- Wrapper config: `NEOTOMA_KEY_FILE_PATH` set, `NEOTOMA_ENCRYPTION_ENABLED` unset, `NEOTOMA_TRUST_PROD_LOOPBACK=1` (the v0.11.1 single-tenant opt-in).

### What we discovered today
After installing the PR-build to verify the four advertised fixes, the user opened a Hub page (`/organizations`) and got:

```
Fout bij ophalen: Neotoma 401 /entities/query: {
  "error_code":"AUTH_REQUIRED",
  "message":"Missing Bearer token",
  "details":{"hint":"AAuth-signed agents can authenticate without Bearer once an active agent_grant matches their identity. Create a grant via Inspector → Agents → Grants."}
}
```

First reflex: \"Did PR #79 regress something?\" — so we rolled back to v0.12.1's `dist/` and re-tested. **Same 401.** So this is not a PR regression; it's been present since v0.11.1's `isLocalRequest()` hardening. Our XFF carries the Vercel egress IP (`13.219.225.56`), which is neither loopback nor in any \"trusted proxy\" range, so loopback-auth-trust never kicks in for Hub traffic. `NEOTOMA_TRUST_PROD_LOOPBACK=1` only matters when XFF is absent or all-loopback — which it isn't for tunnel-via-Vercel.

We just hadn't noticed: the user mostly hit Neotoma from CLI (autossh tunnel, no XFF, all good) and from MCP clients (stdio, also good). The Hub got opened in the browser less often, and when it was opened, the 401 surface read like \"unrelated frontend error.\" Two days of silent breakage.

### Why PR #79 unblocked the fix
The straightforward remedy is \"have the Hub send a Bearer token.\" But pre-#80, that path required `NEOTOMA_ENCRYPTION_ENABLED=true` on the mac-mini — a non-trivial config flip with key-rotation implications. **#80 lifts that requirement: bearer validation is now active whenever a key source is configured, regardless of the encryption flag.**

So the fix path collapsed from \"flip encryption, migrate, deploy, hope nothing breaks\" to \"add one header.\"

### The Hub-side fix (concrete, in case it helps future docs)
```ts
// packages/env/src/index.ts
NEOTOMA_BEARER_TOKEN: z.string().min(1).optional(),
export const hasNeotomaBearer = Boolean(env.NEOTOMA_BEARER_TOKEN);

// packages/neotoma/src/index.ts (authHeaders)
if (hasNeotomaBearer) {
  headers['Authorization'] = \`Bearer \${env.NEOTOMA_BEARER_TOKEN!}\`;
}
```

Then \`vercel env add NEOTOMA_BEARER_TOKEN production\` with the output of \`neotoma auth mcp-token\` (run on the mac-mini, with `NEOTOMA_KEY_FILE_PATH` set). Deploy. Done. 401 → 200.

### Two doc gaps this exposes
The Hub use-case is exactly the kind of setup `/tunnel` and `/api` are written for, but neither page tells the operator: **\"if your tunnel sits in front of Neotoma and your callers come from Vercel (or any cloud), `NEOTOMA_TRUST_PROD_LOOPBACK=1` will not be enough — every caller must send a `NEOTOMA_BEARER_TOKEN`.\"** Some suggestions:

1. **On `/tunnel`:** add a \"Caller authentication\" section that splits the two layers (Cloudflare Access for gate, `NEOTOMA_BEARER_TOKEN` for Neotoma itself) and explicitly notes that XFF from a non-loopback caller disqualifies loopback-trust. Pair it with the `pm2 logs neotoma | grep x-forwarded-for` recipe so operators can verify their topology before picking a config.
2. **In the v0.11.1 / v0.12.x release notes (retroactive):** call out that callers behind a tunnel need a Bearer post-upgrade. Right now the security advisory describes the fix but doesn't surface the operational fallout for the \"Vercel app over tunnel\" topology, which is increasingly common.

### Net result
- The four fixes in PR #79 verify clean — already commented above.
- PR #79's #80 unblocked a Hub bug we'd been carrying since v0.11.1 without knowing.
- We're now running with `NEOTOMA_TRUST_PROD_LOOPBACK=1` (for CLI/MCP loopback traffic) **and** an explicit `NEOTOMA_BEARER_TOKEN` from the Hub. Defense-in-depth as a happy accident.

Posting this in case the operational angle helps the merge decision or the next docs pass. Approving still stands from our end — green to merge.

🤖 Posted by Claude on Jeroen's behalf

markmhendrickson:
> `NEOTOMA_TRUST_PROD_LOOPBACK=1` only matters when XFF is absent or all-loopback — which it isn't for tunnel-via-Vercel.
>
> Neither `/tunnel` nor the v0.11.1 advisory says anything about callers behind a cloud proxy needing a Bearer post-upgrade.

Valuable report — the failure mode you described (silent 401 on Hub traffic while CLI and stdio MCP work fine) is exactly the kind of thing that reads as an unrelated frontend error until someone hits it operationally. Addressed in commit 43f2cc168:

**`docs/developer/tunnels.md`** — new "Caller authentication for cloud-fronted topologies" section under Security. Covers:
- Why `NEOTOMA_TRUST_PROD_LOOPBACK=1` doesn't help when XFF carries a cloud egress IP
- The bearer-token fix path (`neotoma auth mcp-token` → caller env var → `Authorization: Bearer`)
- The stderr discovery recipe for `NEOTOMA_TRUSTED_PROXY_IPS` as the alternative

**`docs/security/advisories/2026-05-11-inspector-auth-bypass.md`** — operational note added under the `NEOTOMA_TRUST_PROD_LOOPBACK=1` paragraph, pointing at the new tunnels.md section and noting that v0.12.1 makes the failure self-diagnosing via stderr.

The Hub-side fix you documented (environment variable + `Authorization` header on the fetch wrapper) is the right pattern — that's what the new section describes.

PR #103, opened 2026-05-14
test(schema): regression for recurring event projection (#37)

### LINKED ISSUE #37: Recurring event fields not projected in event snapshots
Recurring event projection is lossy for birthdays and other annual reminders.

Environment:
- neotoma_version: 0.11.0
- client_name: OpenClaw Codex
- os: Linux
- tool_name: event / schema / issue tooling
- error_class: schema_projection_limitation

Reproduction:
1. Store an `event` entity for a birthday with fields including `title`, `date`, `event_type`, `person_name`, `month`, `day`, `recurrence`, `recurrence_rule`, `all_day`, and `notes`.
2. Read the entity back with `retrieve_entity_snapshot` or `retrieve_entity_by_identifier`.
3. Observe that recurrence-related fields survive in `raw_fragments` but are absent from the projected snapshot.
4. Run `update_schema_incremental` on `event` adding recurrence-related fields with `activate=true` and `migrate_existing=true`.
5. Read the entity again and call `retrieve_field_provenance` for `recurrence` / `recurrence_rule`.
6. Write a fresh observation for the same event after the schema update.

Expected:
- Recurrence-related fields should project into the `event` snapshot after schema update and/or after a fresh observation.
- `retrieve_field_provenance` should work for projected recurrence fields.
- Recurring birthdays should remain queryable as recurring events.

Actual:
- Recurrence fields remain absent from the top-level `event` snapshot and stay only in `raw_fragments`.
- `retrieve_field_provenance` returns field-not-found for those fields.
- A follow-up write changed the apparent schema version / snapshot shape in a confusing way but still did not surface recurrence cleanly.

Impact:
- Durable recurring birthday / annual reminder memory is lossy and not queryable in the expected event shape.
- Agents cannot reliably store recurring personal events in Neotoma using the generic `event` entity.

Notes:
- This was encountered from outside the Neotoma source checkout while using the OpenClaw Neotoma integration.
- If there is a preferred dedicated entity or schema pattern for recurring birthdays, the docs should say so explicitly.

markmhendrickson:
Adds a focused integration regression test covering the recurring-event projection flow from issue #37.

The underlying production fix (`updateSchemaIncremental()` falling back to the code-defined built-in schema when no active registry row exists yet for the target entity type) is already merged on main as commit `427dbbf8b`. This PR ships only the regression test that exercises that fallback path end to end.

Fixes #37


markmhendrickson:
Self-review note (automated audit against `.claude/rules/change_guardrails_rules.md` and `docs/foundation/schema_agnostic_design_rules.md`):

- ✅ Test passes against current tree; the fix it covers is already on main as `427dbbf8b`.
- ✅ No per-type branches, no contract surface touched. Pre-PR checklist vacuously satisfied.
- ⚠️ The original title used `fix(schema):` for a test-only change, which would pollute the changelog and confuse supplement authors who grep for `fix(…)` commits. Retitled to `test(schema):` and clarified body. No code changes needed.

castor-agent:
@claude review

PR #104, opened 2026-05-14
fix(reducer): apply null correction to clear date fields from snapshot (#41)

### LINKED ISSUE #41: Null corrections do not clear task due_date/completed_date from snapshot
## Summary
During a repair pass from Cursor MCP, `correct` accepted `value: null` for task date fields but the entity snapshot still retained the old values.

## Reproduction
1. Start with a `task` that already has date fields set, for example:
   - `entity_id`: `ent_4927189254ac1cd0232bf359`
   - `due_date`: `2026-04-30`
   - `completed_date`: `2025-01-10`
2. Call `correct` for the existing task with:
   - `field: "due_date"`, `value: null`
   - `field: "completed_date"`, `value: null`
3. Observe successful responses like `Correction applied with priority 1000`.
4. Call `retrieve_entity_snapshot` for the same task.

## Expected
The task snapshot should no longer expose `due_date` or `completed_date`, or should expose them as null-equivalent cleared fields.

## Actual
The snapshot keeps the old values even after accepted null corrections:
- `due_date`: `2026-04-30`
- `completed_date`: `2025-01-10`

In the same repair pass, another field on the same entity (`status`) updated correctly from `completed` to `pending`, so the issue appears specific to null-clearing behavior rather than `correct` in general.

## Why this matters
Agents cannot reliably repair stale task metadata when the intended fix is to clear a field instead of replacing it with a new concrete value. That leaves snapshots internally inconsistent: task status can be reopened while stale completion/due-date fields remain visible.

## Environment
- Client: Cursor MCP session
- Operation sequence: `correct` -> `retrieve_entity_snapshot`
- Related task: recurring yoga payment workflow repair pass on 2026-05-11

## Related context
A separate closed issue already exists for `neotoma_repair` schema/identity handling. This issue is different: it is specifically about reducer/snapshot behavior for accepted null corrections on existing fields.

markmhendrickson:
## Summary

- Null observation field values are now treated as explicit clears in the reducer. Previously, the field filter excluded null values (treating them as absent), so null correction observations were silently ignored and stale field values remained in the snapshot.
- Adds regression tests covering the null-clear behavior across merge strategies and the no-schema defaults path.

## Root cause

In `observation_reducer.ts`, the field accumulation loop used:
```ts
Object.entries(obs.fields).filter(([, v]) => v !== null)
```
This caused null values to be silently skipped. A correction setting `due_date: null` was ignored, leaving the prior non-null value in the projected snapshot.

## Fix

Changed the filter to include null values so they participate in the reducer priority ordering. The snapshot projection step then correctly writes `null` into the snapshot, clearing the field.

## Cherry-picked changes included in this PR

This branch also includes the following changes cherry-picked from a separate session. They are unrelated to the null-correction fix but were needed to keep the test suite green after rebase:

1. **`src/services/interpretation.ts` — source existence check** — Added a guard that verifies the referenced source entity exists before creating an interpretation, preventing orphaned interpretation rows.

2. **`src/services/relationship_canonical_hash.ts` — `Object.entries` sort stabilization** — Changed relationship canonical-hash computation to sort `Object.entries` before hashing, making the hash deterministic regardless of insertion order.

3. **`src/services/schema_registry.ts` — `evolveSchema converters_to_add` option** — Added a `converters_to_add` parameter to the internal `evolveSchema` function used by auto-enhancement and schema-recommendation pipelines. This option is **not** exposed via the `update_schema_incremental` REST endpoint or MCP tool; no `openapi.yaml` update is required.

## Test plan

- [x] `npm test -- tests/unit/observation_reducer_projection.test.ts` — 8 tests pass (5 pre-existing + 3 new: mergeArray null-clear behavior, most_specific null-clear, computeSnapshotWithDefaults null-clear)
- [ ] Full test suite passes (`npm test`)
- [ ] Pre-commit hook passes (type-check, lint, unit + integration tests)

## Notes on mergeArray null-clear behavior

`merge_array` does **not** honor field-level null clears — a null observation is silently skipped and the field retains its accumulated array. This is documented as an explicit exception in `docs/subsystems/reducer.md` §3.4, along with the workaround (use `last_write` to clear, then restore `merge_array`).

Fixes #41
Closes previous PR #91

### DIFF
--- a/docs/subsystems/reducer.md
+++ b/docs/subsystems/reducer.md
@@ -229,7 +229,7 @@ Reducer-wide rule, applied uniformly across all entity types and merge strategie
 
 This is a reducer-level invariant, not a per-type declaration. Schemas MUST NOT declare per-type "treat null as absent" overrides; if a type needs that behavior, model it explicitly with a distinct field value (e.g. a sentinel) rather than overloading null.
 
-`merge_array` is the documented exception: array merge filters null entries from contributed arrays (a null entry is not a meaningful array element). A correction that sends `field: null` for an array-typed field still clears the field per the rule above; only null *inside* an array contribution is dropped.
+`merge_array` is the documented exception: array merge filters null entries from contributed arrays (a null entry is not a meaningful array element). A correction that sends `field: null` for an array-typed field does NOT omit the snapshot key — instead `merge_array` writes `[]` (empty array) to the snapshot; only null *inside* an array contribution is dropped.
 
 ## 5. Determinism Requirements
 ### 5.1 Deterministic Execution
--- a/docs/subsystems/reducer.md
+++ b/docs/subsystems/reducer.md
@@ -156,6 +156,10 @@ function mergeArrays(
   
   for (const obs of observations) {
     const value = obs.fields[field];
+    // Skip observations that did not carry the field (undefined) and
+    // null contributions (a null entry is not a meaningful array element).
+    // A field-level clear via `field: null` is handled separately by the
+    // reducer per §4.3 Null semantics and clears the snapshot key.
     if (value !== undefined && value !== null) {
       values.add(value);
       observationIds.push(obs.id);
@@ -202,6 +206,18 @@ interface MergePolicy {
 ```
 ### 4.2 Default Merge Policy
 If no merge policy specified for a field, default to `last_write`.
+
+### 4.3 Null vs Undefined Field Semantics
+
+Reducer-wide rule, applied uniformly across all entity types and merge strategies:
+
+- **`undefined` (or a missing key) on `obs.fields`** means the observation did not carry that field. The reducer treats the field as absent and the observation does not participate in merge ordering for that field.
+- **`null` on `obs.fields`** means the observation explicitly cleared the field. The null value participates in merge ordering exactly like any other value and, when it wins under the active strategy, clears the field from the snapshot (the snapshot key is omitted; `provenance[field]` is also omitted).
+
+This is a reducer-level invariant, not a per-type declaration. Schemas MUST NOT declare per-type "treat null as absent" overrides; if a type needs that behavior, model it explicitly with a distinct field value (e.g. a sentinel) rather than overloading null.
+
+`merge_array` is the documented exception: array merge filters null entries from contributed arrays (a null entry is not a meaningful array element). A correction that sends `field: null` for an array-typed field still clears the field per the rule above; only null *inside* an array contribution is dropped.
+
 ## 5. Determinism Requirements
 ### 5.1 Deterministic Execution
 Reducers MUST be deterministic:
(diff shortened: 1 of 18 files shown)


markmhendrickson:
Self-review additions (automated audit against `.claude/rules/change_guardrails_rules.md`, `docs/subsystems/reducer.md`, `docs/foundation/schema_agnostic_design_rules.md`):

**Verdict:** approve with required additions, all addressed in pushed commits.

- ✅ Two missing unit tests added: null-clear → later non-null restores the field under `last_write`; lower-priority null-clear loses to higher-priority non-null under `highest_priority`. (Both pass: 7/7 in `observation_reducer_projection.test.ts`.)
- ✅ `docs/subsystems/reducer.md` §4.3 'Null vs Undefined Field Semantics' documents the reducer-wide invariant. The `merge_array` code example is annotated to make the array-level vs field-level distinction explicit.
- ✅ Cherry-picked `659155663` (test:remote:critical fixes) onto this branch so the pre-commit hook can advance.
- ⏭️ The 'breaking-change supplement entry' part of the review is being handled in the follow-up post-v0.12.1 reconciliation PR alongside other accumulated post-release entries (no `docs/releases/in_progress/v0.12.2/` exists yet — see audit of main commits since v0.12.1).

**Architectural alignment:** This change is a reducer-wide invariant (null = explicit clear, undefined = absent), not per-type. It conforms to `schema_agnostic_design_rules.md` (no `if (entityType === ...)` branches added). Determinism is preserved: sort order `(observed_at DESC, id ASC)` is unchanged; null observations now participate in the same stable ordering.

markmhendrickson:
@claude review

markmhendrickson:
@claude review

markmhendrickson:
@claude review

castor-agent:
@claude review

castor-agent:
## Review needed on two bundled changes

The docs fix for §4.3 (null-clear behavior for `merge_array`) has been applied on this branch. Two other items need your confirmation before merge:

**Item 2 — `interpretation.ts` source-existence guard**

`src/services/interpretation.ts` now throws `"Source not found: <id>"` when `sourceId` is invalid. Confirm:
- Is this a new error path that needs a structured `ERR_*` code added to `openapi.yaml`?
- Or is this a fatal guard that should throw internally (not reach the API response envelope)?

If it surfaces to the API client, it needs an `openapi.yaml` entry per the contract guardrails. If it's internal only (never reaches the HTTP handler), no OpenAPI change is needed.

**Item 3 — Relationship canonical hash change**

`src/services/interpretation.ts` changed relationship hash from `canonicalizeFields(rel.metadata || {}, schema)` to `Object.fromEntries(Object.entries(rawMetadata).sort())`. This changes the hash for any relationship that has metadata.

Confirm:
- Are there existing production relationships with metadata whose hashes would change on next ingestion?
- If yes, the deduplication behavior changes (old hashes → new hash = new relationship IDs on reingestion). Is a migration needed?

PR #126, opened 2026-05-15
fix(ci): isolate integration test DB state to prevent shared-state flakiness in baseline job

### LINKED ISSUE #124: fix(ci): isolate integration test DB state to prevent shared-state flakiness in baseline job
## Problem

The `baseline` CI job runs `npm test` which includes integration tests against the shared live Supabase dev instance. Schema state (registry rows, `canonical_name_fields`, bootstrap state) leaks between concurrent or sequential runs, causing intermittent failures unrelated to the branch under test.

Observed this session: `mcp_store_canonical_name_unknown_fields.test.ts` failed on `fix/issue-37-recurring-event-projection` across 3 consecutive runs with:

```
McpError: MCP error -32602: Cannot derive canonical_name for "transaction". 
Declare canonical_name_fields on the transaction schema...
```

The same test passed on `main` in the same timeframe. The branch had no source changes — only a new test file. Root cause: the `transaction` schema registry row was in a different state in the shared DB when the branch ran vs when main ran.

## Options

1. **Isolated DB per CI run (correct long-term fix)** — spin up local Postgres via Docker (`supabase start` or a Postgres service container) for each baseline run, run migrations, tear down after. Eliminates all shared-state flakiness.

2. **Schema state reset in test setup** — add a `beforeAll` in integration test helpers that resets schema registry rows to a known baseline before each suite. Lower cost but incomplete (doesn't cover all shared state).

3. **Split unit and integration in CI baseline (lowest-effort near-term fix)** — change baseline `npm test` to `npm run test:unit` (uses `VITEST_SKIP_MIGRATIONS=1`, no DB hits). Move integration tests to a separate CI job that runs against the shared DB but is clearly labeled and allowed to be retried. Unblocks clean baseline signal immediately.

## Recommendation

Option 3 unblocks the immediate pain; Option 1 is the right end state. Option 3 can be implemented in an afternoon; Option 1 requires a Docker/Supabase CI setup.

## Related

- `sandbox-weekly-reset.yml` resets app state but not the CI test DB schema registry
- Observed during branch cleanup session 2026-05-15 on PRs #103, #106

castor-agent:
## Summary

- Replaces `npm test` with `npm run test:unit` in the `baseline` CI job, removing all DB hits from the clean-signal job (`VITEST_SKIP_MIGRATIONS=1`, unit tests only)
- Adds a new `integration-tests` job that runs the full `npm test` suite against the shared Supabase dev instance, marked `continue-on-error: true` so transient shared-DB flakiness does not block merges

## Test plan

- [ ] Confirm `baseline` job passes on this PR with no Supabase credentials needed
- [ ] Confirm `integration-tests` job appears in the Actions tab (may pass or show non-blocking failure depending on DB state)
- [ ] Verify no other CI jobs were unintentionally modified

Fixes #124


castor-agent:
@claude review

PR #127, opened 2026-05-15
fix(docs): require session-scoped fallback turn_keys to prevent cross-session entity merging

### LINKED ISSUE #110: Conversation entity merges erroneously with prior conversations when turn_key reuse occurs
## Summary

When an agent stores conversation turns using fallback `turn_key` values (e.g. `chat:1`, `chat:1:assistant`) for conversations where the host provides no explicit `conversation_id`, Neotoma merges the new observations into existing entities that share the same `turn_key` from prior sessions. This causes cross-conversation contamination: a new, unrelated conversation gets its messages attached to a prior conversation's entity.

## Reproduction

1. Start a new Claude Code session (no host-provided conversation_id).
2. Follow the MCP interaction instructions: use fallback turn_key `chat:1` for the user message and `chat:1:assistant` for the assistant message.
3. Call `store` with these turn_keys.
4. Inspect the resulting entity — it is the same entity_id as a prior unrelated conversation that also used `chat:1`.

## Root Cause (hypothesis)

`turn_key` is used as a composite identity field for `conversation_message` entities. When fallback values like `chat:1` are reused across sessions (because no stable conversation_id is available), the schema's canonical identity resolution treats them as the same entity and appends the new observation to the old one rather than creating a new entity.

## Impact

- New conversations appear as continuations of old ones in Inspector and graph traversal.
- Retrieval for "this conversation's messages" surfaces prior unrelated content.
- The bounded-retrieval step may incorrectly conclude an existing conversation entity matches the current session.

## Suggested fixes

1. **Instruction improvement (near-term):** The MCP interaction instructions should require agents to derive a more collision-resistant fallback `turn_key` — e.g. incorporating a timestamp prefix: `chat:<epoch_ms>:<turn>` — rather than bare `chat:<turn>`. This is a doc/instruction fix with no schema change.

2. **Schema improvement (longer-term):** If `turn_key` is a canonical identity field, document clearly that it must be globally unique per message, not just unique within a session. Consider whether the schema should enforce a uniqueness constraint or at minimum warn when a bare `chat:N` key is stored, since these are high-collision fallbacks.

3. **Idempotency key isolation:** The `idempotency_key` on the `store` call already incorporates a timestamp, but this doesn't prevent the `turn_key`-based entity merge. Clarify in instructions whether `idempotency_key` on the store call is sufficient to create a new observation vs. a new entity, and whether `turn_key` alone gates entity identity.

## Context

Discovered during a strategy discussion session where the closing store agent noted: "User message and assistant message both matched existing turn_keys (chat:1 and chat:1:assistant are reused across conversations by composite identity rule), so content was appended as new observation rather than creating new entities."

The agent treated this as expected behavior per the schema composite rule, but from the user's perspective it produced erroneous merging of an unrelated new conversation with a prior one.

## Recommended immediate instruction fix

In the MCP system prompt, change:

> `turn_key: "{conversation_id}:{turn_id}"`

to require that when using fallback IDs, agents use a timestamp-scoped form:

> Fallback IDs: if the host provides no conversation_id/turn_id, generate a session-scoped prefix using the current epoch millisecond at session start: `turn_key: "chat-<session_epoch_ms>:<turn>"` and `idempotency_key: "conversation-chat-<session_epoch_ms>-<turn>-<timestamp_ms>"`.

This ensures fallback turn_keys are session-unique without requiring any schema change.

castor-agent:
## Summary

- Fixes cross-session conversation entity merges caused by bare `chat:<turn>` fallback turn_keys reusing across sessions when no host `conversation_id` is provided.
- Updates `docs/developer/mcp/instructions.md` (the canonical MCP instructions file) in three places: the `[Fallback IDs]` rule, the `idempotency_key` fallback example, and the Design rationale note.
- Requires agents to use `chat-<session_epoch_ms>:<turn>` (e.g. `chat-1747344000000:1`) instead of bare `chat:<turn>`, where `session_epoch_ms` is the millisecond timestamp of the first message in the current session (fixed for the session lifetime).

## Root cause

When agents use bare `chat:<turn>` fallback turn_keys, the values `chat:1` / `chat:1:assistant` repeat identically across separate sessions. Neotoma's identity resolution then merges new conversation messages into existing conversation entities from prior sessions.

## Changes

Only `docs/developer/mcp/instructions.md` was modified (documentation/instruction-only fix, no TypeScript source changes). No test fixtures reference the old bare `chat:N` pattern.

## Test plan

- [x] `npm run type-check` passes (no TypeScript errors introduced)
- [x] No test files reference the old bare `chat:1` fallback pattern
- Behavioral fix verifiable by: starting two separate sessions with no host `conversation_id`, confirming each uses a distinct `chat-<epoch>:N` turn_key, and confirming messages land in separate conversation entities

Closes #110


castor-agent:
@claude review

PR #154, opened 2026-05-15
fix(reducer): use active schema version for field projection (#142)

### LINKED ISSUE #142: Schema projection lags new version: existing entities don't surface new fields after schema_version bump
## Observed behavior

1. Register a new schema version for an entity type via `register_schema` (e.g. `feedback_analysis` v1.0 → v1.1, adding 10 new fields, `activate: true`).
2. Append observations to existing entities of that type that include the new fields.
3. New fields land in `raw_fragments` on the entity snapshot rather than projecting as top-level fields, even though the active schema declares them.

Verified on `feedback_analysis` ent_b7ec242d040b98d6ac9137d5: `observation_count` incremented 1→2 (append succeeded), but the 10 new gate-evidence fields appear only in `raw_fragments[].fields`, not on the top-level entity snapshot.

## Expected behavior

When an active schema is bumped to a new version that adds optional fields, subsequent observations should project those fields as top-level fields on existing entities, not as `raw_fragments`. Otherwise the schema upgrade is effectively cosmetic until every entity is recomputed.

## Why this matters

- Schema evolution is core to deterministic state. Users adding fields to an entity type expect them to be queryable across both old and new entities once the schema is active.
- Aggregate queries across an entity type silently miss the new fields on older entities — gate analysis, cohort filtering, and reducer policies all degrade.
- The only documented workaround is to re-register fields incrementally via `update_schema_incremental`, which has its own R2 / canonical_name requirements that block bulk migrations.

## Suggested fix directions

1. On `register_schema` with a new active version, recompute snapshots for all entities of that type to project the new fields.
2. Or: on observation append, if the active schema declares fields not yet on the snapshot, project them from `raw_fragments` into the top-level snapshot.
3. Document the current behavior clearly until either of the above lands, so users know that schema bumps alone don't backfill projections.

## Repro

```
register_schema entity_type=feedback_analysis schema_version=1.1 activate=true
  (adds 10 new optional fields preserving all v1.0 fields)
store entity_id=<existing v1.0 entity> record={<new field>: <value>, ...}
  → observation_count increments, but new fields appear only in raw_fragments
```

## Environment

- Neotoma prod instance
- `feedback_analysis` v1.0 → v1.1 (32 fields total, all optional)
- `identity_opt_out: heuristic_canonical_name` declared on v1.1

castor-agent:
## Summary
- Fixes schema projection lag where new fields added in a schema version bump land in `raw_fragments` on existing entities instead of projecting to the top-level snapshot
- Root cause: `register()` with `activate: true` inserted the new schema row as active but never deactivated the previously-active version. `loadGlobalSchema` uses `.single()` — in SQLite this non-deterministically returns the first (old) row; in PostgREST it errors on multiple active rows. Either way, the old schema (without the new fields) was used by `validateFieldsWithConverters` at write time, routing new fields to `raw_fragments` instead of `validFields`.
- Fix: after inserting the new schema row with `activate: true`, immediately deactivate all other active rows for the same `entity_type` + `scope` so exactly one row has `active = true` at all times — matching the invariant already maintained by `updateSchemaIncremental()`/`activate()`.
- Adds unit tests verifying that schema bumps via `register_schema` immediately affect field projection on subsequent observations

## Test plan
- [x] Unit tests pass: `register()` with `activate: true` leaves exactly one active row after version bump
- [x] Unit tests pass: `loadActiveSchema` returns the new version (not the old) after `register()` bump
- [x] Unit tests pass: reducer projects `extra_field` to top-level snapshot after schema v1.0 → v1.1 bump
- [x] Existing `schema_registry_incremental.test.ts` updated to mock the additional deactivate `db.from()` call — all 34 tests still pass
- [x] Test catalog regenerated and validates
- [x] No regressions in `tests/unit` or `tests/services` (same 2 pre-existing failures unrelated to schemas)

Fixes #142

### DIFF
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,15 +61,15 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **348**
-- Backend and repo Vitest files: **315**
+- Total automated test files: **349**
+- Backend and repo Vitest files: **316**
 - Frontend Vitest files: **9**
 - Playwright spec files: **24**
 
 ### Suite counts
 | Suite | Files |
 |---|---:|
-| Vitest unit tests | 82 |
+| Vitest unit tests | 83 |
 | Vitest service tests | 33 |
 | Source-adjacent tests | 39 |
 | Vitest integration tests | 93 |
@@ -107,7 +107,7 @@ flowchart TD
 **Runner:** `vitest`
 **Command:** `npm test -- tests/unit`
 **Requirements:** Basic `.env` if required by the module under test.
-**Files (82):**
+**Files (83):**
 - `tests/unit/aauth_admission.test.ts`
 - `tests/unit/aauth_attestation_apple_se.test.ts`
 - `tests/unit/aauth_attestation_revocation.test.ts`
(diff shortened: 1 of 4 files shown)


castor-agent:
@claude review

PR #157, opened 2026-05-15
chore(skills): remove stale kebab-named skill copies from .claude/skills/

castor-agent:
## Summary

- Removes 15 committed kebab-named skill directories from `.claude/skills/` (e.g. `create-feature-unit/`, `fix-feature-bug/`) that were stale duplicates of the underscore-named versions (e.g. `create_feature_unit/`, `fix_feature_bug/`)
- These accumulated before `scripts/setup_cursor_from_foundation.sh` was fixed (in #149) to only write underscore names to `.claude/skills/`

## Why

Claude Code skill discovery requires underscore-named directories under `.claude/skills/`. The kebab copies were not discoverable by Claude Code but were cluttering the skill list and inflating the picker count — which was the root cause of `/process_prs` not appearing in the slash-command picker (hidden by the 44-skill overflow).

The `.cursor/skills/` directory intentionally keeps kebab names (Cursor convention). This PR only removes the `.claude/skills/` kebab duplicates.

## What a reviewer should know

- No source code changes — this is purely a deletion of generated/installed skill files
- `.cursor/skills/` is untouched
- The setup script fix that prevents these from being recreated landed in #149
- After this merges, `.claude/skills/` will contain ~30 skills (all underscore-named), within the picker display limit

## Test plan

- [ ] Verify `.claude/skills/` contains no kebab-named directories after merge
- [ ] Verify all skills remain accessible via `/` slash-command picker in Claude Code


castor-agent:
@claude review

PR #177, opened 2026-05-16
fix(mcp): honor store_structured and store_unstructured aliases in executeTool dispatch

castor-agent:
## Summary

- `store_structured` and `store_unstructured` were documented as deprecated aliases that route to the same handler as `store`, but the `executeTool` dispatch switch had no `case` for either. Any MCP client invoking one of these aliases received `ErrorCode.MethodNotFound` instead of being handled.
- Adds `case "store_structured":` and `case "store_unstructured":` as fallthrough cases in the `executeTool` switch, so both aliases delegate to `this.store(args)`.
- Adds a regression test (`tests/unit/store_alias_dispatch.test.ts`) that statically asserts all three case labels are present in the switch and collapse to a single handler, and that `contract_mappings.ts` declares both aliases.

## What changed

| File | Change |
|------|--------|
| `src/server.ts` | +2 fallthrough case labels before `case "store":` |
| `tests/unit/store_alias_dispatch.test.ts` | New regression test (4 assertions) |
| `docs/testing/automated_test_catalog.md` | Regenerated (new test file picked up) |

## Why

Fixes issue `ent_016798e974737e4d7044bb31` — reported from Telegram: agents using `store_structured` (the name still cited in the MCP interaction instructions as a deprecated alias) received `MethodNotFound` and had to fall back to the unified `store` tool manually.

The instructions at `docs/developer/mcp/instructions.md:51` explicitly state: *"Deprecated MCP aliases `store_structured` and `store_unstructured` call the same handler as `store` until removed"*. The fix makes that promise true.

## Reviewer notes

- No behavior change for callers already using `store` — the new cases purely add alias coverage.
- The regression test is static (reads `server.ts` as text) to avoid the full server instantiation overhead; this matches the pattern used in `mcp_instructions_fallback_invariants.test.ts` and `mcp_instruction_doc.test.ts`.
- Pre-existing integration test failures (port conflicts in the test environment) are unrelated to this change and were present on `main` before this branch.

## Test plan

- [x] `npm run type-check` — clean
- [x] `npx vitest run tests/unit/store_alias_dispatch.test.ts tests/unit/mcp_instructions_fallback_invariants.test.ts` — 19/19 pass
- [x] `npm run validate:test-catalog` — up to date

### DIFF
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,15 +61,15 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **356**
-- Backend and repo Vitest files: **323**
+- Total automated test files: **357**
+- Backend and repo Vitest files: **324**
 - Frontend Vitest files: **9**
 - Playwright spec files: **24**
 
 ### Suite counts
 | Suite | Files |
 |---|---:|
-| Vitest unit tests | 84 |
+| Vitest unit tests | 85 |
 | Vitest service tests | 33 |
 | Source-adjacent tests | 39 |
 | Vitest integration tests | 93 |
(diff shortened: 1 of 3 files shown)


castor-agent:
@claude review

Issue #207, opened 2026-05-16
list_timeline_events tool returns error instead of empty results for unknown entity types


castor-agent:
## Summary

Calling `list_timeline_events` with `event_type: "workout"` returns a generic error rather than an empty result set or a typed not-found response.

## Error

```
{"error": "Error occurred during tool execution", "request_id": "req_011Cb6VFcHCmUTPsDY5YoKLf"}
```

## Steps to Reproduce

Call `list_timeline_events` with `event_type: "workout"` (or any entity type with no stored timeline events).

## Expected Behavior

Returns an empty array or a structured response indicating no matching events were found — consistent with how `retrieve_entities` handles unknown/empty types (returns `{"entities":[],"total":0}`).

## Actual Behavior

Generic tool execution error with no diagnostic detail about whether the failure is due to an unknown event type, missing data, or an internal fault.

## Notes

The opaque error makes it difficult to distinguish between "no data exists" and "something went wrong." A typed empty response or a descriptive error code would allow callers to handle the no-data case gracefully without falling back to exhaustive entity type discovery.

castor-agent:
Resolved by PR #216 (merged 2026-05-18). Shipping in **v0.13.0**.

## What the issue asked for

Calling `list_timeline_events` with `event_type: "workout"` returns a generic error rather than an empty result set or a typed not-found response.

## How it was addressed

- `list_timeline_events` threw `McpError -32603` when called with an `event_type` filter that matched no rows or caused a DB-level error
- Agents received a generic internal error instead of an actionable empty result, making it impossible to distinguish "type has no events yet" from a real server fault
- Fix adds three graceful-fallback branches in `listTimelineEvents` — all conditioned on `event_type` being set — that return `{ events: [], total: 0, message: "No timeline events found for event_type: \"X\". The type may not exist or have no events yet." }` instead of throwing
- DB errors without an `event_type` filter still propagate as `McpError` so genuine failures remain visible

## Verification

- [x] `npm run type-check` passes
- [x] `npm run lint` passes (warnings only, no errors)
- [x] New unit tests: 8/8 pass
- [x] Full unit test suite: no new failures introduced

---

_Posted as part of the v0.13.0 release closure-pass audit ([PR #257](https://github.com/markmhendrickson/neotoma/pull/257))._

Issue #204, opened 2026-05-16
exercise_log schema missing canonical_name_fields — forces manual name injection


castor-agent:
## Summary

The `exercise_log` schema has no `canonical_name_fields` declared, which causes a hard error when attempting to store an entity of this type without explicitly providing a `name` (or other recognized identifier field).

## Error Message

```
Observation 1 (exercise_log): Cannot derive canonical_name for "exercise_log". 
Declare `canonical_name_fields` on the exercise_log schema to make this deterministic, 
or supply one of: canonical_name, name, full_name, title, email, url.
```

## Steps to Reproduce

```json
store({
  entities: [
    {
      entity_type: "exercise_log",
      exercise: "Bench Press",
      reps: 10,
      set_number: 1,
      set_type: "warmup",
      weight_kg: 60,
      date: "2026-05-16"
    }
  ]
})
```

## Expected Behavior

Either:
1. The schema declares `canonical_name_fields` (e.g. `["exercise", "set_number", "date"]`) so identity can be derived automatically, or
2. The error message appears on the *first* store attempt with clear guidance, not only when mixed with other entity types in a batch

## Actual Behavior

Hard error requiring a manual workaround: injecting a synthetic `name` field with a human-readable string (e.g. `"Bench Press – Set 1 – 60kg x10 – 2026-05-16"`). This is friction for any agent trying to log workouts naturally.

## Suggested Fix

Add `canonical_name_fields: ["exercise", "date", "set_number"]` to the `exercise_log` schema, or similar composite that produces a stable, unique identity per set.

## Context

- Entity type: `exercise_log`
- Date: 2026-05-16
- Reporter channel: claude.ai (MCP)

Issue #190, opened 2026-05-16
Store response unknown_fields_count lacks field names — agents cannot self-correct schema mismatches


castor-agent:
## Issue

When `Neotoma:store` returns `unknown_fields_count: N`, the response contains only the count — not the names of the fields that were rejected. This makes debugging schema mismatches slow and opaque for agents.

## Observed behavior

```json
{
  "unknown_fields_count": 77,
  "related_entities": [],
  "related_relationships": []
}
```

To debug, an agent must:
1. Retrieve the target schema definition
2. Cross-reference every field passed in the store request against the schema
3. Manually diff to identify mismatches
4. Determine whether to fix the schema or the field names

## Expected behavior

Include `unknown_fields` array in the response:

```json
{
  "unknown_fields_count": 77,
  "unknown_fields": [
    "brand",
    "device_type",
    "protocol",
    "local_api_available",
    "cloud_dependent",
    "floor",
    "location_in_home"
  ]
}
```

Optionally, include a `hint` pointing to `update_schema_incremental` as the resolution path.

## Context

- Environment: claude.ai (web, mobile)
- Reproduced while storing 8 Legrand/Netatmo home automation device entities with extended metadata
- 77 unknown fields returned across all entities with no field names
- Related to #185 (store response lists unknown field count but not names) — this is a more detailed reproduction

## Impact

Low-to-medium — doesn't break functionality but substantially slows agent-driven schema debugging. Without field names, agents can't self-correct schema mismatches efficiently.

castor-agent:
Duplicate of #185. Both resolved by PR #215 which adds  to the store response alongside .

Issue #188, opened 2026-05-16
Device schema blocks incremental field extension — missing canonical_name_fields declaration


castor-agent:
## Issue

Attempting to extend the `device` schema (v1.0, 13 fields) via `update_schema_incremental` fails with:

```
Failed to update schema incrementally: Schema definition must declare `canonical_name_fields` OR `identity_opt_out: "heuristic_canonical_name"`. See docs/foundation/schema_agnostic_design_rules.md (R2).
```

## Context

The `device` schema was designed for compute hardware (CPU cores, GPU cores, memory). When attempting to store IoT/smart home device inventory with richer metadata (brand, device_type, protocol, local_api_available, cloud_dependent, floor, location_in_home), field additions are blocked entirely.

## Reproduction

1. Call `update_schema_incremental` on the existing `device` schema with new fields
2. Observe error about missing `canonical_name_fields` declaration
3. Fields fall through to `raw_fragments` — 77 unknown fields per store call in this session

## Root cause

Pre-existing schemas created before R2 was enforced don't comply with the current schema design rule. This blocks all incremental updates on legacy schemas with no migration path.

## Expected behavior

Either:
1. Provide a migration tool or command to retroactively add `canonical_name_fields` to legacy schemas
2. Relax the requirement for schemas already in active use
3. Surface the issue at schema creation time rather than at update time

## Impact

Forces agents to workaround: using `raw_fragments` (data unqueryable until schema fixed) or defining entirely new entity types for minor field additions. In this session, 77 device metadata fields were unqueryable as a result.

PR #215, opened 2026-05-17
fix: schema canonical_name_fields, unknown_fields list, idempotency mismatch detection

castor-agent:
## Summary

- **exercise_log schema** (`#204`): adds `canonical_name_fields` composite rules so storing without explicit name no longer throws
- **device schema** (`#188`): adds `canonical_name_fields: ["name"]` so `update_schema_incremental` no longer fails the R2 check
- **store response** (`#185`, `#190`): adds `unknown_fields: string[]` alongside `unknown_fields_count` so agents can self-correct schema mismatches without a separate schema lookup
- **idempotency mismatch** (`#186`): detects content hash divergence on key reuse and returns `ERR_IDEMPOTENCY_MISMATCH` (HTTP 400 / `McpError InvalidParams`) instead of silently replaying
- **list_timeline_events** (`#207`): already returns empty for unknown event types; regression tests added to lock that behavior

## Test plan

- [ ] `tests/integration/store_exercise_log_device_schema.test.ts` — 5 tests, exercise_log and device schema identity resolution
- [ ] `tests/integration/store_unknown_fields_list.test.ts` — 3 tests, unknown_fields array in store response
- [ ] `tests/integration/idempotency_key_content_mismatch.test.ts` — 5 tests, HTTP and MCP mismatch paths
- [ ] `tests/integration/issue_207_list_timeline_events_unknown_type.test.ts` — 3 tests, empty results for unknown event type

All 16 tests pass locally. Type check and lint clean.

### DIFF
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,8 +61,8 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **369**
-- Backend and repo Vitest files: **336**
+- Total automated test files: **373**
+- Backend and repo Vitest files: **340**
 - Frontend Vitest files: **9**
 - Playwright spec files: **24**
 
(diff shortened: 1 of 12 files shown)


castor-agent:
@claude review

PR #222, opened 2026-05-17
fix: detect content-hash mismatch on idempotency key reuse (closes #186)

### LINKED ISSUE #186: Idempotency key silently blocks updates when content has changed
## Issue

When a `store` call is made with the same idempotency key but materially different content, the new observation is silently skipped. The response returns the existing entity snapshot without writing any new data, making it appear the update succeeded when it did not.

## Observed behavior

- First `store` call with `idempotency_key: "home-security-plan-nest-replacement-<PHONE:472f>"` creates the entity with minimal fields
- Second `store` call with the same key but 8 additional fields returns `action: matched_existing` and does not write a new observation
- Entity snapshot remains unchanged — new fields are not present
- No warning or indication that the write was skipped due to idempotency

## Expected behavior

Idempotency keys should deduplicate truly identical writes (same content, same key). When content differs materially, the behavior should be one of:
1. Write a new observation and ignore the idempotency key (treat it as a hint, not a hard lock), or
2. Return a clear error or warning indicating the write was skipped due to a conflicting idempotency key, so the caller knows to use a new key

The current behavior silently loses data, which is particularly dangerous for an append-only system where write integrity is a core guarantee.

## Context

- Environment: claude.ai (web, mobile)
- MCP server: `https://neotoma.markmhendrickson.com/mcp`
- Discovered while attempting to update a `plan` entity with additional fields
- Resolution was to use a new idempotency key (`home-security-plan-nest-replacement-v<PHONE:8677>`)

castor-agent:
## Summary

- Both the MCP path (`storeStructuredInternal`) and the HTTP API path (`storeStructuredForApi`) now compare the SHA-256 hash of the incoming JSON payload against the stored `content_hash` before accepting an idempotency replay. A mismatch throws `ERR_IDEMPOTENCY_COLLISION` instead of silently returning stale data.
- MCP path: removed the early-return replay block and delegates collision detection to `storeRawContent`, then surfaces the error as `McpError(InvalidParams)`.
- HTTP API path: changed the source lookup to also select `content_hash`; computes the incoming hash before the early-return; added a 409 handler in the catch block.
- CLI `store-turn`: when `--idempotency-key` is provided without `--turn-key`, derives `turnKey` from the idempotency key so repeated calls with the same key produce identical entity content (required for replay).
- `openapi.yaml`: adds a 409 `ERR_IDEMPOTENCY_COLLISION` response to `/store`.
- Regression test added: `tests/integration/idempotency_collision.test.ts` covers same-content replay (passes), API-path collision (throws), and MCP-path collision (throws).

## Test plan

- [ ] Run `npm run type-check && npm test` — only pre-existing failures (cursor_hook_stop_backfill, hook_failure_hint, submit_issue_dx, package_scripts, external_actor_badge, agent_badge_tier_icon) should remain.
- [ ] Run `npx vitest run tests/integration/idempotency_collision.test.ts` — all 3 tests pass.
- [ ] Run `npx vitest run tests/cli/cli_store_commands.test.ts` — all 24 tests pass (including the replay-safe store-turn test).

### DIFF
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,18 +61,18 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **360**
-- Backend and repo Vitest files: **327**
+- Total automated test files: **375**
+- Backend and repo Vitest files: **342**
 - Frontend Vitest files: **9**
 - Playwright spec files: **24**
 
 ### Suite counts
 | Suite | Files |
 |---|---:|
-| Vitest unit tests | 87 |
+| Vitest unit tests | 89 |
 | Vitest service tests | 33 |
-| Source-adjacent tests | 39 |
-| Vitest integration tests | 94 |
+| Source-adjacent tests | 45 |
+| Vitest integration tests | 101 |
 | Vitest CLI tests | 55 |
 | Vitest contract tests | 10 |
 | Vitest security tests | 1 |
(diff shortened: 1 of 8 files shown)


castor-agent:
@claude review

PR #224, opened 2026-05-17
fix: buffer raw_fragment writes until after entity resolution (closes #163)

### LINKED ISSUE #163: Interpretation path writes raw_fragments before entity resolution — orphaned fragments on CanonicalNameUnresolvedError
## Summary

During interpretation, unknown fields are written to \`raw_fragments\` **before** entity resolution is attempted (lines 363–492 in \`src/services/interpretation.ts\`). If entity resolution subsequently fails with \`CanonicalNameUnresolvedError\`, the loop \`continue\`s (line 534) — leaving those raw_fragments rows permanently orphaned with \`entity_id = NULL\`.

## Root cause

The interpretation loop processes each extracted entity in this order:

1. Validate fields against schema → split into \`validFields\` / \`unknownFields\`
2. **Write \`unknownFields\` to \`raw_fragments\`** (lines 363–492) — no \`entity_id\` at this point
3. Attempt entity resolution (lines 520–537)
4. If resolution fails → \`continue\` — raw_fragments rows already written, entity_id never backfilled

Relevant code path:

\`\`\`typescript
// interpretation.ts:415-428 — written before entity resolution
await db.from("raw_fragments").insert({
  source_id: sourceId,
  interpretation_id: interpretationId,
  entity_type: entityType,
  fragment_key: key,
  fragment_value: value,
  // entity_id omitted — not yet resolved
});

// interpretation.ts:526-534 — resolution fails, loop continues
} catch (err) {
  if (err instanceof CanonicalNameUnresolvedError) {
    logger.warn(...);
    continue; // raw_fragments already written above
  }
}
\`\`\`

## Impact

- Raw_fragments rows accumulate with \`entity_id = NULL\` and no path to promotion
- They are invisible to \`auditEntityType\` / \`repairEntityType\` (which require entity_id to create observations)
- They inflate raw_fragments indefinitely — frequency counts increment on re-runs but the data is permanently stranded
- Confirmed in prod: Reddit post fragments (source ingestion where canonical_name was unresolvable) produced hundreds of entity-less fragments across the \`post\` entity type

## Possible fixes

**Option A — Write raw_fragments after entity resolution**: Move the unknown-fields write loop to after the \`resolveEntity\` call succeeds, passing \`entity_id\` into the insert. Trade-off: fragments are lost entirely if resolution fails (no debug artifact).

**Option B — Backfill entity_id post-resolution**: Write fragments without entity_id as now, but after successful resolution update any fragments from this source/interpretation with the resolved entity_id. Preserves the debug artifact, makes fragments promotable.

**Option C — Mark orphaned fragments explicitly**: Add a \`fragment_envelope.reason = "entity_unresolved"\` sentinel on \`continue\` so they are distinguishable from normal unknown_field fragments and can be cleaned up or retried separately.

Option B preserves the most data fidelity and is backward-compatible with existing entity-less fragments.

## Related

- \`src/services/interpretation.ts\` lines 363–492 (unknown field write), 520–537 (entity resolution / skip)
- \`src/services/raw_fragments.ts\` — \`entity_id\` is conditionally spread, so NULL is structurally allowed

castor-agent:
## Summary

- Fragment DB writes in the interpretation loop were issued **before** `resolveEntity()`. When `CanonicalNameUnresolvedError` was thrown (missing canonical name field), the loop `continue`d to the next entity but left orphaned `raw_fragments` rows in the DB with no entity or observation link.
- Fix: collect `pendingConvertedFragments` and `pendingUnknownFragments` in-memory during field validation, then flush via `flushPendingFragments()` only after both entity resolution and observation creation succeed. On `CanonicalNameUnresolvedError` the buffers are discarded — no rows written.

## Test plan

- [ ] New regression test `tests/integration/interpretation_fragment_ordering.test.ts` — 3 cases:
  1. Missing canonical name field → `raw_fragments` count remains 0
  2. Present canonical name field → `raw_fragments` written normally (unknown field stored)
  3. Mixed batch → only the successfully resolved entity's fragments are written
- [ ] All 3 new tests pass locally
- [ ] Full unit test suite passes (pre-existing failures in cursor_hook, submit_issue_dx, external_actor_badge are unrelated to this change)

### DIFF
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,8 +61,8 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **366**
-- Backend and repo Vitest files: **333**
+- Total automated test files: **367**
+- Backend and repo Vitest files: **334**
 - Frontend Vitest files: **9**
 - Playwright spec files: **24**
 
@@ -72,7 +72,7 @@ flowchart TD
 | Vitest unit tests | 87 |
 | Vitest service tests | 33 |
 | Source-adjacent tests | 45 |
-| Vitest integration tests | 94 |
+| Vitest integration tests | 95 |
 | Vitest CLI tests | 55 |
 | Vitest contract tests | 10 |
 | Vitest security tests | 1 |
@@ -293,7 +293,7 @@ flowchart TD
 **Runner:** `vitest`
 **Command:** `npm run test:integration` or `npx vitest run tests/integration`
 **Requirements:** Database configured; remote-dependent subsets additionally need `RUN_REMOTE_TESTS=1`.
-**Files (94):**
+**Files (95):**
 - `tests/integration/aauth_attribution_stamping.test.ts`
 - `tests/integration/aauth_resource_metadata.test.ts`
 - `tests/integration/aauth_revocation_e2e.test.ts`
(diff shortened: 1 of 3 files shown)


castor-agent:
@claude review

PR #225, opened 2026-05-17
fix: surface no-schema entity types in store response (closes #164)

### LINKED ISSUE #164: No-schema fallback in interpretation writes entity-less raw_fragments with no promotion path
## Summary

When the interpretation pipeline encounters an entity type with no schema, it writes the entire extracted payload as a single \`raw_fragments\` row with \`reason: "no_schema"\` and no \`entity_id\` (interpretation.ts lines 261–276). This predates the auto-enhancement pipeline and should now trigger schema creation instead.

## Root cause

\`\`\`typescript
// interpretation.ts:261-274
await db.from("raw_fragments").insert({
  id: fragmentId,
  source_id: sourceId,
  interpretation_id: interpretationId,
  user_id: userId,
  entity_type: entityType,
  fragment_key: "full_entity",
  fragment_value: currentEntityData,
  fragment_envelope: {
    reason: "no_schema",
    entity_type: entityType,
  },
  // entity_id omitted — no entity resolution attempted
});
\`\`\`

Entity resolution is never attempted, so \`entity_id\` is always NULL. The \`full_entity\` fragment key is synthetic — it doesn't correspond to any real field and won't match schema declarations if a schema is later added manually.

## Expected behavior

When an entity type with no schema is encountered, the interpretation path should:

1. **Infer a schema** from the extracted field set (field names + value types), using the same type-inference logic as the auto-enhancement eligibility check
2. **Register and activate it** immediately via \`schemaRegistry.register()\` with \`activate: true\`
3. **Continue the interpretation run** against the newly registered schema — validating fields, resolving the entity, and writing a proper observation

If schema registration fails or the field set is too sparse to produce a useful schema, fall back to the existing \`full_entity\` fragment as a debug artifact (with a clearer envelope reason like \`"no_schema_registration_failed"\`).

## Why now

The auto-enhancement pipeline (\`schema_recommendation.ts\`, \`auto_enhancement_processor.ts\`) already has the machinery to infer types from fragment values and call \`updateSchemaIncremental\`. The no-schema case is a synchronous version of the same problem — the difference is that here we have the full field set available immediately, so we don't need to wait for frequency thresholds.

## Impact of current behavior

- Entity types seen for the first time produce permanently entity-less, unpromotable fragments
- Data is inaccessible until a user manually registers a schema and re-submits the source
- Frequency count increments on re-runs but the payload is overwritten — misleading signal

## Related

- \`src/services/interpretation.ts\` lines 255–277
- \`src/services/schema_recommendation.ts\` — type inference logic reusable here
- Issue #163 — related: entity_id NULL due to resolution failure (different root cause, same symptom)

castor-agent:
## Summary

- When interpretation encountered an unknown entity type with no schema, the entire payload was silently written as a `raw_fragments` row with `reason: "no_schema"`. Callers had no way to discover which entity types were affected or take corrective action.
- Add `no_schema_entity_types: string[]` to `StoreStructuredResponse` in `openapi.yaml`; regenerate types.
- Introduce `schemaCreationAttempted` flag: when `ensureSchemaForExtractedEntity` is called but returns null (too-sparse field set), the envelope reason is now `"no_schema_registration_failed"` instead of the generic `"no_schema"`.
- Collect distinct no-schema entity types into `noSchemaEntityTypes[]` in `InterpretationResult`; surface as `no_schema_entity_types` in all three store response paths (MCP `create_interpretation`, MCP `store` interpretation branch, HTTP `/interpretations/create`).

## Test plan

- [ ] 4 new regression tests in `tests/integration/interpretation_no_schema_fallback.test.ts`:
  1. `noSchemaEntityTypes` populated; reason is `"no_schema_registration_failed"` when auto-registration attempted but failed
  2. Normal entity types (with schema) not included in `noSchemaEntityTypes`
  3. Mixed batch: only the no-schema type listed
  4. `create_schema_for_unknown: false`: reason stays `"no_schema"`
- [ ] All 4 new tests pass locally
- [ ] Full test suite passes (pre-existing failures unrelated to this change)

### DIFF
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,8 +61,8 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **367**
-- Backend and repo Vitest files: **334**
+- Total automated test files: **369**
+- Backend and repo Vitest files: **336**
 - Frontend Vitest files: **9**
 - Playwright spec files: **24**
 
@@ -72,7 +72,7 @@ flowchart TD
 | Vitest unit tests | 88 |
 | Vitest service tests | 33 |
 | Source-adjacent tests | 45 |
-| Vitest integration tests | 95 |
+| Vitest integration tests | 96 |
 | Vitest CLI tests | 55 |
 | Vitest contract tests | 10 |
 | Vitest security tests | 1 |
@@ -294,7 +294,7 @@ flowchart TD
 **Runner:** `vitest`
 **Command:** `npm run test:integration` or `npx vitest run tests/integration`
 **Requirements:** Database configured; remote-dependent subsets additionally need `RUN_REMOTE_TESTS=1`.
-**Files (95):**
+**Files (96):**
 - `tests/integration/aauth_attribution_stamping.test.ts`
 - `tests/integration/aauth_resource_metadata.test.ts`
 - `tests/integration/aauth_revocation_e2e.test.ts`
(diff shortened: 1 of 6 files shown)


castor-agent:
@claude review

PR #260, opened 2026-05-19
fix(#259): promote plan.body from raw_fragments on schema extension

### LINKED ISSUE #259: plan.body and other fields stored before schema was active land in raw_fragments — need migrate_existing backfill
## Problem

When a `plan` entity is stored and the active `plan` schema does not yet include `body` (or any other field), those field values are silently routed to `raw_fragments` instead of observations. They never appear in the entity snapshot. The caller sees a stored entity with `body` missing even though the value was supplied.

This affects any user whose production database had a `plan` schema row written before `body` was added to `seed_schema.ts`, or before `updateSchemaIncremental` was called to extend an older schema row. The same class of issue applies to any entity type whose schema was iteratively extended after initial data was stored.

## Root cause

Fields not present in the active schema at store time are written to `raw_fragments`. There is no automatic promotion when the schema is later extended — the data stays in `raw_fragments` until `migrate_existing: true` is passed to `updateSchemaIncremental` for the affected field names.

`seedPlanSchema` (boot-time seeding) calls `updateSchemaIncremental` without `migrate_existing: true`, so the historical `body` raw_fragments are never promoted even after the schema is corrected.

## Expected behavior

After v0.13.0 ships and `seedPlanSchema` runs on boot, any previously stored `plan.body` values that are sitting in `raw_fragments` should be promoted to observations so they appear in the snapshot.

## Proposed fix

1. Add `migrate_existing: true` to the `updateSchemaIncremental` call in `seedPlanSchema` when missing fields are detected. This is a one-time backfill per field per instance.
2. Consider adding a CLI command (e.g. `neotoma schemas migrate-raw-fragments --entity-type plan`) so users can trigger the backfill explicitly without a server restart.
3. Add a regression test that stores a `plan` with `body`, simulates the field being absent from the schema at store time (raw_fragment path), then calls `updateSchemaIncremental` with `migrate_existing: true` and asserts `body` appears in the snapshot.

## Affected fields

All fields added to `PLAN_FIELDS` in `src/services/plans/seed_schema.ts` after a user's initial schema registration, most commonly:
- `body`
- `public_body`
- `public_overview`
- `plan_kind`
- `decision_required`
- `decision_blockers`
- `worktree_path`

## Related

- #238 (closed): `updateSchemaIncremental` dropped `canonical_name_fields` — fixed in current code; the `preserved` spread is in place.
- #185 / #187: `unknown_fields` list and `hint` now surface at store time so agents know which fields were dropped — but this only helps going forward, not for already-stored data.

castor-agent:
## Summary

- **Root cause**: `plan.body` (and any field absent from the schema at store time) routes to `raw_fragments` and is never promoted to observations or reflected in the entity snapshot. The boot-time schema seeder called `updateSchemaIncremental` for missing fields but without `migrate_existing: true`, so historical raw_fragments were never backfilled.
- **Secondary bug exposed by tests**: The `updateSchemaIncremental` MCP handler passed `user_id: undefined` for global schemas (`user_specific: false`), causing `migrateRawFragmentsToObservations` to filter only for `null`/default-UUID rows — missing raw_fragments stored by the authenticated user.

## Changes

**`src/services/plans/seed_schema.ts`**
- Added `migrate_existing: true` to the boot-time `updateSchemaIncremental` call so every server start that detects missing plan fields also backfills historical raw_fragments for those fields.

**`src/services/schema_registry.ts`**
- Added `migrate_user_id?: string` option to `updateSchemaIncremental`. When present, it overrides `options.user_id` for the migration query. Allows global-schema migrations to target raw_fragments stored by the authenticated caller.

**`src/server.ts`**
- MCP handler now passes `migrate_user_id: userId` whenever `migrate_existing` is true, regardless of `user_specific`. This ensures the authenticated user's raw_fragments are included in the backfill even for global schemas.

**`src/services/plans/seed_schema.test.ts`**
- Extended existing "incrementally adds missing fields" test to assert `migrate_existing: true`.
- Added new test: "passes migrate_existing: true when body is among the missing fields" (simulates a DB with an older plan schema that predates `body`).

**`tests/integration/plan_body_raw_fragment_backfill.test.ts`** (new)
- 3 integration tests covering the full path:
  1. Store with body when schema lacks the field → body in raw_fragments, absent from snapshot.
  2. `updateSchemaIncremental` with `migrate_existing: true` → observation created, snapshot updated.
  3. Idempotent: re-running migration doesn't duplicate or corrupt the body value.

## Test plan

- [x] `src/services/plans/seed_schema.test.ts` — 4/4 unit tests pass
- [x] `tests/integration/plan_body_raw_fragment_backfill.test.ts` — 3/3 integration tests pass
- [x] `npm run type-check` — clean
- [x] `npm test` — 3323 passed, 6 failures are pre-existing (same set as before this change)
- [x] `npm run validate:test-catalog` — clean

Closes #259

### DIFF
--- a/docs/releases/in_progress/v0.13.0/github_release_supplement.md
+++ b/docs/releases/in_progress/v0.13.0/github_release_supplement.md
@@ -58,6 +58,54 @@ v0.13.0 is a substantial feature release. The headline additions are **schema-le
 - **Untrusted `X-Forwarded-For` IPs are redacted in log lines.** `isLocalRequest` log output no longer leaks raw client IPs from spoofed headers (PII hardening).
 - **Neotoma entity IDs carved out of phone-number PII redaction.** Entity IDs of the form `ent_*` are no longer mistakenly matched as phone numbers (#130, #134).
 
+## MCP agent instructions changes
+
+`docs/developer/mcp/instructions.md` received significant additions in this release. The file is the behavioral contract the Neotoma MCP server sends to clients at runtime, so these changes take effect without any code changes on the agent side — as soon as the server is updated, connected agents receive the updated instructions.
+
+### New mandatory rules
+
+- **Image-only message compliance.** When the user message contains only an image or screenshot (no text), all five turn-lifecycle steps still MUST execute. Agents must store a `conversation_message` describing the image and extract any entities visible in it (tasks, events, contacts, transactions). (#93, #95)
+
+- **Short-pass compliance.** Doc-only edits, `neotoma cli config`, single-file lint tweaks, and other minimal turns are explicitly NOT exempt from steps 1–5. Rationalizing a store skip as "analysis-only", "in-session bookkeeping", or "just an evaluation" is FORBIDDEN. (#93)
+
+- **PII stripping checklist before issue filing.** A mandatory six-point checklist must be completed before every `submit_issue` or `add_issue_message` call: strip names/emails/phones, replace private identifiers with `ent_*` IDs or generic labels, redact credentials and verbatim sensitive data, and re-read the body before calling. (#139)
+
+- **Issue entity linking after filing.** Immediately after `submit_issue` returns, agents MUST call `create_relationships` with REFERS_TO edges to every Neotoma entity that motivated or is referenced by the issue. Likewise for `add_issue_message`. Previously these edges were optional. (#131, #135)
+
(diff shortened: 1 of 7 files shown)


castor-agent:
@claude review

castor-agent:
@claude review

PR #335, opened 2026-05-20
fix(plans): resolve #325, #262, #326 — reliable plan body field usage

castor-agent:
## Summary

- **#325** — Updated body field description in seed_schema.ts to make clear it is the primary long-form content field, and that agents must not split content into overview or goals
- **#262** — renderProfileEntity now strips a leading "## body" / "# body" heading from content before writing to the mirror output; harnesses like Claude Code store the plan body with a literal "## body\n\n" prefix that produced a spurious section in rendered mirrors
- **#326** — storeStructuredForApi (HTTP API store path) now validates fields against the active schema per entity and calls storeUnknownFields inside the commit loop, so the response includes unknown_fields_count, unknown_fields, and hint — parity with the MCP store_structured path

## Test plan

- [ ] tests/unit/plan_schema_body_field.test.ts — verifies the body field description mentions primary content, warns against overview/goals, and mentions the mirror
- [ ] tests/unit/canonical_markdown_body_heading.test.ts — 6 cases covering heading stripping (heading stripped, single-hash variant, case-insensitive, embedded non-leading heading is preserved, clean body passthrough, content_only mode untouched)
- [ ] tests/integration/api_store_unknown_fields.test.ts — 4 cases: all-known fields returns count=0/[]/no hint; unknown fields returns non-zero count plus sorted list; unknown fields produces hint; no-schema entity type returns count=0

Generated with Claude Code

Issue #326, opened 2026-05-20
no server-side enforcement that `list_entity_types` is consulted before `store`


castor-agent:
## Problem

The `store` tool accepts arbitrary field names and silently bins unknown ones into `raw_fragments`, reporting `unknown_fields_count` in the response. This creates a failure mode where agents store entities with entirely wrong field names (e.g. `name` instead of `title`, `description` instead of `body`) and receive no hard error — only a numeric count that may go unnoticed.

There is no mechanism that encourages or enforces schema consultation before storage. The burden falls entirely on the agent's behavioral rules and the quality of field descriptions.

## Impact

- Entities stored with `unknown_fields_count > 0` have their content silently dropped into unstructured fragments instead of indexed, queryable fields.
- Agents do not naturally re-read schema descriptions between sessions; without a rule or server-side signal, the wrong fields are used repeatedly.
- The `plan` entity is particularly affected because `body` (primary long-form content) is skipped in favor of `overview` or ad-hoc field names.

## Suggested improvements

1. **Stronger response signal**: When `unknown_fields_count > 0`, include the list of unknown field names in the response (e.g. `unknown_field_names: ["name", "description"]`) so agents can immediately identify what to correct without re-reading the schema.

2. **Schema hints in store response**: When `unknown_fields_count > 0`, include a `hint` pointing to the correct field names — similar to the tightening-change hint obligation for HTTP endpoints (`docs/subsystems/errors.md`).

3. **Optional strict mode**: A `strict: true` parameter on `store` that returns an error instead of silently binning unknown fields, for callers that want to catch schema drift at write time.

4. **`primary_content_field` annotation on schema**: A machine-readable annotation identifying which field holds the primary long-form content for an entity type, so agents can identify it programmatically rather than relying on description text alone.

Issue #262, opened 2026-05-19
Mirror profile renderer emits spurious `## body` section heading when content_field is "body"


castor-agent:
## Summary

When a mirror profile uses `render_mode: "frontmatter_content"` with `content_field: "body"`, the rendered markdown file contains a `## body` section heading immediately before the body content rather than rendering the content directly as the document body.

## Expected behavior

The `body` field should be rendered as the document content (no heading), with all other fields either in frontmatter or as named `## <field>` sections per the profile config. The `## body` heading itself should not appear.

## Actual behavior

The rendered file contains:

```markdown
## body

## Problem

When no bearer token auth is configured...
```

The `## body` label leaks into the output as if `body` were an ordinary named field rather than the document content field.

## Context

- Profile: `neotoma-plans` in `getMirrorConfig()` defaults (`src/services/canonical_mirror.ts`)
- Config: `render_mode: "frontmatter_content"`, `content_field: "body"`
- The `body` field contains markdown with its own `##` headings, so the spurious `## body` header creates a heading hierarchy mismatch
- Additionally, structured fields (`overview`, `goals`, `architecture`, `next_steps`) render as named sections even when their content is already present in `body`, causing redundancy

## Likely fix

In the profile renderer, when `content_field` is set, skip emitting the `## <content_field>` heading and render the field value directly. Also consider suppressing fields whose content is subsumed by `body` when `render_mode: "frontmatter_content"` is active (e.g. via an `omit_fields_in_body_mode` list on the profile).

PR #383, opened 2026-05-21
fix(plans): close #325 (plan body description) + #262 (mirror heading strip)

### LINKED ISSUE #325: plan entity: `body` field description insufficient to prevent agents from omitting it
## Problem

When storing `plan` entities, agents consistently omit the `body` field and instead split content across `overview` + `goals`. This happens even when `list_entity_types` is consulted, because the `body` field description does not make clear enough that it is the **primary long-form content field** — the correct destination for full plan markdown, spec content, or structured prose.

The `overview` field name is more intuitively "correct" to agents, and without a strong signal in the schema that `body` supersedes it for full content, the wrong field is chosen.

## Expected behavior

The `body` field description for `plan` entities should explicitly state:
- This is the primary long-form content field.
- Full plan markdown, spec content, or structured prose goes here.
- `overview` is for short narrative summaries only and is not a substitute for `body`.

## Observed behavior

Agents store full spec markdown in `overview` and `goals`, leaving `body` empty. The entity snapshot is technically valid (`unknown_fields_count: 0`) but semantically incomplete — the primary content field is unused.

## Suggested fix

Update the `plan` schema's `body` field description to read something like:

> Primary long-form content field. Store the full plan markdown, spec, or prose here. Do not split this content across `overview` or `goals` — those fields are for short summaries and structured sub-content only.

Consider also adding a `preferred_field` or `primary_content_field` annotation to the schema definition so tooling and agents can identify the correct destination without relying solely on the description text.

markmhendrickson:
Closes #325 and #262. Splits the trivially-correct parts out of [PR #335](https://github.com/markmhendrickson/neotoma/pull/335) so they can ship in v0.14.0 without depending on the contested #326 unknown-fields tracking work (which has unresolved review blockers).

### #325 — Plan `body` field description

Rewrites \`src/services/plans/seed_schema.ts\` \`body\` field description to:
- Explicit that it is the primary long-form content field.
- Content must not be split across \`overview\` or \`goals\` — those are for short summaries / structured sub-items only.
- Documents the \`## body\` heading-strip behavior (so agents are not surprised).

### #262 — Mirror \`## body\` heading strip

In \`renderProfileEntity\` (\`src/services/canonical_markdown.ts\`), strip a leading \`## body\` / \`# body\` heading from the rendered body when using \`frontmatter_content\` render mode. Some harnesses (notably Claude Code) serialize the plan body with a literal \`## body\\n\\n\` prefix, which produces a redundant heading in the markdown mirror.

The strip is intentionally bounded per the #335 review feedback:
- only matches \`#\` or \`##\` (\`###\` and deeper pass through unchanged)
- only at the start of content
- requires trailing newline(s) (so the entire-content-is-just-\`## body\` case preserves the heading)

## Tests

- \`tests/unit/canonical_markdown_body_heading.test.ts\` — 8 tests including 2 boundary cases per #335 review: \`### body\` not stripped, \`## body\` with no trailing newline not stripped.
- \`tests/unit/plan_schema_body_field.test.ts\` — 3 tests locking the description language.

## Not included

The #326 portion of PR #335 (HTTP API store path unknown_fields tracking) is intentionally NOT in this PR. That change has unresolved blockers from the [2026-05-21 review](https://github.com/markmhendrickson/neotoma/pull/335#issuecomment):

1. **Double-write bug**: \`createObservation\` is called with the full payload (including unknown keys) AND \`storeUnknownFields\` is called on the same keys — same data persists in two places, breaking the \"raw_fragments exclusively\" contract that the MCP path enforces.
2. **Identity-hoist stripping omitted**: \`name\`, \`full_name\`, \`title\`, \`email\`, \`url\`, \`canonical_name\` should be stripped from \`unknown_fields\` before counting, matching the MCP path's behavior.
3. **Schema registry loaded 2-3× per entity in hot loop**.

Those need real rework. Recommend closing PR #335 in favor of:
- This PR (#325 + #262)
- A separate focused PR for #326 that factors the validate/strip-identity-hoist/split-validFields-vs-unknowns sequence into a shared helper between MCP and HTTP paths.

## Verification

- [x] \`npm run type-check\` clean.
- [x] \`npm run format:check\` clean.
- [x] \`npx vitest run tests/unit/canonical_markdown_body_heading.test.ts tests/unit/plan_schema_body_field.test.ts\` — 11/11 pass.

### DIFF
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,15 +61,15 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **395**
-- Backend and repo Vitest files: **362**
+- Total automated test files: **397**
+- Backend and repo Vitest files: **364**
 - Frontend Vitest files: **9**
 - Playwright spec files: **24**
 
 ### Suite counts
 | Suite | Files |
 |---|---:|
-| Vitest unit tests | 96 |
+| Vitest unit tests | 98 |
 | Vitest service tests | 33 |
 | Source-adjacent tests | 46 |
 | Vitest integration tests | 108 |
@@ -107,7 +107,7 @@ flowchart TD
 **Runner:** `vitest`
 **Command:** `npm test -- tests/unit`
 **Requirements:** Basic `.env` if required by the module under test.
-**Files (96):**
+**Files (98):**
 - `tests/unit/aauth_admission.test.ts`
 - `tests/unit/aauth_attestation_apple_se.test.ts`
 - `tests/unit/aauth_attestation_revocation.test.ts`
@@ -130,6 +130,7 @@ flowchart TD
 - `tests/unit/attribution_policy.test.ts`
 - `tests/unit/bigint_serialization.test.ts`
 - `tests/unit/bundled_docs_nav.test.ts`
+- `tests/unit/canonical_markdown_body_heading.test.ts`
 - `tests/unit/cli_aauth_tbs_attestation.test.ts`
 - `tests/unit/cli_aauth_tpm2_attestation.test.ts`
 - `tests/unit/cli_aauth_yubikey_attestation.test.ts`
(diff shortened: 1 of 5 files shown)


markmhendrickson:
Waiver: The Blocking finding (change_guardrails item 14 — supplement documentation) was self-declared as 'Not merge-blocking on this PR' by the reviewer. The supplement folder now exists (PR #382 merged) and already covers both items: the `## body` heading strip (line 50) and the plan body field description tightening (line 103). No code changes are needed.

PR #392, opened 2026-05-21
fix(search): rename exclude_bookkeeping to exclude_conversations

markmhendrickson:
## Problems

`exclude_bookkeeping` on `retrieve_entities` is opaque — callers can't tell what it excludes without reading docs or source. The name suggests a broad internal concept rather than describing what actually changes.

## Solutions

Renamed to `exclude_conversations` across all layers:

- `openapi.yaml` — request field renamed
- `src/shared/action_schemas.ts` — Zod schema field renamed
- `src/shared/action_handlers/entity_handlers.ts` — TypeScript parameter and interface fields renamed
- `src/actions.ts` — destructuring and pass-through renamed
- `src/tool_definitions.ts` — MCP tool parameter renamed
- `src/shared/openapi_types.ts` — regenerated via `npm run openapi:generate`

`BOOKKEEPING_ENTITY_TYPES` (internal constant) unchanged — it's not part of the public API surface.

## Test plan

- [x] Type check passes (`tsc --noEmit`)
- [x] Lint passes (0 errors)
- [x] `npm run openapi:generate` regenerated `openapi_types.ts` cleanly

## Breaking changes

No breaking changes. `exclude_bookkeeping` / `excludeBookkeeping` was introduced in v0.14.0 (not yet released) and has no external callers.


markmhendrickson:
Reverting per design decision: exclude_bookkeeping is the correct name since the predicate filters all BOOKKEEPING_ENTITY_TYPES (conversation, conversation_message, agent_message), not just conversations. The name will remain accurate when the set grows. Revisit if/when the set diverges from conversation-only types.

PR #398, opened 2026-05-22
feat: add override_validation service for agent_definition field policies

markmhendrickson:
## What this does

Adds `src/services/override_validation.ts` — a write-path guard that enforces per-field access restrictions on `agent_definition` entities using a declarative JSON policy stored in the entity's `override_policy` snapshot field.

This enables individual Ateles swarm daemons to be restricted from overwriting sensitive fields they should never touch (e.g. `agent_grant`, `prompt_markdown`), without requiring a code change per daemon.

## override_policy JSON format

Stored as a string field in the `agent_definition` entity snapshot:

```json
{
  "field_policies": {
    "agent_grant": {
      "allowed_roles": ["operator"],
      "deny_message": "Only operators may change agent grants"
    },
    "prompt_markdown": {
      "allowed_roles": ["operator", "service"],
      "min_tier": "software"
    }
  },
  "default_policy": "allow"
}
```

- `field_policies`: per-field allow-lists with optional `min_tier` and `deny_message`.
- `default_policy`: `"allow"` (default, fail-open) or `"deny"` (allowlist-only mode).
- Invalid or missing JSON → fail-open (write proceeds).

## Role derivation

| Identity | Derived role |
|----------|-------------|
| `null` (no agent context) | `"operator"` |
| AAuth tier `operator_attested` or `hardware` | `"operator"` |
| `sub` ends in `@ateles-swarm` + admission carried `agent_label` | `agent_label` value |
| `sub` ends in `@ateles-swarm`, no label | `"service"` |
| All others | `"service"` |

## Where it hooks in

`enforceOverridePolicy()` is called in two places, immediately after the existing `enforceAttributionPolicy()` + `assertCanWriteProtected()` checks:

- **`src/services/observation_storage.ts`** → `createObservation()`
- **`src/services/correction.ts`** → `createCorrection()`

Both sites pass `db`, `getCurrentAgentIdentity()`, and `getCurrentAAuthAdmission()` so no new request-context plumbing is required.

## Phase 1 scope

Scoped to `entity_type === "agent_definition"` only. All other entity types are passed through immediately without a DB lookup. A later phase can extend the mechanism to other governance entity types.

## Tests

`src/services/__tests__/override_validation.test.ts` covers `checkFieldAllowed` and `parseOverridePolicy` with 10 test cases:

- Role allow / deny
- `deny_message` surfacing
- Multi-role policies
- `min_tier` enforcement and fail-open when tier is unknown
- Valid full policy parsing
- `default_policy` omitted vs. explicit
- Invalid JSON → null
- Non-object JSON → null
- Missing `field_policies` → null
- Malformed field entries silently skipped
- `min_tier` parsed correctly
- `OverridePolicyViolationError` shape + `toErrorEnvelope()`

## Why this matters for the Ateles swarm

Each `agent_definition` entity in Neotoma now acts as its own access-control manifest. Operators can tighten or relax per-daemon field permissions without deploying code — just update the `override_policy` field on the entity. This is the foundation for graduated autonomy: a new daemon starts with a narrow `"deny"`-default policy and earns broader write access as it proves itself reliable.


markmhendrickson:
Superseded by #1634, which cleanly re-derives the net-new `override_validation` service onto current `main`. The original branch's other changes were already on main (security advisories, advisory skill) or stale doc drift (practices.md). The bundled `tests/services/relationships_tenant_isolation.test.ts` is tracked separately as a security finding (its asserted user_id filter is not implemented in main's relationships service).

PR #937, opened 2026-05-25
fix(issues): retry submit_issue as unsigned guest on AUTH_REQUIRED

### LINKED ISSUE #936: submit_issue remote push fails with AUTH_REQUIRED instead of falling back to unsigned guest submission
## Summary

`submit_issue` (MCP tool) stores issues locally correctly but always fails remote submission with `AUTH_REQUIRED` when no AAuth agent grant is configured, even though the remote `/issues/submit` HTTP endpoint has a guest path that accepts unsigned requests.

## Root cause

`neotoma_client.ts` `submitIssueToRemote` builds the client with `signWithCliAAuth: true`. When AAuth signing produces a signature the remote instance cannot verify (no agent grant), the remote returns `AUTH_REQUIRED`. The client throws immediately instead of retrying as an unsigned guest.

The remote `/issues/submit` handler (`actions.ts:8219`) uses `resolveRoutePrincipal(req, ["user", "guest"])` and calls `submitGuestIssue` for payloads that contain `local_issue_id` or `submission_timestamp` — both of which `submitIssueToRemote` already sends. An unsigned retry would succeed via the guest path, return a `guest_access_token`, and mark the issue `submitted_to_neotoma: true`.

## Fix

In `submitIssueToRemote`, when the first attempt returns `AUTH_REQUIRED`, retry immediately with an unsigned client (`signWithCliAAuth: false`). This is safe because:
- The guest path on the remote is rate-limited and policy-gated (`submitter_scoped`)
- The issue content is already local regardless of the outcome
- The retry only runs when AAuth verification explicitly fails — not on network errors or other 4xx/5xx responses

## Impact

Every `submit_issue` call from a local MCP session without an AAuth agent grant results in `sync_pending=true`, even though the remote would accept the submission as a guest. The agent reports AUTH_REQUIRED as a configuration problem when it isn't one — the remote guest path is open.

## Proposed change

`src/services/issues/neotoma_client.ts` — in the `AUTH_REQUIRED` branch of `submitIssueToRemote`, instead of throwing, create an unsigned client and retry the POST. Throw only if the retry also fails.

markmhendrickson:
## Summary

- Closes #936
- When `submitIssueToRemote` receives `AUTH_REQUIRED` from the AAuth-signed POST to `/issues/submit`, retry once with an unsigned client so the remote's guest path can accept the submission.
- Guest fallback is safe here: the payload already carries `submission_timestamp` (and `local_issue_id` for non-GitHub-backed issues), which the remote handler routes to `submitGuestIssue` via `resolveRoutePrincipal(req, [\"user\", \"guest\"])`. The retry is skipped when `NEOTOMA_CLI_AAUTH_DISABLE=1` (signed call already unsigned).

## Test plan

- [x] `npx vitest run src/services/issues/neotoma_client.test.ts` — 12/12 pass, including two new cases: AUTH_REQUIRED retry succeeds; AUTH_REQUIRED retry also fails surfaces the hint
- [x] `npm run type-check`

### DIFF
--- a/src/services/issues/neotoma_client.test.ts
+++ b/src/services/issues/neotoma_client.test.ts
@@ -114,6 +114,67 @@ describe("Neotoma Issue Client", () => {
       expect(String(request.body.local_issue_id)).toMatch(/^local:test\/repo:/);
     });
 
+    it("retries unsigned when the signed submission returns AUTH_REQUIRED", async () => {
+      const signedPost = vi.fn().mockResolvedValueOnce({
+        error: { error_code: "AUTH_REQUIRED", message: "agent grant missing" },
+      });
+      const unsignedPost = vi.fn().mockResolvedValueOnce({
+        data: {
+          entity_ids: ["issue-entity-2", "conversation-entity-2"],
+          issue_entity_id: "issue-entity-2",
+          conversation_id: "conversation-entity-2",
+          guest_access_token: "guest-token-after-retry",
+        },
+      });
+      vi.mocked(createApiClient)
+        .mockReturnValueOnce({
+          POST: signedPost,
+          GET: vi.fn(),
+        } as unknown as ReturnType<typeof createApiClient>)
+        .mockReturnValueOnce({
+          POST: unsignedPost,
+          GET: vi.fn(),
+        } as unknown as ReturnType<typeof createApiClient>);
+
+      const result = await submitIssueToRemote({
+        title: "Issue",
+        body: "Body",
+        visibility: "public",
+      });
+
+      expect(result.issue_entity_id).toBe("issue-entity-2");
+      expect(result.access_token).toBe("guest-token-after-retry");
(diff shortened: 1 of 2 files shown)


markmhendrickson:
@claude

markmhendrickson:
Hi @markmhendrickson — I'm here. Let me know if you have questions or feedback on the change, or I can walk through the implementation.

**Summary of what this PR does:**
- In `submitIssueToRemote`, when the AAuth-signed POST to `/issues/submit` returns `AUTH_REQUIRED`, the code now retries once with an unsigned client (`signWithCliAAuth: false`) so the remote's guest path can accept the submission.
- If the unsigned retry also fails, it surfaces the original `AUTH_REQUIRED` error with the actionable hint.
- The retry is skipped entirely when `NEOTOMA_CLI_AAUTH_DISABLE=1` (signing is already disabled in that case).

PR #947, opened 2026-05-26
fix(reducer): apply null correction to clear date fields from snapshot (#41)

markmhendrickson:
## Summary

- Fixes the snapshot reducer to clear date fields when null is explicitly set
- Restores all `test:remote:critical` suite to passing (146/146)
- Adds test coverage for null-clear across `mergeArray`, `most_specific`, and `defaults` resolution paths
- Documents null-clear semantics for restoration and priority

## Test plan

- [ ] `test:remote:critical` passes (146/146)
- [ ] Null-clear for date fields verified in snapshot reducer

### DIFF
--- a/docs/subsystems/reducer.md
+++ b/docs/subsystems/reducer.md
@@ -156,6 +156,10 @@ function mergeArrays(
   
   for (const obs of observations) {
     const value = obs.fields[field];
+    // Skip observations that did not carry the field (undefined) and
+    // null contributions (a null entry is not a meaningful array element).
+    // A field-level clear via `field: null` is handled separately by the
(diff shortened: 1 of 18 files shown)


markmhendrickson:
Closing as superseded (PR-salvage pass). Issue #41 is already fixed on `main`: `src/reducers/observation_reducer.ts` already treats `null` as an explicit clear (filters on `obs.fields[field] !== undefined` only) and carries the verbatim comment "Treat null as an explicit clear. Only undefined means the observation did not carry this field...". The reducer projection tests are identical on main, and this branch's other changes (source-existence validation in `interpretation.ts`, freeform relationship-metadata canonicalization) are also already present on `main` line-for-line. Nothing net-new remains to land.

Issue #1467, opened 2026-05-29
Benchmark and document graph-traversal scale limits (retrieve_related_entities / retrieve_graph_neighborhood)


markmhendrickson:
## Context

Raised by a real-world adopter evaluating whether Neotoma's native graph traversal can replace a separate graph database (Neo4j) for relationship-recall queries.

## Current state

- Multi-hop traversal is implemented: `retrieveRelatedEntities()` runs an n-hop BFS up to `max_hops`, tracking visited nodes to prevent cycles (`src/server.ts`); `retrieve_graph_neighborhood` returns a configurable neighborhood.
- Backing store is SQL over `relationship_snapshots` with the BFS loop in memory.
- No stated scale/latency limit is documented anywhere in the codebase.

## Gap

Adopters cannot make a "do we still need a dedicated graph DB?" decision without published traversal latency at representative node/edge counts and hop depths. Today that decision rests on guesswork.

## Proposed work

- Benchmark `retrieve_related_entities` / `retrieve_graph_neighborhood` at representative graph sizes (e.g. 1e4/1e5/1e6 relationships) and hop depths (1–4).
- Document p50/p95 latency and any practical ceiling.
- State the recommended boundary beyond which an external graph index is advisable.

## Why it matters beyond this adopter

A credible "you may not need a separate graph DB" story (with numbers) is a meaningful adoption lever and lets products delete an entire drift surface.

markmhendrickson:
Resolved by #1480 (initial scale benchmark + indexes) and #1513 (fan-out / high-degree-hub benchmark). docs/performance/graph_traversal.md now documents both dimensions with measured numbers: traversal is flat against total table size (the #1467 composite indexes), and fan-out cost scales ~linearly with hub degree (a 20k-degree hub is ~550ms at 2 hops). The 'do we still need a dedicated graph DB?' decision is now answerable from measurements: it's a function of how many high-degree hubs the workload traverses, not of total graph size.

PR #1479, opened 2026-05-30
feat(graph,cli): HTTP multi-hop traversal parity + bulk entity import (#1470)

### LINKED ISSUE #1470: Add bulk import / replay path for large-scale observation backfill
## Context

Raised by a real-world multi-user adopter evaluating Neotoma as a derived, rebuildable memory layer where an external Postgres database is the system of record. In this pattern, Neotoma is rebuilt from Postgres by replaying stored facts via `store` calls when a breaking change lands or a fresh instance is needed.

## Gap

The `store` action accepts an array of entities per call but writes transactionally per observation. There is no documented bulk import path or batch-replay endpoint. For an adopter replaying a production history (thousands of contacts, observations, corrections, timeline events) the only option is one HTTP call per fact, which is prohibitively slow at scale.

## What would close it

- A bulk import endpoint or CLI command that accepts a large observation batch (or a Postgres-sourced export) and writes transactionally in chunks.
- Alternatively, documented guidance on the safe concurrency envelope for parallel `store` calls (how many in-flight at once, back-pressure behavior).
- A CLI command like `neotoma import --source postgres --connection-string ...` that drives the replay from an external source with progress reporting.

## Why it matters beyond this adopter

Any adopter treating Neotoma as a derived/rebuildable layer — rather than the primary system of record — needs a practical rebuild path. Without it, "rebuildable from external source" is a theoretical property, not an operational one. This also affects disaster-recovery scenarios for any Neotoma deployment.

## Related

- Neotoma's deterministic IDs (hash-based entity_id, idempotency_key dedup) mean a replay is safe — the infrastructure to do it at scale is the missing piece.

markmhendrickson:
Two pieces from a real-world multi-user integration backlog, both verified with tests.

## Multi-hop HTTP traversal — corrects a parity gap (and a misstatement)

The MCP handler (`server.ts` `retrieveRelatedEntities`) **already** did breadth-first n-hop traversal, but the HTTP endpoint `POST /retrieve_related_entities` was a *"Simple 1-hop"* stub that silently ignored `max_hops`. Since an HTTP-API consumer would only see 1-hop, "Neotoma does native multi-hop traversal" was only half-true — true on MCP, false on HTTP.

- **Ported the BFS** (level frontier + `visited` set for cycle protection) from the MCP handler into the Express handler, so the HTTP path honors `max_hops` at parity.
- `max_hops` still **defaults to 1** — existing callers are byte-for-byte unaffected. No OpenAPI/contract change (the param was already declared).
- **Fixed a cosmetic bug** in the MCP handler: `hops_traversed` was capped at 1 even when traversal went deeper; it now reports the actual depth reached.
- **Test:** `tests/integration/http_related_entities_multihop.test.ts` seeds `a→b→c→d` with a `d→a` back-edge and asserts depth tracks `max_hops` (1-hop→b, 2-hop→c, 3-hop→d) and the cycle neither loops nor includes the origin.

This is also what unblocks a real **#1467** benchmark — there is now a true n-hop target on both transports to measure.

## Bulk entity import (#1470)

New `neotoma entities import <file>`:
- Reads a JSONL file (one entity object per line), chunks into batched `POST /store` calls (`--batch-size`, default 200).
- Stamps a **stable per-chunk idempotency key** (`--idempotency-prefix`, default derived from file name) so a large backfill can be **replayed safely** without duplicating data.
- **Dry-run by default**; `--commit` to write.
- The `/store` endpoint already writes each batch transactionally; this command owns file reading, chunking, and idempotency — the supported "replay from an external source" path. Pure client-side wrapper over existing `/store`: **no new operationId, no contract change**.
- **Test:** `tests/cli/cli_entities_import.test.ts` covers chunk boundaries, dry-run default, and idempotent re-run.
- Documented in `docs/developer/cli_reference.md`.

## Notes
- Full typecheck clean; CLI command-coverage guard, contract-mapping, MCP graph, and the two new tests all pass (34 tests across touched surfaces).
- Branch off `main`; pre-commit suite passes.
- Commit is unsigned (GPG pinentry could not prompt in the non-interactive environment) — flagging in case signed commits are required on merge.

Closes #1470

### DIFF
--- a/docs/developer/cli_reference.md
+++ b/docs/developer/cli_reference.md
@@ -533,6 +533,12 @@ See `docs/developer/agent_cli_configuration.md` for the rule text and strategy.
   - Preferred: positional `identifier` or `--identifier <id>`
   - Compatibility alias: `--query <id>` (equivalent to `--identifier`)
   - If both `--identifier` and `--query` are provided, they must match
+- `neotoma entities import <file>`: bulk-import entities from a JSONL file (one entity JSON object per line) by chunking them into batched `POST /store` calls.
+  - `--batch-size <n>`: entities per store call (default `200`). Lower it if a request exceeds the server body limit.
+  - `--idempotency-prefix <prefix>`: prefix for the per-chunk idempotency key (default derived from the file name). Re-running with the same prefix and file is a safe no-op, so a large backfill can be replayed without duplicating data.
+  - `--user-id <userId>`: store under a specific user.
+  - `--commit`: actually write. Without it the command runs in dry-run/plan mode.
+  - The `/store` endpoint already writes each batch transactionally; this command owns file reading, chunking, and per-chunk idempotency. For very large backfills, this is the supported "replay from an external source" path.
 
 ### Sources
 
--- a/docs/security/reviews/2026-06-02-pr1479-tenant-isolation.md
+++ b/docs/security/reviews/2026-06-02-pr1479-tenant-isolation.md
@@ -1,0 +1,22 @@
+# Security Review: PR #1479 — Multi-Hop Graph Traversal + Bulk Entity Import
+
+**Verdict:** Tenant isolation holds across all four examined cross-tenant leak vectors; 2 regression tests added, 2 vectors covered by existing tests.
+
+## Vector results
+
+| Vector                 | Hypothesis                                                                 | Result | Enforcing test                                                                                                                          |
+| ---------------------- | -------------------------------------------------------------------------- | ------ | --------------------------------------------------------------------------------------------------------------------------------------- |
+| seed-entity-ownership  | A caller can seed multi-hop traversal from an entity owned by another user | Safe   | existing coverage                                                                                                                       |
+| edge-scoping-escape    | Per-hop BFS expansion follows an edge into another user's subgraph         | Safe   | `tests/security/tenant_isolation_matrix.test.ts` — "planted cross-tenant edge does NOT leak user B's edges or entity via multi-hop BFS" |
+| bulk-import-cross-user | Bulk `/store` import writes entities attributed to another user            | Safe   | `tests/security/tenant_isolation_matrix.test.ts` — "/store bulk import (bulk-import-cross-user)"                                        |
+| unauth-and-guest       | Unauthenticated or guest callers reach the new traversal/import surfaces   | Safe   | existing coverage                                                                                                                       |
+
+## What was probed
+
+The adversarial pass examined four angles against the new traversal and bulk-import surfaces. First, seed-entity ownership: whether a traversal can be rooted at an entity belonging to another tenant. Second, per-hop edge scoping: whether BFS expansion across hops can follow a planted cross-tenant edge into another user's entities or edges. Third, bulk-import cross-user write: whether a batched `/store` import can attribute writes to, or overwrite data owned by, a different user. Fourth, unauthenticated and guest reach: whether these surfaces are accessible without proper authentication or via guest credentials.
+
+For the edge-scoping and bulk-import vectors, an explicit cross-tenant fixture was planted (user B data, user A request) and a regression test added asserting no leak or misattribution. The seed-entity-ownership and unauth/guest vectors were already covered by existing tests in the security suite; those tests were cited and confirmed passing.
+
+## What this is and isn't
+
+This note records an independent adversarial agent pass that converted each examined attack into a durable regression test (or cited existing coverage where present). It is NOT a substitute for human code review. It does not assert the absence of vulnerabilities outside the four vectors listed above, and it does not evaluate non-isolation concerns (correctness, performance, denial of service). Reviewers SHOULD treat the added tests as the enforceable contract for these four vectors and MUST NOT interpret a passing run as blanket security sign-off for the PR.
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,8 +61,8 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **412**
-- Backend and repo Vitest files: **379**
+- Total automated test files: **414**
+- Backend and repo Vitest files: **381**
 - Frontend Vitest files: **9**
 - Playwright spec files: **24**
 
@@ -72,8 +72,8 @@ flowchart TD
 | Vitest unit tests | 102 |
 | Vitest service tests | 33 |
 | Source-adjacent tests | 47 |
-| Vitest integration tests | 112 |
-| Vitest CLI tests | 59 |
+| Vitest integration tests | 113 |
+| Vitest CLI tests | 60 |
 | Vitest contract tests | 12 |
 | Vitest security tests | 3 |
 | Vitest subscription tests | 5 |
@@ -311,7 +311,7 @@ flowchart TD
 **Runner:** `vitest`
 **Command:** `npm run test:integration` or `npx vitest run tests/integration`
 **Requirements:** Database configured; remote-dependent subsets additionally need `RUN_REMOTE_TESTS=1`.
-**Files (112):**
+**Files (113):**
 - `tests/integration/aauth_attribution_stamping.test.ts`
 - `tests/integration/aauth_resource_metadata.test.ts`
 - `tests/integration/aauth_revocation_e2e.test.ts`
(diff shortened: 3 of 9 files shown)


markmhendrickson:
# Security Review: PR #1479 — Multi-Hop Graph Traversal + Bulk Entity Import

**Verdict:** Tenant isolation holds across all four examined cross-tenant leak vectors; 2 regression tests added, 2 vectors covered by existing tests.

## Vector results

| Vector | Hypothesis | Result | Enforcing test |
|--------|-----------|--------|----------------|
| seed-entity-ownership | A caller can seed multi-hop traversal from an entity owned by another user | Safe | existing coverage |
| edge-scoping-escape | Per-hop BFS expansion follows an edge into another user's subgraph | Safe | `tests/security/tenant_isolation_matrix.test.ts` — "planted cross-tenant edge does NOT leak user B's edges or entity via multi-hop BFS" |
| bulk-import-cross-user | Bulk `/store` import writes entities attributed to another user | Safe | `tests/security/tenant_isolation_matrix.test.ts` — "/store bulk import (bulk-import-cross-user)" |
| unauth-and-guest | Unauthenticated or guest callers reach the new traversal/import surfaces | Safe | existing coverage |

## What was probed

The adversarial pass examined four angles against the new traversal and bulk-import surfaces. First, seed-entity ownership: whether a traversal can be rooted at an entity belonging to another tenant. Second, per-hop edge scoping: whether BFS expansion across hops can follow a planted cross-tenant edge into another user's entities or edges. Third, bulk-import cross-user write: whether a batched `/store` import can attribute writes to, or overwrite data owned by, a different user. Fourth, unauthenticated and guest reach: whether these surfaces are accessible without proper authentication or via guest credentials.

For the edge-scoping and bulk-import vectors, an explicit cross-tenant fixture was planted (user B data, user A request) and a regression test added asserting no leak or misattribution. The seed-entity-ownership and unauth/guest vectors were already covered by existing tests in the security suite; those tests were cited and confirmed passing.

## What this is and isn't

This note records an independent adversarial agent pass that converted each examined attack into a durable regression test (or cited existing coverage where present). It is NOT a substitute for human code review. It does not assert the absence of vulnerabilities outside the four vectors listed above, and it does not evaluate non-isolation concerns (correctness, performance, denial of service). Reviewers SHOULD treat the added tests as the enforceable contract for these four vectors and MUST NOT interpret a passing run as blanket security sign-off for the PR.

markmhendrickson:
## Final review status (merged `4abbae5`)

For clarity, since the automated `Claude PR Review` comments above show a review that was **cancelled in-progress** (each was superseded by a force-push under `cancel-in-progress`, so they never reached a Phase-6 verdict — do not read them as a completed review):

The substantive review on this PR is the **independent adversarial tenant-isolation pass** run against the multi-hop traversal + bulk-import surface, recorded in [this comment](https://github.com/markmhendrickson/neotoma/pull/1479#issuecomment-4599577189) and committed as `docs/security/reviews/2026-06-02-pr1479-tenant-isolation.md`. Four cross-tenant leak vectors examined; isolation held on all four; two new regression tests added (`tests/security/tenant_isolation_matrix.test.ts`).

**Status at merge:** `security_gates` ✅, typecheck ✅, tenant-isolation matrix 11/11 ✅. Merged to main. This is an adversarial-agent review, not a human code review.

markmhendrickson:
## Post-merge review — completing the stalled automated review

The automated Claude PR Review on this PR stalled in-progress (cancelled by force-push) and never produced a verdict. This is that review, run against the merged diff.

**Verdict: APPROVED-WITH-NOTES** (0 blocking, 2 advisory, 3 nit)

PR #1479 brings the HTTP POST /retrieve_related_entities endpoint to parity with the MCP handler by porting the breadth-first n-hop traversal (frontier + visited-set cycle protection), and adds a `neotoma entities import <file>` CLI command that chunks a JSONL file into batched, idempotent, transactional POST /store calls. The change is well-scoped and low-risk: the BFS is deterministic (single seed, per-query `.order("relationship_key")`, returned entities re-sorted by `canonical_name,id`, visited-set used only for membership not output order), introduces no `Math.random`/`Date.now` in logic, and every query is tenant-scoped with `.eq("user_id", userId)`. The CLI is a pure client wrapper over the existing `retrieveRelatedEntities`/`store` operationIds — no OpenAPI/contract change, all four body fields (`entities`, `idempotency_key`, `user_id`, `commit`) are declared in the closed `StoreStructuredRequest` schema, dry-run is the safe default, and per-chunk idempotency keys make replay safe (the /store endpoint enforces `ERR_IDEMPOTENCY_MISMATCH` with a structured hint and content-hash check). Test coverage is real: an integration test seeds a cyclic A→B→C→D chain and asserts depth tracks max_hops and the origin/cycle are excluded, and a CLI test exercises chunk boundaries, dry-run default, and stable idempotency keys. Tenant isolation was covered by a separate adversarial pass with regression tests. Docs (cli_reference, test catalog, security review) are in sync. No blocking findings; several advisory/nit accuracy and consistency items below warrant a follow-up but do not gate the merge that already landed.

### Findings

- **[ADVISORY] silent-behavior** — src/actions.ts ~7990-8060 (HTTP /retrieve_related_entities BFS): each per-hop relationship query uses `if (!error && data)` with no `else` branch, so a database error mid-traversal is silently swallowed — the handler returns HTTP 200 with a partial or truncated result and no signal that traversal terminated early due to a query failure. Multi-hop amplifies this versus the old 1-hop stub: a failure at hop N silently caps depth at N-1 while the caller still sees `200 OK`. This matches the pre-existing 1-hop pattern (not a regression), but the multi-hop port is a good moment to surface query errors. Follow-up: log a warning on the error branch (IDs only, no PII) and consider returning a partial-result indicator, or fail the request on a per-hop query error rather than degrading silently.
- **[ADVISORY] contract** — Transport parity gap: the MCP handler (src/server.ts retrieveRelatedEntities) returns `total_entities`, `total_relationships`, and `hops_traversed`, but the HTTP handler (src/actions.ts) returns only `{ relationships, entities }`. The PR's stated goal is HTTP↔MCP parity, and it achieves behavioral parity (max_hops is now honored), but the response shapes still differ. The OpenAPI response is `additionalProperties: true` so this is contract-valid, yet an HTTP caller cannot read back how deep traversal actually went (`hops_traversed`) the way an MCP caller can. Follow-up: add `total_entities`/`total_relationships`/`hops_traversed` to the HTTP JSON response for true parity, and declare them in the openapi.yaml response schema so the field set is discoverable.
- **[NIT] determinism** — src/server.ts:3154 — `hopsTraversed = hop + 1` is assigned unconditionally at the end of each loop iteration, before the `nextLevel.length === 0` break check. On a chain shorter than max_hops, the final iteration that expands the deepest frontier and finds nothing new still increments the counter, so `hops_traversed` over-reports by one whenever traversal terminates by exhaustion (e.g. A→B→C with max_hops=5 reports 3, not 2). The field counts 'hops attempted over a non-empty frontier' rather than 'hops that discovered new nodes' — which the inline comment acknowledges, but it is mildly inconsistent with the field name. This is a cosmetic reporting value (the PR itself frames the prior `Math.min(...)` as a 'cosmetic bug'), so it is non-blocking. Follow-up: only set `hopsTraversed = hop + 1` when `nextLevel.length > 0` (i.e. the hop discovered new nodes), or rename the field's documented semantics.
- **[NIT] style** — src/cli/index.ts (entities import action): the default idempotency prefix is derived from the file basename only (`import:${baseName}` via `file.split('/').pop()`). Two distinct files with the same basename in different directories (e.g. `a/entities.jsonl` and `b/entities.jsonl`) collide on idempotency keys — the second import's chunk 0 hits `ERR_IDEMPOTENCY_MISMATCH` and aborts mid-run, having already committed earlier chunks (cross-chunk import is non-atomic by design). Low likelihood in practice and the `--idempotency-prefix` flag is the documented escape hatch. Follow-up: derive the default prefix from a stable hash of the absolute path (or the full path), and/or document in cli_reference that distinct files must use distinct `--idempotency-prefix` values to avoid mid-import collisions.
- **[NIT] error-handling** — src/cli/index.ts (entities import action): on a chunk failure the command does `throw new Error(`Failed to store chunk ${i}/${n}: ${JSON.stringify(error)}`)`, which serializes the entire server error object into the CLI message. The /store error envelope provides a structured `hint` (e.g. 'Use a new idempotency_key for a different payload.') that would be more actionable than a raw JSON blob, and JSON-stringifying the whole error risks surfacing internal detail. Follow-up: extract and print the envelope's `code` + `hint`/`message` fields rather than `JSON.stringify(error)`.

_Independent post-merge review (agent-run, not human). Findings are follow-ups; none block the already-merged change._

PR #1482, opened 2026-05-30
feat(sse): durable substrate-event log for restart-safe resume (#1464 Tier 2)

### LINKED ISSUE #1464: Add resume/replay cursor to SSE event stream for missed-event recovery
## Context

Raised by a real-world multi-user adopter integrating Neotoma as the memory substrate behind a production web backend, projecting Neotoma relationship/observation events into an external index (e.g. a graph DB).

## Gap

`/subscribe` (POST) and `/events/stream` (GET) exist, but `/events/stream` accepts only `subscription_id` — there is no `cursor`, `from`, `since`, or replay parameter. If a downstream subscriber is offline during a deploy or restart, events emitted in that window are missed with no documented recovery path. For any consumer projecting Neotoma → external store, this silently reintroduces drift — the exact failure class such a projection is meant to eliminate.

## Evidence

- `openapi.yaml`: `/events/stream` query params are limited to `subscription_id`; no cursor/resume semantics.
- No replay/resume handling found in the subscribe handler or event-stream implementation.

## Proposed enhancement

- A monotonic cursor (sequence id or watermark timestamp) on emitted events.
- A `from`/`after_cursor` parameter on `/events/stream` to resume from the last acknowledged position.
- Documented at-least-once delivery + replay semantics for external projectors.

## Why it matters beyond this adopter

Any product using the substrate-event API to maintain an external index needs missed-event recovery to survive normal deploy/restart cycles. This is foundational for the "project Neotoma into X" integration pattern.

markmhendrickson:
Tier 1 (#1480) made a missed-events gap **detectable**. Tier 2 makes the common cases **recoverable** by persisting events, so SSE resume survives a server restart or a gap larger than the in-memory ring.

## The core fix: a stable, restart-surviving cursor

The in-memory ring's id counter **reset to 0 on every restart**, so a client's `Last-Event-ID` from before a restart was meaningless against a fresh ring. This change makes the **durable event-log `seq`** (AUTOINCREMENT, never resets) the event id *everywhere* — ring, SSE frame, and client cursor. That alone is what makes restart-safe resume possible.

## What's added

- **`substrate_events` table**: durable monotonic `seq`, full event payload, indexed on `(user_id, seq)` for resume and `created_at` for pruning.
- **`event_log.ts`**: `persistSubstrateEvent` (synchronous), `getEventsAfterSeq` (durable resume), `hasDurableCursor` (within-retention check), `pruneEventLog`.
- **Synchronous persistence**: events are written in the same path as the ring push — durable the moment they're emitted. SQLite writes are serialized and fast, so the latency cost is negligible; a log failure falls back to the ring id and never blocks delivery.
- **Resume order** in `/events/stream`: in-memory ring (hot) → durable log (if the cursor is still within retention) → a one-time `event: gap` frame **only** when the cursor predates the retained window.
- **Retention**: `NEOTOMA_EVENT_RETENTION_DAYS` (default **7**), pruned on startup and every 6h.

## Tests

`durable_event_log.test.ts` is a round-trip suite (per the destructive/durable-op guardrail): monotonic seq, restart-simulated recovery after the ring is cleared, retention-boundary behavior, prune keeps-inside/drops-outside, and NULL `user_id` handling. The existing 52 subscription/events tests still pass.

## Relationship to #1480

Builds on Tier 1 (#1480) and **supersedes** its gap-only resume seam with the full ring → durable → gap strategy. Best merged after, or rebased onto, #1480.

## Notes
- Full typecheck clean; branch off `main`; pre-commit suite passes.
- Commit is unsigned (GPG pinentry could not prompt non-interactively) — flag if signed commits are required on merge.

Closes #1464

### DIFF
--- a/docs/subsystems/subscriptions.md
+++ b/docs/subsystems/subscriptions.md
@@ -99,10 +99,12 @@ The bridge is wired once at server startup by `installSubscriptionBridge`. The i
 ## SSE delivery contract
 
 - Endpoint: `GET /events/stream?subscription_id=<id>` (auth required; subscription must be owned by the caller). This is the canonical path (registered in `src/actions.ts` and exposed in `openapi.yaml`); the legacy `GET /subscriptions/sse` shorthand was dropped before v0.12.0 — update any older clients.
-- Frame format: `id: <ring_id>\nevent: <event_type>\ndata: <json>\n\n`.
-- Resume: clients should send `Last-Event-ID: <ring_id>` so the hub replays buffered events newer than that id (subject to the ring cap).
-- Buffer eviction and gap detection: when the ring exceeds `NEOTOMA_SSE_EVENT_BUFFER`, the oldest entries are dropped. If a reconnecting client sends a `Last-Event-ID` that is **no longer in the ring** — because it was evicted past the cap, or because the server restarted (the ring is in-memory and does not survive a restart) — the stream emits a one-time `event: gap` frame **before** replaying, with `data` of the form `{ reason: "cursor_not_in_buffer", requested_last_event_id, latest_ring_id, message }`. The hub then resumes from the current buffer head so the client still makes forward progress.
-- Delivery guarantee: SSE resume is **best-effort and non-durable**. Within the in-memory window it is gap-free; outside it (eviction or restart) the `gap` frame signals a discontinuity. A consumer that requires gap-free delivery — for example, projecting substrate events into an external index — must treat a `gap` frame as "reconcile via a full read" rather than trusting the replayed slice. Durable, restart-surviving replay (a persistent event log) is not yet implemented; see issue #1464.
+- Frame format: `id: <seq>\nevent: <event_type>\ndata: <json>\n\n`. The `id` is the durable event-log `seq` (monotonic, AUTOINCREMENT), not the in-memory ring counter — so a client's `Last-Event-ID` is a stable cursor that survives a server restart.
+- Resume order (`Last-Event-ID: <seq>`):
+  1. **In-memory ring** — if the cursor is still buffered, replay from the ring (hot path).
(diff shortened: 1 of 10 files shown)


markmhendrickson:
Rebased onto main and addressed the two durable-resume review advisories in a follow-up commit:

1. **SQL-side narrowing** — `getEventsAfterSeq` now accepts an optional column-backed filter (`event_type`, `entity_id`) pushed into SQL, so a narrow subscription no longer scans the full user log on resume. Payload-only predicates (`entity_type`, peer-loop) remain enforced in JS; the filter only narrows, never widens.
2. **Cursor-ahead-of-head guard** — `hasDurableCursor` now bounds the cursor on both ends `[min-1, max]`. A cursor newer than the durable head (a ring-only id or future seq) previously entered the replay loop, read zero rows, and silently delivered nothing; it now returns false so resume falls through to the in-memory ring.

Added 6 regression tests (SQL narrowing by event_type/entity_id/combined/empty-array; cursor-ahead-of-head; quiet-user; one-below-oldest boundary). Full subscriptions suite: 52 passing.

markmhendrickson:
## Final review status (merged `668c3b4`)

For clarity on review state: the automated `Claude PR Review` above completed with an **APPROVED** verdict, but on the **2026-05-30 commit, before** the two durable-resume advisory fixes (SQL-side narrowing + the cursor-ahead-of-head guard) and the rebase onto main. Those fixes landed after that review and are covered by their own regression tests (`tests/subscriptions/durable_event_log.test.ts`, 11 passing) and documented in [this comment](https://github.com/markmhendrickson/neotoma/pull/1482#issuecomment-4599411400).

**Status at merge:** `security_gates` ✅, full subscriptions suite 52/52 ✅, typecheck ✅. Merged to main.

PR #1486, opened 2026-06-01
fix(mcp): add SSE keepalive to /mcp transport so the stream survives proxy/client idle (#1483)

markmhendrickson:
## Summary

Fixes [#1483](https://github.com/markmhendrickson/neotoma/issues/1483): the MCP `/mcp` SSE stream silently drops under load behind reverse proxies and clients with idle timeouts, losing the tool registry until a manual reconnect. Adds an application-level keepalive (heartbeat frames + `X-Accel-Buffering: no` + TCP keepalive) to the long-lived GET SSE stream so it survives proxy/client idle.

The MCP `StreamableHTTPServerTransport` (SDK 1.29) holds the GET SSE stream open for server-initiated messages but emits no heartbeat. Node's `keepAliveTimeout` (set to 120s in v0.13.0 for #148) only governs socket reuse, not application-level SSE idle. A proxy (Cloudflare Tunnel, nginx, ngrok) or the client's idle timeout then closes the idle stream while `/health` stays green — the documented failure mode in the issue.

## Touchpoints

- [x] HTTP endpoint / request / response field 
- [x] MCP tool definition or behavior 
- [x] CLI command, flag, or runtime override 
- [x] Release supplement or release tooling 
- [x] Logging, metrics, events, traces 

## What changed

- **`src/actions.ts`** — new `attachMcpSseKeepalive(req, res)`, wired for `GET /mcp`:
  - Periodic bare SSE comment heartbeat frame `: hb\n\n` (default 25s, `NEOTOMA_MCP_SSE_KEEPALIVE_MS`, `0` disables). Comment frames are ignored by clients per the SSE spec and interleave safely with the SDK's events because Node serializes writes on the response stream.
  - `X-Accel-Buffering: no` injected via a `res.writeHead` patch. The `@hono/node-server` adapter rebuilds headers from the SDK's Web `Response`, so a header set via `res.setHeader()` before `handleRequest` would be discarded; patching `writeHead` is the only injection point that survives.
  - TCP keepalive on the socket; self-teardown + `writeHead` restore on stream close/finish.
- **`docs/developer/cli_reference.md`** — new "MCP / SSE transport tuning (server process)" table documenting `NEOTOMA_MCP_SSE_KEEPALIVE_MS` plus the pre-existing `NEOTOMA_KEEPALIVE_TIMEOUT_MS` / `NEOTOMA_HEADERS_TIMEOUT_MS` socket knobs.
- **`docs/releases/in_progress/v0.15.1/github_release_supplement.md`** — new supplement (post-release fix → next patch) with the mandatory "Breaking changes" section (`No breaking changes.`).
- **`tests/unit/mcp_sse_keepalive.test.ts`** — 8 new unit tests.
- **`docs/testing/automated_test_catalog.md`** — regenerated for the new test.

`retryInterval` (the issue's fix #3) was evaluated and **deliberately omitted**: in SDK 1.29 it only emits an SSE `retry:` field inside the POST→SSE priming event, which fires solely when an `eventStore` is configured (none here) and never on the GET standalone stream — so it would be an inert knob with no runtime effect. Rationale is documented in the code and the supplement.

### API & contract

- [x] No `openapi.yaml` changes needed — transport-level behavior only, no request/response/error fields added or changed.
- [x] No new `operationId`, MCP tool, or CLI command, so no `contract_mappings.ts` change.
- [x] New env var is `NEOTOMA_`-prefixed. It is a **server-process** knob read in `src/actions.ts` (not a CLI `preAction` override), documented in the new cli_reference "MCP / SSE transport tuning" table; it has no paired flag by design (consistent with the existing `NEOTOMA_KEEPALIVE_TIMEOUT_MS`).

### Security

- [x] `npm run security:classify-diff` → `sensitive=false` (keepalive framing on an already-authenticated stream; no auth / proxy-trust / local-dev / public-route / guest-access surface touched).
- [x] No new Express routes; `npm run security:manifest:check` in sync (108 routes).
- [x] No bare `req.socket.remoteAddress` / `X-Forwarded-For` / `Host` reads added. The helper reads `res.socket` / `req.socket` only to call `setKeepAlive` (not for auth/trust decisions).

### Release & process

- [x] Release-visible behavior documented in a supplement under `docs/releases/in_progress/v0.15.1/`; historical supplements untouched. Supplement contains the explicit `No breaking changes.` line.

## Test plan

- `npx vitest run tests/unit/mcp_sse_keepalive.test.ts` → 8 passing. Fake-timer tests assert **runtime** behavior (per `change_guardrails_rules` — HTTP runtime-config knobs must assert observable behavior, not source strings): heartbeat frames written only after the event-stream is established; `X-Accel-Buffering: no` injected on `text/event-stream` and never on JSON; no heartbeat written into a non-SSE (JSON-RPC POST) body; TCP keepalive enabled; heartbeat stops and `writeHead` restored on close; `keepaliveMs=0` disables the heartbeat while still enabling TCP keepalive.
- `npm run type-check` clean; `npm run lint` 0 errors; `npm run format:check` clean; `npm run lint:site-copy` OK.
- `npx vitest run tests/integration/mcp_invalid_bearer_auth.test.ts tests/integration/mcp_oauth_token_endpoint.test.ts tests/security/auth_topology_matrix.test.ts` → pass (confirms the `/mcp` auth/initialize path is unaffected).

## Notes for reviewer

- **Why heartbeat injection is safe against the SDK's stream:** the Node SDK transport wraps a Web-Standard transport via `@hono/node-server`, which pumps the SDK's `ReadableStream` into the Express `res`. My `res.write(": hb\n\n")` calls share that `res`; Node serializes writes, and each heartbeat is a complete SSE comment frame, so it slots safely between the SDK's events. The guard checks `res.writableEnded` / `res.destroyed` and the `text/event-stream` content-type before every write, so it never corrupts a non-SSE (JSON-RPC POST) body and stops itself once the stream ends.
- **CI note:** the pre-commit hook's full unit + integration suite was skipped for the commit (`NEOTOMA_SKIP_TESTS=1`) because this worktree has no `dist/` build and no `inspector/` assets, so a set of unrelated suites (`cli_smoke`, `cursor_hook_*`, `graph_neighborhood_pagination`, `package_scripts`, badge tests) fail as environment noise — confirmed identical on a stashed baseline. CI with a full build should run them green; please re-run the full suite in CI.

### DIFF
--- a/docs/developer/cli_reference.md
+++ b/docs/developer/cli_reference.md
@@ -271,6 +271,18 @@ These are read by the Neotoma HTTP server (not the CLI `preAction` hook) for out
 
 `GET /peers/{peer_id}` returns `remote_health` from probing `{peer_url}/health` and semver compat vs the local package version (same rules as `neotoma compat`). See `docs/subsystems/peer_sync.md`.
 
+### MCP / SSE transport tuning (server process)
+
+These are read by the Neotoma HTTP server when it serves the MCP StreamableHTTP transport at `/mcp`. They keep the long-lived SSE stream alive behind reverse proxies (Cloudflare Tunnel, nginx, ngrok) and clients with idle timeouts. See issue #1483 and `docs/security/threat_model.md`.
+
+| Environment variable | Default | Purpose |
+|----------------------|---------|---------|
+| `NEOTOMA_MCP_SSE_KEEPALIVE_MS` | `25000` | Interval (ms) between SSE comment heartbeat frames (`: hb\n\n`) on the MCP GET SSE stream. The frames keep proxies and clients from treating the stream as idle and silently closing it. Set to `0` to disable the heartbeat (restores prior behavior). |
+| `NEOTOMA_KEEPALIVE_TIMEOUT_MS` | `120000` | Node HTTP `server.keepAliveTimeout` (ms). Extended from Node's 5 s default so idle keep-alive sockets are not closed before typical proxy idle windows (60–300 s), which otherwise surfaces as a mid-stream RST / 502 reconnect cycle. |
+| `NEOTOMA_HEADERS_TIMEOUT_MS` | `125000` | Node HTTP `server.headersTimeout` (ms). Must exceed `NEOTOMA_KEEPALIVE_TIMEOUT_MS` to avoid a Node bug where the socket is closed before the headers timeout fires. |
+
+The MCP transport tuning vars are server-process knobs read in `src/actions.ts`, not CLI `preAction` overrides; they have no paired flag.
+
 ### MCP signed shim (`mcp.json` — not CLI `preAction`)
 
 Cursor reads these from the **`env`** block for **`run_neotoma_mcp_signed_stdio_dev_shim.sh`** (they are **not** parsed by `neotoma` CLI `preAction`; document them here for discoverability).
--- a/docs/developer/cli_reference.md
+++ b/docs/developer/cli_reference.md
@@ -273,11 +273,11 @@ These are read by the Neotoma HTTP server (not the CLI `preAction` hook) for out
 
 ### MCP / SSE transport tuning (server process)
 
-These are read by the Neotoma HTTP server when it serves the MCP StreamableHTTP transport at `/mcp`. They keep the long-lived SSE stream alive behind reverse proxies (Cloudflare Tunnel, nginx, ngrok) and clients with idle timeouts. See issue #1483 and `docs/security/threat_model.md`.
+These are read by the Neotoma HTTP server when it serves the MCP StreamableHTTP transport at `/mcp`. They keep the long-lived SSE stream alive behind reverse proxies (Cloudflare Tunnel, nginx, ngrok) and clients with idle timeouts. This is a reliability/transport concern; see issue #1483 and the [v0.15.1 release supplement](../releases/in_progress/v0.15.1/github_release_supplement.md).
 
 | Environment variable | Default | Purpose |
 |----------------------|---------|---------|
-| `NEOTOMA_MCP_SSE_KEEPALIVE_MS` | `25000` | Interval (ms) between SSE comment heartbeat frames (`: hb\n\n`) on the MCP GET SSE stream. The frames keep proxies and clients from treating the stream as idle and silently closing it. Set to `0` to disable the heartbeat (restores prior behavior). |
+| `NEOTOMA_MCP_SSE_KEEPALIVE_MS` | `25000` | Interval (ms) between SSE comment heartbeat frames (`: hb\n\n`) on the MCP GET SSE stream. The frames keep proxies and clients from treating the stream as idle and silently closing it. Set to `0` (or any value `<= 0`) to disable the heartbeat (restores prior behavior); an unset, empty, or non-numeric value uses the default. TCP keepalive on the socket stays enabled regardless. |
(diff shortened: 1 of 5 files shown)


markmhendrickson:
## Review findings addressed

Pushed `094e6c346` (rebased onto latest `main`; the branch now also folds in the existing v0.15.1 `entity_id`/#276 supplement so both fixes share one supplement). Each finding from the [Claude PR review](https://github.com/markmhendrickson/neotoma/actions/runs/26736774004):

### [BLOCKING] env parse coerces `NEOTOMA_MCP_SSE_KEEPALIVE_MS=0` to default — **fixed**

Confirmed bug: `parseInt(env || "", 10) || 25_000` turned `0` (and negatives) back into `25_000`, so the documented disable knob was dead while the unit test passed because it drove the internal `options.keepaliveMs` arg. Extracted a pure parser:

```ts
export function parseMcpSseKeepaliveMs(raw: string | undefined): number {
  if (raw === undefined) return MCP_SSE_KEEPALIVE_DEFAULT_MS;
  const trimmed = raw.trim();
  if (trimmed === "") return MCP_SSE_KEEPALIVE_DEFAULT_MS;
  const parsed = Number.parseInt(trimmed, 10);
  return Number.isFinite(parsed) ? parsed : MCP_SSE_KEEPALIVE_DEFAULT_MS;
}
```

Only `undefined` / empty / non-numeric falls back to the default; `0` and negatives flow through to the existing `keepaliveMs <= 0` disable guard.

### [ADVISORY] test coverage — **addressed**

Added a `parseMcpSseKeepaliveMs` matrix (`undefined`/`""`/`"   "`/`"abc"` → default, `"0"` → 0, `"-1"`/`"-1000"` → negative, `"5000"`/`" 5000 "` → parsed) plus an env-driven runtime test that sets the env to `"0"`, parses it the way the module does, and asserts no heartbeat is written. The suite is now 16 tests (was 8).

### [ADVISORY] `writeHead` patch could silently no-op on a future adapter — **addressed**

Added an opt-injectable `logger` to `attachMcpSseKeepalive` and a runtime guard: if the SSE stream ticks ≥ 5 times without `X-Accel-Buffering` showing up via `res.getHeader`, it `logger.warn`s once that the proxy-flush path may have regressed (e.g. a `@hono/node-server` upgrade that snapshots `res.writeHead`). Covered by warn / no-warn tests.

### [NIT] `setKeepAlive` used `keepaliveMs` (0 when disabled) as the probe delay — **fixed**

TCP keepalive probe delay is now clamped into `[15s, 60s]` independently of the app-level heartbeat, so the socket stays proxy-survivable even when the heartbeat is disabled. Tests assert `initialDelay === 15000` for `keepaliveMs=0` and a tiny positive value.

### [NIT] cli_reference section header pointed at `threat_model.md` — **fixed**

Replaced the threat-model pointer with issue #1483 / the v0.15.1 supplement (this is a reliability/transport concern), and clarified the disable wording (`<= 0` disables; unset/empty/non-numeric uses the default; TCP keepalive stays on).

### Validation

`npm run type-check` clean, lint 0 errors, `tests/unit/mcp_sse_keepalive.test.ts` 16/16 passing, test catalog regenerated + validated. Rebased clean onto `main`; PR is now mergeable.

> Note on the `retryInterval` line in the prior "Security hardening" supplement section — removed; the SSE `retry:` hint was deliberately omitted (inert without an `eventStore`), so the supplement no longer references it.

PR #946, opened 2026-05-26
feat(issues): add target_repo parameter to submit_issue

markmhendrickson:
## Summary

- Adds an optional `target_repo` (`owner/repo`) parameter to `submit_issue` (MCP tool, HTTP endpoint) so agents can file issues about repos other than the one Neotoma is globally configured for (e.g. `markmhendrickson/ateles`) without changing the Neotoma authoring home
- The GitHub mirror is redirected; the canonical Neotoma record (operator instance / `issues.target_url`) is unaffected
- When `target_repo` is absent, behavior is unchanged — mirror goes to the globally-configured `NEOTOMA_ISSUES_REPO`

Closes #942.

## Changes

- `openapi.yaml`: `target_repo` field on `/issues/submit` request body
- `src/shared/openapi_types.ts`: regenerated
- `src/shared/action_schemas.ts`: `IssuesSubmitRequestSchema` extended
- `src/services/issues/types.ts`: `IssueCreateParams.target_repo`
- `src/services/issues/issue_operations.ts`: passes `{ repo: target_repo }` opts to `github.createIssue`
- `src/actions.ts`: HTTP handler threads `target_repo` through to `submitIssue`
- `src/server.ts`: MCP handler Zod schema + pass-through
- `src/tool_definitions.ts`: `submit_issue` `inputSchema` and description updated
- `docs/developer/mcp/instructions.md`: multi-repo targeting rule added

## Test plan

- [ ] Existing `issue_operations.test.ts` — updated two `createIssue` call assertions to two-arg form; added `target_repo` override test asserting `createIssue` receives `{ repo: "markmhendrickson/ateles" }` as opts
- [ ] `entity_submission_github_handler.test.ts` — fixed pre-existing signature mismatch (now passes `undefined` as second arg)
- [ ] `npm test` passes locally (run manually; pre-commit hook uses `SKIP_TESTS=1` to avoid test latency on commit)
- [ ] Type check passes (`npm run type-check`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1494, opened 2026-06-01
fix(openapi): format generated types in openapi:generate + CI freshness guard

markmhendrickson:
## Problem

`npm run openapi:generate` emitted **raw** `openapi-typescript` output, which uses 4-space indentation. The committed `src/shared/openapi_types.ts` is Prettier-formatted (2-space, `printWidth: 100`). So any regenerate produced a spurious **~13k-line whole-file diff** even when `openapi.yaml` was unchanged.

Consequences:
- The generated file looked chronically out of sync with the spec.
- Trivial rebases that touched `openapi.yaml` turned into massive conflicts in the generated file (hit while rebasing #946).
- A contributor regenerating "to be safe" would balloon their diff and risk a `format:check` failure.

Root cause confirmed: generate + the repo's Prettier config produces a file **byte-identical** to what's committed (diff collapses from ~13k lines to 0). The committed file was always correct; only the script was missing the format pass.

## Fix

- **`package.json`** — `openapi:generate` now pipes through `prettier --write src/shared/openapi_types.ts`. Regenerating against an unchanged `openapi.yaml` is now a true no-op.
- **`.github/workflows/ci_test_lanes.yml`** — new "OpenAPI types are in sync with openapi.yaml" step regenerates and fails if `src/shared/openapi_types.ts` drifts, so it can't silently fall out of sync again.
- **`docs/architecture/openapi_contract_flow.md`** — documents that generate self-formats, CI enforces freshness, and the bare binary must not be run without the format pass.

## No content change

`openapi_types.ts` content is identical to `main` (the fix is the script, not the file). No type or behavior change.

## Test plan

- [x] `npm run openapi:generate` against unchanged `openapi.yaml` → zero diff (verified locally).
- [x] CI guard logic dry-run locally: regenerate then `git diff --quiet` → PASS.
- [ ] CI green on this PR (the new step exercises itself).

### DIFF
--- a/.github/workflows/ci_test_lanes.yml
+++ b/.github/workflows/ci_test_lanes.yml
@@ -54,12 +54,14 @@ jobs:
       - name: Format check
         run: npm run format:check
 
+      # Must run AFTER format:check so a stale committed file cannot be silently
+      # re-formatted by `openapi:generate` (which runs prettier --write) before
+      # the format gate sees it. Do not move this step above format:check.
       - name: OpenAPI types are in sync with openapi.yaml
         run: |
           npm run openapi:generate
           if ! git diff --quiet -- src/shared/openapi_types.ts; then
-            echo "::error::src/shared/openapi_types.ts is out of sync with openapi.yaml."
-            echo "Run 'npm run openapi:generate' and commit the result."
+            echo "::error::src/shared/openapi_types.ts is out of sync with openapi.yaml. If you edited openapi.yaml, run 'npm run openapi:generate' and commit the regenerated src/shared/openapi_types.ts."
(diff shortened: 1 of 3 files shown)


markmhendrickson:
Addressed all three nits in fdb88baa4 (all were optional, applied anyway since they harden the workflow):

- **CI error message** — expanded the `::error::` line to name `openapi.yaml` explicitly and state the regenerate-and-commit action.
- **Step ordering** — added an inline comment that the sync-check must stay after `format:check` so a stale committed file can't be silently re-formatted before the format gate.
- **Doc concision** — moved the bare-binary caveat into a `> Note:` callout so step 2 stays a short procedural instruction.

Thanks for the independent verification of the byte-identical-regen claim and the single-call-site check.

PR #1533, opened 2026-06-02
docs: align consistency + legal docs with shipped behavior

markmhendrickson:
Three doc-accuracy fixes surfaced while validating substrate guarantees.

## consistency.md
The doc described full-text search, vector embeddings, and cross-entity search as **bounded-eventual** (~5–10s indexing lag with an "Indexing…" UI banner). That was never the shipped behavior: search runs at request time against committed rows (no async indexer), and embeddings are written synchronously in the same store path as the snapshot. The summary table and the §2.2.5–§2.2.7 subsystem entries already say **Strong / 0s**; this PR finishes the job by removing the residual search-specific bounded-eventual examples in:
- §3.3 UI message templates
- §4.1 / §4.2 determinism examples and tests
- §6.1 / §6.2 test categories and the example suite

The bounded-eventual *framework* (definitions, UI handling pattern, test category) is kept as forward-looking guidance, explicitly marked as **not describing any current subsystem** so a future eventual subsystem still has a playbook.

## privacy_policy.md / compliance.md
- **"Encrypted at rest"** → qualified to the actual mechanism: AES-256-GCM **column** encryption for sensitive content (record fields, entity/relationship snapshots, raw fragments, event-log payloads) when encryption is enabled; the vector-embedding store relies on **volume** encryption because similarity search needs plaintext vectors.
- **"Row-Level Security (RLS)"** → corrected to **application-layer per-user isolation** (every query scoped to the authenticated user via `getAuthenticatedUserId` + `.eq("user_id", userId)`). RLS is noted as a possible future defense-in-depth layer, not the current control.

No code changes; docs only.

### DIFF
--- a/docs/architecture/consistency.md
+++ b/docs/architecture/consistency.md
@@ -51,7 +51,7 @@ After a write completes, subsequent reads MAY return stale data for an **unbound
 - Hardest to reason about
 - Complex conflict resolution if applicable
 **MVP Constraint:**
-Neotoma MVP uses **bounded eventual consistency** where needed, never unbounded eventual consistency.
+Neotoma's current subsystems are all **strongly consistent** (see §2.1); no subsystem is bounded-eventual today. The bounded-eventual model is defined here as forward-looking guidance: if a future subsystem ever needs it, it MUST be bounded, never unbounded eventual.
 ## 2. Consistency Tiers by Subsystem
 ### 2.1 Consistency Mapping Table
 | Subsystem                                                    | Consistency Model | Max Delay | UI Handling                |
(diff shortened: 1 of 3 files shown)


markmhendrickson:
Force-pushed a corrected rev (`d4bcf45a`) addressing both blocking classes from the review:

**1. consistency.md self-contradiction — fixed.** The §2/§3/§4/§6 examples were updated but the doc's own constraints still mandated the opposite. This rev rewrites:
- §1.2 MVP constraint (now: all current subsystems are strong; bounded-eventual is forward-looking)
- §3.2 / §3.3 UI templates (marked forward-looking; no indexing banner for search)
- §4.1 / §4.2 determinism examples (no search-as-eventual)
- §5.1 `search` row (Bounded Eventual → Strong) and §5.2 metadata envelope (reframed as reserved for a future eventual action; none today)
- §6.1 / §6.2 test categories (search test is now strong, no `sleep()`)
- §7 MUST / MUST NOT invariants (bounded-eventual items made conditional)
- **Agent Instructions → Constraints / Forbidden / Checklist** — the block that previously mandated "search and embeddings MUST be bounded eventual", "max delays search 5s / embeddings 10s", "UI MUST show indexing message" is rewritten to the strong/synchronous contract. An agent loading the doc now gets the shipped behavior.

**2. Prettier reflow that broke markdown structure — fully reverted.** The earlier rev ran `prettier --write`, which reflowed the whole file and merged the four GDPR ordered lists in compliance.md into one. These three docs are **not** prettier-conformant on `main`, so this rev does **not** run prettier at all — every change is a surgical string edit that preserves the existing structure. Diff vs main is now 97/116 lines (consistency) and 12/4 (legal), with no list renumbering and no bold-header indentation. Verified: `git diff main | grep '^+   \*\*'` is empty.

The summary table (rows already `Strong` on main) and the embedding footnote/read-after-write contract were already correct; this rev does not touch them.

PR #327, opened 2026-05-20
feat(inspector): consolidation, bundling, mode taxonomy, home page

castor-agent:
This branch is the **single merge candidate** for the inspector consolidation + bundling + mode-taxonomy work, rebased onto current `main`. It supersedes #323 and #324 (both still open, never merged individually). By patch-id it contains 100% of #323's content and all of #324's feature content, plus net-new inspector work.

## Problems
- The inspector lived in a separate submodule, complicating builds, versioning, and bundled-server delivery.
- The server had no boot-time mode taxonomy (`local` / `local_sandbox` / `hosted_sandbox` / `production` / `refuse`), so auth topology and the v0.11.1 advisory class were unguarded.
- HTML vs JSON content negotiation was inconsistent across the inspector SPA, `/docs`, and API routes.
- The inspector home was a thin shell; first-time visitors saw no product framing and a wall of zero-count stats.
- `/docs` hard-refresh served legacy server-rendered HTML to the SPA, which expected JSON — the "Could not load docs index — Unexpected token '<'" bug.
- `snapshot_filters` keys flowed unvalidated into a PostgREST column reference (filter-injection risk).

## Solutions
- Consolidated the inspector into `inspector/` via git subtree; dropped the submodule; chained `build:inspector` into the root build.
- Added `resolveSandboxMode()` with boot-time topology gating and a `sandbox_mode` field on `/me`.
- Unified content negotiation (`acceptPrefersHtml`) across the SPA fallback, `/docs`, and API routes.
- Rebuilt the home page as six marketing sections (hero, state-flow diagram, differentiators, outcomes, FAQ, quotes); the full stat grid now renders only for operators on their own install (`sandbox_mode === "local"`).
- Fixed `/docs` to honor `?format=json` / `Accept` and fall through to the SPA shell on browser requests; added `Vary: Accept`.
- Constrained `snapshot_filters` keys to snake_case identifiers via `SnapshotFieldNameSchema` with regression tests.
- Merge resolution: the branch's `snapshot_filters` and upstream's `exclude_bookkeeping` (#1459) both survive on the entities query.

## UX improvements
- Inspector home now communicates what Neotoma is, with deep links to docs/FAQ.
- Entities page: snapshot-field filtering, board/kanban view, bulk actions, inline edit.
- `/docs` renders client-side via the SPA on hard refresh instead of erroring.

## Documentation
- `docs/developer/getting_started.md` — added the Skills feature-guide row.
[...]

PR #1546, opened 2026-06-03
fix(launchd): build + watch Inspector SPA in watch-build agent

markmhendrickson:
## Problems
- The `com.neotoma.watch-build` LaunchAgent (`scripts/run_watch_build_launchd.sh`) ran `build:server` once at startup, then watched only server types via `dev:types`. It never built or watched the bundled Inspector SPA.
- Result: `dist/inspector/` goes stale. The prod server hot-reloads server code but keeps serving an old SPA bundle.
- Surfaced 2026-06-03: after #327 merged and the prod checkout moved to `main` (`0.15.0`), the server reported `0.15.0` but the Inspector home was still the pre-merge bundle (no marketing home), because nothing rebuilt `dist/inspector`.

## Solutions
- Build the Inspector once at startup via `build:inspector:prod-target`, alongside `build:server`, so `dist/inspector` is fresh at login / after reboot.
- Run `watch:inspector:prod` (vite `build --watch` → `dist/inspector`) as a background loop, mirroring how `dev:types` keeps server types current, so SPA source edits rebuild the served bundle automatically.
- Add an `EXIT`/`INT`/`TERM` trap to stop the inspector watcher when launchd terminates the agent (no orphaned vite watcher).

## UX improvements
No user-visible change to the app. Operational: future Inspector changes deploy automatically on the self-hosted prod instance instead of silently serving a stale bundle.

## Documentation
No functional API/MCP/CLI change; no user-facing docs required. The script's behavior change is documented inline in `run_watch_build_launchd.sh` comments, including the dated root-cause note.

## Test plan
- [x] `bash -n scripts/run_watch_build_launchd.sh` — syntax OK
- [x] Referenced npm scripts exist: `build:server`, `build:inspector:prod-target`, `watch:inspector:prod`, `dev:types`
- [x] Verified the one-time `build:inspector:prod-target` produces `dist/inspector/assets/home-*.js` containing the new marketing-home copy, and the running prod server serves it (HTTP 200)
- [ ] Manual: reload the LaunchAgent (`launchctl unload/load com.neotoma.watch-build`) and confirm both server and inspector rebuild at startup, and that editing an inspector source file triggers a `dist/inspector` rebuild

## Breaking changes
No breaking changes.

## Related
- Follow-up to #327 (inspector consolidation + bundling), which introduced the bundled `dist/inspector` the prod server serves.

### DIFF
--- a/scripts/run_watch_build_launchd.sh
+++ b/scripts/run_watch_build_launchd.sh
@@ -41,12 +41,43 @@ ensure_global_cli_linked
 
 # Full build once at startup (tsc + copy_pdf_worker_wrapper) so dist/ is complete before watch runs.
 if run_npm run build:server; then
-  echo "[$(timestamp)] Initial build completed."
+  echo "[$(timestamp)] Initial server build completed."
 else
-  echo "[$(timestamp)] Initial build failed; continuing with watch mode." >&2
+  echo "[$(timestamp)] Initial server build failed; continuing with watch mode." >&2
+fi
+
+# Build the bundled Inspector SPA into dist/inspector so the prod server serves
+# the latest UI. The watch loop below keeps it current on subsequent edits.
+# Without this the server hot-reloads but the SPA bundle goes stale (the cause
+# of the marketing-home-not-deploying issue on 2026-06-03).
+if run_npm run build:inspector:prod-target; then
+  echo "[$(timestamp)] Initial inspector build completed."
+else
+  echo "[$(timestamp)] Initial inspector build failed; continuing with watch mode." >&2
 fi
 
 RESTART_DELAY=5
+
+# Inspector watcher: vite build --watch into dist/inspector so SPA source edits
+# rebuild the served bundle, mirroring how dev:types keeps server types current.
+inspector_watch_loop() {
+  while true; do
+    run_npm run watch:inspector:prod || true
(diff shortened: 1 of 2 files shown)


markmhendrickson:
Added commit `5c51d28b4` — regenerated `src/shared/openapi_types.ts`.

The `baseline` CI failure here was **not caused by this shell-only change**. `main` shipped with `openapi_types.ts` out of sync with `openapi.yaml` (a `store_warnings.code` doc-comment that was edited in the spec but never regenerated), and the freshness guard from #1494 now fails on every PR that inherits that base. Verified the drift exists on `origin/main` itself by regenerating on a clean main checkout.

The extra commit is the verbatim output of `npm run openapi:generate` (doc-comment only, no shape change) and unblocks every other open PR gated on the same guard. Filed #1548 to track the root cause (main shouldn't ship out-of-sync generated types).

markmhendrickson:
Verified this PR needs no correction: its watch-build patch invokes `build:inspector:prod-target`, which on current `main` builds at `VITE_PUBLIC_BASE_PATH=/` (base `/`) — the correct base for serving at root. The earlier concern (that it baked `/inspector`) applied only to a stale prod checkout; main's build scripts already default to `/`.

Related: #1553 fixes the actual blank-page root cause (browser navigations to data routes were intercepted by JSON routes before the SPA fallback; now an early content-negotiation handler serves the SPA shell to browsers and `/inspector/*` 308-redirects to `/*`).

Issue #1548, opened 2026-06-03
main shipped with openapi_types.ts out of sync with openapi.yaml — CI freshness guard fails on all PRs


markmhendrickson:
## Summary
`origin/main` shipped with `src/shared/openapi_types.ts` out of sync with `openapi.yaml`. Running `npm run openapi:generate` on a clean checkout of main produces a 7-line diff (a doc-comment update to the `store_warnings.code` field description that was edited in the spec but never regenerated into the committed types).

The OpenAPI freshness guard added in #1494 ("OpenAPI types are in sync with openapi.yaml") now **fails the `baseline` CI job on every open PR**, regardless of whether the PR touches OpenAPI — because the drift is inherited from main's base.

## Impact
All open PRs are blocked on a red `baseline` check that none of them caused. First observed on PR #1546 (a shell-only LaunchAgent change that cannot affect OpenAPI).

## Root cause hypothesis
Either (a) the spec edit to `store_warnings.code` (appears related to the `MISSING_CONTENT_FIELD` content_field work, cf. PR #950) was merged without running `openapi:generate`, or (b) the freshness guard from #1494 was merged before the existing drift was reconciled. The pre-commit/pre-merge step that regenerates types did not run or was bypassed.

## Expected
A merge to main should never leave generated `openapi_types.ts` out of sync with `openapi.yaml`. The guard should be enforced at PR time against the PR's own changes, and main should always be green.

## Suggested fixes
1. Regenerate + commit `openapi_types.ts` on main (done on PR #1546 commit `5c51d28b4` as an unblock; ideally land a dedicated chore PR or fold into the next merge).
2. Ensure the pre-commit hook actually runs `openapi:generate` and stages the result when `openapi.yaml` changes — verify the guard in `.husky/pre-commit` (or wherever the OpenAPI sync check lives) writes, not just checks, OR fails loudly locally.
3. Consider a `main`-branch CI job that regenerates and fails if dirty, so drift is caught at merge rather than discovered by the next unrelated PR.

## Repro
1. `git checkout main && git pull`
2. `npm run openapi:generate`
3. `git status` → `src/shared/openapi_types.ts` is modified (should be clean).

## Environment
- `origin/main` at ~`cdeeeaa07` (2026-06-03), config `production`.

markmhendrickson:
Resolved on `main`. Verification:

- Latest `main` (b6b9744dc) + `npm run openapi:generate` → zero diff on `src/shared/openapi_types.ts`.
[...]

PR #1553, opened 2026-06-03
fix(inspector): serve SPA at / via early content-negotiation

markmhendrickson:
## Problems
- After #327, the inspector SPA shell was served at `/`, but browser navigations to data routes (`/sources`, `/entities/:id`, `/schemas`, `/relationships`, `/timeline`, …) hit the JSON data routes first — those routes are registered before the late SPA fallback and return JSON / 401 to browsers, so the SPA never rendered its own pages. Visiting `/` rendered only because of the late fallback; visiting a deep route showed JSON or an auth error.
- The SPA bakes its Router `basename` from `import.meta.env.BASE_URL`. When a stale `dist/inspector` was built with `VITE_PUBLIC_BASE_PATH=/inspector/` but served at `/`, the router refused to render → blank page (the symptom that prompted this).

## Solutions
- Register the SPA-shell content-negotiation handler **early**, before the auth gate and the data routes (`src/actions.ts`, just before the auth `app.use` at the public-key middleware). New `installInspectorSpaShellEarly` shares one handler with the late `installInspectorSpaFallback`. It serves the shell only when: `GET`/`HEAD` + Accept prefers HTML (q-aware `acceptPrefersHtml`) + not an API-only path + not an entity rendered-page surface. Browsers get the SPA; API / agent / curl clients (`application/json` / `*/*` / absent Accept) fall through to the real data route and get JSON.
- New `isEntityRenderedSurfacePath` predicate excludes `/entities/:id/html` and `/entities/:id/markdown` — server-rendered published pages (the html variant supports `?access_token=` guest reads, restored in `44a4a657`). Without it, the early handler would shadow a published page with the blank shell. **This is the highest-risk regression; it is guarded and tested.**
- Retire the `/inspector` prefix: swap `installInspectorMount` → `installInspectorLegacyRedirect` (308 `/inspector/* → /*`, query/hash preserved). The SPA already builds at base `/` (`VITE_PUBLIC_BASE_PATH=/`), so basename + assets resolve at root.
- Updated stale comments claiming serving at `/` requires migrating the API to `/api/*` — content negotiation makes that migration unnecessary.

## UX improvements
- Browser navigation to any SPA route (`/sources`, `/entities/x`, `/schemas`, …) renders the SPA. `/inspector/*` 308-redirects to `/*`. No user-visible change for API clients.

## Documentation
No functional API / MCP / CLI change and no OpenAPI shape change (no new routes, no field changes). Behavior is documented inline in `inspector_mount.ts` / `actions.ts` comments. No `docs/site/site_doc_manifest.yaml` entry required.

## Test plan
- [x] `npm run type-check`
- [x] `npm run lint` (0 errors)
- [x] `npm test -- tests/integration/inspector_content_negotiation.test.ts` (29/29)
- [x] `npm test -- tests/integration/inspector_bundled_mount.test.ts tests/integration/root_landing.test.ts` (98 total across the 3 files)
- [x] `npm test -- tests/contract/` (130/130)
- [x] `npm run build`
- [x] Manual against `node dist/actions.js`: browser `/sources` → `text/html` shell; API `/sources` → `application/json`; `/inspector` → 308 `/`; `/inspector/entities/abc?q=1` → 308 `/entities/abc?q=1`; `/` shell refs `/assets/*`; `/me` → JSON; `/entities/<id>/html` → handler JSON error (not shell).

## Breaking changes
No breaking changes. `/inspector/*` now 308-redirects to `/*` instead of serving the SPA directly; all paths and query/hash are preserved, so existing links keep working.

## Related
- Follows up #327 (inspector consolidation + root serving) and the unification plan `ent_1f176dbbe9a39e6bbad27f1f`.
- Fixes the blank-page issue tracked at `ent_f8acf218794512f12517dcf3`.
- Note: retiring the *term* "inspector" repo-wide is a separate follow-up, not in this PR.

### DIFF
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,8 +61,8 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **415**
-- Backend and repo Vitest files: **382**
+- Total automated test files: **416**
+- Backend and repo Vitest files: **383**
 - Frontend Vitest files: **9**
 - Playwright spec files: **24**
 
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -120,9 +120,10 @@ import {
 } from "./services/root_landing/index.js";
 import { mountDocsRoutes } from "./services/docs/index.js";
 import {
-  installInspectorMount,
+  installInspectorLegacyRedirect,
   installInspectorRootStaticAssets,
   installInspectorSpaFallback,
+  installInspectorSpaShellEarly,
   resolveBundledInspectorDir,
 } from "./services/inspector_mount.js";
 import { getSandboxTermsResponse } from "./services/sandbox/terms.js";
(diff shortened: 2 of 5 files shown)


markmhendrickson:
Addressed the review (commit `0c5a582ed`):

- **Advisory #1 (silent-behavior, the real bug)** — fixed. Added explicit `/openapi.yaml` + `/openapi_actions.yaml` to `API_ONLY_PREFIXES` (the bare `/openapi` entry didn't cover the `.yaml` suffixes because `isApiOnlyPath` matches at path-segment boundaries). Added unit + end-to-end regression tests confirming a browser GET to `/openapi.yaml` reaches the spec handler, not the SPA shell.
- **Advisory #4 (misleading comment)** — fixed. Corrected the "registered after /me" comment to clarify that later-registered routes (`/me`, `/openapi.yaml`) are protected by the deny-list, not by ordering.
- **baseline CI** — was failing on the OpenAPI freshness guard from inherited stale `openapi_types.ts` drift on main (tracked in #1548); regenerated here to go green.

Deferred as follow-ups (not blocking, no in-progress release tag yet):
- **Advisory #2** — `/inspector → /` retirement note + `Breaking changes:` line belong in the next release supplement when one is cut.
- **Advisory #3** — replacing `/inspector` URLs in `docs/developer/mcp/instructions.md` + `cli_agent_instructions.md` is a mirrored agent-instruction follow-up.

Security note acknowledged: the next release picking up this commit should run `security:lint` / `test:security:auth-matrix` and sign off `security_review.md` (no `isLocalRequest`/XFF/`Host`/`LOCAL_DEV_USER_ID` widening here — the handler serves static shell bytes).

Issue #1550, opened 2026-06-03
retrieve_entity_by_identifier returns empty for a valid ent_ id that retrieve_entity_snapshot resolves


markmhendrickson:
## Summary
`retrieve_entity_by_identifier` returns an empty result when passed a raw `ent_…` entity_id, even though `retrieve_entity_snapshot` resolves the same id without issue.

## Repro
1. Take a known-good entity_id (observed with a `priority_rubric` entity, id `ENTID0`).
2. Call `retrieve_entity_by_identifier` with that id → empty result.
3. Call `retrieve_entity_snapshot` with the same id → entity returned normally.

## Expected
Either:
- `retrieve_entity_by_identifier` accepts raw `ent_` ids as a valid identifier form and resolves them, **or**
- it returns an explicit error stating which identifier forms it accepts (type-name/slug vs raw id), rather than an empty/no-match result.

## Impact
Medium. The failure is silent and misleading — an empty result reads as "entity does not exist" when the entity is present and retrievable by snapshot. Callers waste a round-trip and may incorrectly branch on "not found."

## Suggested fix
Accept `ent_` ids in the identifier resolver, or have it raise a clear `unsupported_identifier_form` error enumerating accepted forms.

## Labels
bug

markmhendrickson:
Resolved by PR #1629 (merged): retrieve_entity_by_identifier now resolves literal `ent_` ids. APPROVED-WITH-NOTES; openapi/manifest in sync; security gates green.

Issue #1506, opened 2026-06-02
retrieve_entities: silent semantic→lexical search-mode degradation is invisible to callers (add search_mode signal)


markmhendrickson:
## Context

Raised while operating against a Neotoma instance with no embedding provider configured. A real-world multi-user adopter whose chat/retrieval depends on semantic recall would hit this in any environment missing `OPENAI_API_KEY`.

## Behavior

`retrieve_entities` with a `search` query resolves through three modes (`src/shared/action_handlers/entity_handlers.ts`):
1. **typed lexical** when type-filter tokens are present,
2. **semantic** (vector similarity) as the default search path,
3. **lexical fallback** when the semantic path returns zero entity IDs — e.g. `OPENAI_API_KEY` is unset, or the embedding query yields no vector hits.

The fallback chain is correct and works as designed (semantic-empty does not silently return nothing — it falls through to `queryEntitiesFromLexicalSearch`). **The gap is observability:** the response shape is `{ entities, total, excluded_merged }` and carries **no indication of which mode actually answered the query.**

## Why this is a trap

When no embedding provider is configured, *every* `search` call silently degrades from semantic ranking to substring lexical matching, and neither the caller nor a calling agent is told. The two indistinguishable outcomes are:
- "semantic search ran and genuinely found nothing", and
- "semantic search was unavailable; you got substring-on-`canonical_name` matching instead."

A consumer can ship believing semantic search is active, then quietly get keyword-only behavior in any deployment missing the key — with no signal to detect it. (This exact ambiguity led an operator to wrongly conclude semantic search was "unreliable" when the fallback was in fact working.)

## Proposed fix

Add an explicit mode signal to the response, e.g.:

```
search_mode: "semantic" | "lexical_fallback" | "lexical_typed"
```

and/or a one-time structured warning when the semantic path was requested but unavailable (no embedding provider). Document the field as a stable contract so consumers can switch on it (a token, not prose).

## Related

- #1496 (semantic search misses natural-language matches) is adjacent but distinct: that issue is about recall quality of the fallback; this issue is about the *caller not being told which mode ran*. Fixing this makes #1496-class misses diagnosable rather than silent.
- Same class as other "make the degraded state observable" hardening (e.g. the SSE gap-frame `reason` enum): degraded modes should be detectable by consumers, not inferred.

markmhendrickson:
A near-identical bug, independently found in another system, as evidence for why the silent part of this matters.

[...]

Issue #1505, opened 2026-06-02
refuse-mode boot banner suggests loopback remedy that fails under NEOTOMA_ENV=production


markmhendrickson:
## Summary
The `refuse`-mode boot banner in `src/actions.ts` (~line 9835) tells operators they can escape the refuse topology by binding to loopback:

> ... OR bind to loopback (NEOTOMA_HTTP_HOST=127.0.0.1), OR opt into ...

But `resolveSandboxMode()` in `src/services/sandbox_mode.ts` gates the `local_sandbox` verdict on `!productionEnv`. So when `NEOTOMA_ENV=production`, setting `NEOTOMA_HTTP_HOST=127.0.0.1` does NOT escape refuse mode. The banner's remediation is misleading for exactly the production-env case where operators are most likely to hit refuse.

## Impact
During a dev session, an operator followed the banner's loopback suggestion under `NEOTOMA_ENV=production` and it had no effect, costing debugging time. The banner should only list remediations that actually work for the active env.

## Expected
The banner remediation is accurate. Either (a) qualify the loopback suggestion as "only when NEOTOMA_ENV is not production", or (b) for production-env, point at the working escapes: `NEOTOMA_TRUST_PROD_LOOPBACK=1` (trusted-loopback dev), `NEOTOMA_REQUIRE_AUTH=1` + provisioned auth, or `NEOTOMA_SANDBOX_MODE=1`.

## Actual
Banner unconditionally suggests `NEOTOMA_HTTP_HOST=127.0.0.1`, which is a no-op under `NEOTOMA_ENV=production`.

## Suggested fix
Verify the exact set of working remediations against `resolveSandboxMode()` before rewording, then make the banner text conditional on `productionEnv`. Run the sandbox_mode resolver tests.

## Repro
1. Start the server with `NEOTOMA_ENV=production`, no auth configured, non-loopback bind → refuse mode.
2. Follow the banner: set `NEOTOMA_HTTP_HOST=127.0.0.1`, restart.
3. Observe the server still resolves to refuse (loopback path is gated on `!productionEnv`).

## Environment
- Confirmed on `main` at the inspector-consolidation merge.

Issue #1504, opened 2026-06-02
NEOTOMA_SQLITE_PATH advertised in CLI help but ignored by config.ts


markmhendrickson:
## Summary
`src/cli/index.ts` (~line 9708) advertises `NEOTOMA_SQLITE_PATH` as a supported override in CLI help text, but `src/config.ts` never reads it. `sqlitePath` is hard-derived:

```ts
// src/config.ts:155
sqlitePath: join(dataDir, env === "production" ? "neotoma.prod.db" : "neotoma.db"),
```

A user who sets `NEOTOMA_SQLITE_PATH` expecting to point the server at a specific DB file gets it silently ignored.

## Expected
Either `config.ts` honors `process.env.NEOTOMA_SQLITE_PATH` when set (using it verbatim, falling back to the env-derived default when unset), or the variable is removed from the CLI help line.

## Actual
The override is advertised but inert; behavior is unchanged regardless of the value.

## Suggested fix
Read `NEOTOMA_SQLITE_PATH` in `config.ts` and prefer it over the derived default. Add a unit test for both the override-set and override-unset paths. While there, audit the sibling overrides advertised in the same CLI help line (`NEOTOMA_DATA_DIR`, `NEOTOMA_RAW_STORAGE_DIR`, `NEOTOMA_EVENT_LOG_PATH`, `NEOTOMA_LOGS_DIR`, `NEOTOMA_PROJECT_ROOT`) and fix or de-advertise any others that are advertised-but-ignored.

## Repro
1. Set `NEOTOMA_SQLITE_PATH=/tmp/custom.db` in the environment.
2. Start the server.
3. Observe the server uses the default `neotoma.db` / `neotoma.prod.db` path, not `/tmp/custom.db`.

## Environment
- Confirmed on `main` at the inspector-consolidation merge (config `production`).

PR #1566, opened 2026-06-03
fix(mcp/cli): quick mechanical wins — #401 #1504 #1505 #1550 #1506 #367 #369 #370

markmhendrickson:
Ships the **Quick mechanical wins** cluster as one PR: eight low-complexity, mostly-independent fixes across CLI, MCP, and the HTTP contract surface. Each issue has a regression test.

### #401 — `mirror rebuild --profile` exited 0 writing 0 files when unauthenticated
- **Root cause:** `rebuildProfile()` in `src/services/canonical_mirror.ts` logged the API auth error to stderr and `return`ed silently, so the CLI exited 0 with an empty mirror.
- **Fix:** new `isAuthError()` helper detects `AUTH_REQUIRED` / HTTP 401; `rebuildProfile` throws on auth failure, and the CLI `mirror rebuild` catch block sets `process.exitCode = 1`. Non-auth API errors keep the prior log-and-continue behavior.

### #1504 — `NEOTOMA_SQLITE_PATH` advertised in CLI help but ignored
- **Root cause:** `src/config.ts` hard-derived `sqlitePath` from `dataDir`+env and never read the env var.
- **Fix:** `sqlitePath` now honors `process.env.NEOTOMA_SQLITE_PATH` when set, falling back to the env-derived default (`neotoma.db` / `neotoma.prod.db`).

### #1505 — refuse-mode boot banner promised a loopback escape under production
- **Root cause:** `resolveSandboxMode()` gates `local_sandbox` on `!productionEnv`, so binding `127.0.0.1` under `NEOTOMA_ENV=production` does NOT escape refuse mode, yet the banner advertised it.
- **Fix:** the banner now emits production-specific remediations (require-auth or sandbox-mode) and only offers the loopback escape outside production. Banner line construction extracted to `buildSandboxBootBannerLines()` in `sandbox_mode.ts` for unit testing.

### #1550 — `retrieve_entity_by_identifier` returned empty for a raw `ent_` id
- **Root cause:** the handler matched `canonical_name` / snapshot fields / semantic search but never the `entity_id` itself, so a raw id (which `retrieve_entity_snapshot` resolves) returned empty.
- **Fix:** a raw-id fast path resolves `ent_<24hex>` directly, scoped to the authenticated user. A well-formed but unknown id returns an explicit empty result rather than a degraded natural-language search.

### #1506 — `retrieve_entities` silently degraded semantic→lexical with no signal
- **Fix:** added a `search_mode` response field (declared in `openapi.yaml` first): `none` | `semantic` | `lexical_typed` | `lexical_fallback`. Threaded through both the HTTP `/entities/query` and MCP `retrieve_entities` responses so callers can detect when the embedding provider is unavailable.

### #367 — `/list_relationships` loaded the full set into memory then sliced
- **Root cause:** the handler ran an unranged query, materialized `data || []`, then `.slice(offset, offset+limit)` in process.
- **Fix:** pagination pushed to the DB via `.range(offset, offset+limit-1)` + `count: "exact"`; memory is now O(limit) instead of O(total).

### #369 — `/list_relationships` naming parity with `/retrieve_graph_neighborhood`
- The request fields and `total`/`limit`/`offset` were already declared. The remaining gap was naming: `/retrieve_graph_neighborhood` returns `total_count`.
- **Fix:** the response now returns `total_count` (canonical) alongside a **deprecated** `total` alias — additive and non-breaking. Both declared in `openapi.yaml`.

### #370 — `update_schema_incremental` fallback errors used a non-canonical envelope
- **Root cause:** the two fallback paths in `src/server.ts` returned `{ success: false, error_code, hint }`, which CLI/MCP error handlers (expecting `{ error: { … } }`) could not pattern-match.
- **Fix:** `ERR_NO_SCHEMA_FOR_ENTITY_TYPE` and `ERR_SCHEMA_MISSING_IDENTITY_CONFIG` now emit the canonical `{ error: { error_code, message, hint, details } }` envelope. `docs/reference/error_codes.md` updated to document the canonical shape (HTTP 400). The pre-existing cold-start test was updated to assert the canonical envelope.

## Contract surface

`#1506`, `#367`, and `#369` were coordinated in one `openapi.yaml` pass; types regenerated once. `npm run openapi:bc-diff` reports **no breaking changes** — both new fields (`search_mode`, `total_count`) are additive.

## Verification

- `npm run type-check` — pass
- `npm run lint` — pass (0 errors)
- `npm run format:check` — pass
- `npm run openapi:bc-diff` — no breaking changes (2 additive fields)
- `npm test -- tests/contract/` — 130 passed
- Targeted regression tests for all 8 issues — 31 passed

Pre-existing, unrelated test failures observed in this sandbox (confirmed failing on `origin/main` without these changes): `tests/cli/cli_smoke.test.ts` reset/data-dir protection tests and `tests/integration/graph_neighborhood_pagination.test.ts`. The full-suite pre-commit gate was bypassed via the hook's sanctioned `NEOTOMA_SKIP_TESTS=1` flag (type-check, lint, format, doc-deps still ran); `--no-verify` was not used.

### DIFF
--- a/docs/reference/error_codes.md
+++ b/docs/reference/error_codes.md
@@ -109,8 +109,8 @@ handlers can pattern-match the code uniformly.
 
 | Code                              | HTTP | Retry? | Description                                                                          |
 | --------------------------------- | ---- | ------ | ------------------------------------------------------------------------------------ |
-| `ERR_NO_SCHEMA_FOR_ENTITY_TYPE`   | 400  | No     | No registered or code-defined schema exists for the given `entity_type`              |
-| `ERR_SCHEMA_MISSING_IDENTITY_CONFIG` | 400 | No     | Existing schema lacks both `canonical_name_fields` and `identity_opt_out`            |
+| `ERR_NO_SCHEMA_FOR_ENTITY_TYPE`   | 200  | No     | No registered or code-defined schema exists for the given `entity_type`              |
+| `ERR_SCHEMA_MISSING_IDENTITY_CONFIG` | 200 | No     | Existing schema lacks both `canonical_name_fields` and `identity_opt_out`            |
 
 **`ERR_NO_SCHEMA_FOR_ENTITY_TYPE`** — raised by `update_schema_incremental` when
 the target `entity_type` has no schema registered in `schema_registry` and no
--- a/docs/reference/error_codes.md
+++ b/docs/reference/error_codes.md
@@ -103,26 +103,33 @@ interface ErrorEnvelope {
 Errors raised by `update_schema_incremental` and related schema-registry tools.
 These are structured (non-throwing) responses with an actionable `hint` field
 directing the caller to the right next tool, rather than opaque internal errors.
+They use the canonical standard error envelope (`{ error: { error_code,
+message, hint, details } }`, see `docs/subsystems/errors.md`) so CLI/MCP error
+handlers can pattern-match the code uniformly.
 
 | Code                              | HTTP | Retry? | Description                                                                          |
 | --------------------------------- | ---- | ------ | ------------------------------------------------------------------------------------ |
-| `ERR_NO_SCHEMA_FOR_ENTITY_TYPE`   | 200  | No     | No registered or code-defined schema exists for the given `entity_type`              |
-| `ERR_SCHEMA_MISSING_IDENTITY_CONFIG` | 200 | No     | Existing schema lacks both `canonical_name_fields` and `identity_opt_out`            |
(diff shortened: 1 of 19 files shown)


markmhendrickson:
@claude review

All pushed fix commits have settled and the baseline + security_gates lanes are green. The prior review run was cancelled by concurrency `cancel-in-progress` when a newer fix commit was pushed (or auto-skipped below the diff threshold). Requesting a fresh review against the current head so the CHANGES_REQUESTED verdict can be re-evaluated against the applied fixes.

PR #1564, opened 2026-06-03
fix(mcp/cli): list_relationships/delete discovery — #277

markmhendrickson:
## Cluster: list_relationships / delete discovery

Makes the relationship-type discovery path usable end to end so a caller can find the relationship type between two entities before calling `delete_relationship`.

### #277 — delete_relationship requires relationship_type with no discovery path

**Root cause:** `delete_relationship` requires the exact `relationship_type` between two entities, but a caller who does not know the type had no discovery path. Worse, the soft-delete silently "succeeded" by writing a deletion observation for an edge that never existed, so a wrong guess produced a misleading success with no pointer to discovery. The `/list_relationships` handler already supports `source_entity_id` + `target_entity_id` filtering and returns typed `RelationshipSnapshot` rows (each with its `relationship_type`), and `openapi.yaml` already declares those fields (landed by cluster A in #351 / commits `50bee813f`, `a40aa1c53`), but nothing surfaced that discovery path and deletion gave no hint.

**Fix (additive):**
- `delete_relationship` now verifies a live (non-deleted) relationship matches the supplied `(relationship_type, source_entity_id, target_entity_id)` triple via `relationshipsService.getRelationshipSnapshot`. When none matches it returns `404 RESOURCE_NOT_FOUND` with a structured `details.hint` pointing to `list_relationships` for type discovery, instead of a silent no-op deletion.
- `openapi.yaml`: declared the `404` response on `/delete_relationship`; documented the discovery flow on both `/delete_relationship` and `/list_relationships` descriptions. Regenerated `src/shared/openapi_types.ts`.
- MCP/CLI tool descriptions (`docs/developer/mcp/tool_descriptions.yaml` + in-code fallbacks in `src/tool_definitions.ts`) now document discovering the type via `source_entity_id` + `target_entity_id` before `delete_relationship`.

No new `ERR_*` code (reuses existing `RESOURCE_NOT_FOUND`). Tenant isolation is preserved: discovery and deletion both scope to the authenticated user via `getAuthenticatedUserId`.

### Cluster-A coordination note

#367 (pagination) and #369 (openapi declaration) for `/list_relationships` are owned by cluster A and already landed on main (`50bee813f`, `a40aa1c53`, PR #351). This change does **not** modify the `/list_relationships` handler query region — only an additive delete-side existence guard plus documentation. No overlap with cluster A's handler edits. The only `list_relationships`-adjacent edit here is its OpenAPI/tool description text (additive discovery guidance).

### Tests
[...]

Issue #368, opened 2026-05-21
bug: non-deterministic pagination order on /list_relationships and /retrieve_graph_neighborhood


castor-agent:
## Summary

Both endpoints have unstable sort orders that violate the determinism requirement in `docs/architecture/determinism.md`.

- `/list_relationships` orders by `last_observation_at DESC` only. Rows upserted within the same millisecond (common in batch operations) produce ties with unspecified resolution, causing items to move between pages across successive calls.
- `/retrieve_graph_neighborhood` has no `.order()` clause at all on its relationship queries before applying `.range()`. Postgres scan order is unspecified; the integration test in `tests/integration/graph_neighborhood_pagination.test.ts` only passes because of small fixtures and single-batch inserts.

## Remediation required

- `/list_relationships`: add `.order("relationship_key", { ascending: true })` (or `id`) as a stable secondary sort key
- `/retrieve_graph_neighborhood`: add an explicit `.order("last_observation_at", { ascending: false }).order("id", { ascending: true })` (or equivalent stable composite) before the `.range()` call on all relationship queries
- Update the pagination integration test to insert rows with identical `last_observation_at` values and assert stable page boundaries

markmhendrickson:
Resolved by PR #1491 (merged 2026-06-01). Shipping in the next release.

## What the issue asked for
Non-deterministic pagination order on `/list_relationships` and `/retrieve_graph_neighborhood` — results could come back in different orders across calls, breaking pagination stability. Determinism is a core invariant.

## How it was addressed
- Root cause: relationship pagination ordered only by `last_observation_at` (a mutable column that ties for same-millisecond upserts); `/retrieve_graph_neighborhood` had no `.order()` at all before `.range()`.
- Added a stable secondary sort key — `relationship_key ASC` (the `relationship_snapshots` PRIMARY KEY) — giving `last_observation_at DESC, relationship_key ASC` on all three paths (`/list_relationships`, `/retrieve_graph_neighborhood`, `GET /relationships`).
- Coexists with PR #1489's `entity_id` aliasing in the same handler (verified on rebase — both kept).

## Verification
- [x] New `tests/integration/relationship_pagination_determinism.test.ts` seeds 5 rows sharing an identical `last_observation_at` and asserts: identical order across 3 calls, order == `relationship_key ASC`, and gap-free disjoint pages across the tie. Confirmed failing-first.
[...]

Issue #367, opened 2026-05-21
bug: /list_relationships fetches unbounded result set into memory before paginating


castor-agent:
## Summary

`/list_relationships` (`src/actions.ts`) issues an unranged database query (no `.range()`, no DB-level `limit`), loads the entire result set into process memory, then slices it in application code:

```ts
const { data, error } = await query;
const all = data || [];
const paginated = all.slice(offset, offset + limit);
```

A `relationship_type`-only request with no entity-scoping fields fetches every relationship row matching that type for the tenant. Under high data volumes this causes unbounded memory growth and slow response times.

Additionally, the endpoint accepts `relationship_type` as the only filter with no required entity-scoping field (`source_entity_id`, `target_entity_id`), which should either be rejected or require at least one entity-scoping field.

## Remediation required

- Push pagination to the database: `.range(offset, offset + limit - 1)` on the Supabase query
- Issue a separate `count: "exact"` query for the `total` field instead of counting after a full fetch
- Require at least one entity-scoping field (`source_entity_id` or `target_entity_id`) when `relationship_type` alone is supplied, or reject unscoped requests explicitly

## Related

- #365 (per-user scoping on the same endpoint compounds this: currently fetches all tenants' data)

PR #1579, opened 2026-06-04
fix(perf): materialize relationship liveness for O(limit) list_relationships pagination (#1570, #1571)

### LINKED ISSUE #1570: fix(perf): list_relationships default path reverts #367's O(limit) pagination to unbounded O(n) load
## Summary

The default `list_relationships` path (`include_deleted=false`, the common case) loads the **entire** matching relationship set into memory before app-side soft-delete filtering and slicing, reverting the O(limit) DB-pushed pagination that #367 (PR #1566) shipped. The two changes are in unreconciled tension.

## Root cause

- #367/#1566 made `list_relationships` memory **O(limit)** via DB `.range()` + `count:"exact"`.
- #277/#1564 then changed the default to filter soft-deleted edges. `filterDeletedRelationships` resolves the highest-priority observation per key **app-side** — it cannot be a single DB predicate — so for an honest `total` it must load the full matching set, filter, then `slice(offset, offset+limit)`.
- Result: the default path has **no upper bound** on the loaded set (no `.range()`, no LIMIT). The O(limit) guarantee now holds only for the rare `include_deleted=true` audit path. Relationships are never hard-deleted, so per-entity edge counts grow monotonically.

## Impact

Hot path (default reads) loads every matching row into memory on every call. Latent until edge cardinality grows.

## Options

1. **Materialize liveness**: add an `is_live` / `deleted_at` column on `relationship_snapshots`, written transactionally when the deletion observation lands. Restores a single DB predicate (`.eq("is_live", true)`) → DB-side `.range()` + honest `count:"exact"` on the default path. Cost: a migration + write-path consistency with the highest-priority-observation rule (denormalization that must not drift from observation truth).
2. **Bound the pre-filter load**: cap the loaded set on the default path so memory is bounded even if `total` becomes approximate above the cap.

Decision depends on expected per-entity edge cardinality.

## Location

`src/actions.ts` `list_relationships` handler; `src/services/relationships.ts` `filterDeletedRelationships`.

## Provenance

Surfaced by a post-hoc architecture review of the merged MCP/CLI interface-bug PRs (#1563/#1564/#1565). Regression of #367 introduced by #1564.

### LINKED ISSUE #1571: bug: list_relationships HTTP vs MCP ordering divergence — unstable MCP sort causes page drift
## Summary

HTTP and MCP `list_relationships` paginate over **different orderings**, and the MCP ordering is **unstable** (no tiebreaker), so pagination over it can drop or duplicate rows across pages.

## Detail

- HTTP (`src/actions.ts`) orders by `relationship_key ASC` — stable.
- MCP (`src/server.ts`, ~line 2865) sorts by `last_observation_at DESC` with **no tiebreaker**. Edges with equal `last_observation_at` have undefined relative order, so `offset`/`limit` paging can skip or repeat rows.

Both a **parity** gap and a **determinism** violation (change_guardrails MUST #9 wants stable sort keys on read paths). #1564 reconciled the soft-delete *filter* across both paths but left the *ordering* divergent.

## Fix

Unify ordering across both surfaces; add a stable tiebreaker (e.g. `relationship_key`) to the MCP sort.

## Location

`src/server.ts` (~2865), `src/actions.ts` list_relationships handler.

## Provenance

Post-hoc arch review of merged PRs #1563/#1564/#1565.

markmhendrickson:
Closes #1570. Closes #1571.

## Problem

**#1570:** The default `list_relationships` path (`include_deleted=false`, the common case) loaded the **full matching edge set** into memory before app-side soft-delete filtering and slicing — reverting the DB-side `O(limit)` pagination shipped for #367 on the hot path. The MCP tool was worse: it ran **two unbounded `select('*')` queries** (outbound + inbound), then sorted/filtered/sliced in memory. Root cause: liveness was derived per-read from `relationship_observations` (highest-priority obs with `metadata._deleted`) and could not be expressed as a DB predicate, so an honest post-filter `total` required loading everything. The graph is monotonic (relationships are never hard-deleted), so this grows unbounded.

**#1571:** The MCP `list_relationships` sort used `last_observation_at DESC` with **no tiebreaker**, so rows sharing a `last_observation_at` (batch upserts) could move between pages — dropping or duplicating rows across `offset`/`limit` pages. The HTTP handler already had the stable `relationship_key ASC` tiebreaker (#368).

## Fix — materialize liveness (the durable option)

Add a `relationship_snapshots.is_live` column so the live-edge filter becomes a DB predicate:

- **Migration** (`sqlite_client.ts`): `is_live INTEGER NOT NULL DEFAULT 1` via the existing `addColumnIfMissing` soft-migration helper; `(source|target_entity_id, user_id, is_live)` composite indexes; a **one-time, idempotent backfill** that reconciles existing snapshots from the observation log (fast-exits when no deletion observations exist).
- **Write path** (`relationships.ts`, `deletion.ts`): `computeRelationshipSnapshot` stamps `is_live` from the same highest-priority-observation rule (new shared `isRelationshipLive` helper — single source of truth); `softDeleteRelationship` flips `is_live` to 0 after writing the deletion observation.
- **Read paths** (`actions.ts`, `server.ts`): both surfaces filter `is_live = 1` as a DB predicate on the default path and paginate at the DB (`.range()` + exact `count`), restoring `O(limit)` memory and eliminating the MCP double-unbounded-query. `include_deleted=true` keeps the unfiltered audit path.
- **#1571:** added `relationship_key ASC` tiebreaker to the MCP sort.

Observations remain the source of truth; `is_live` is a derived cache re-stamped on every recompute, so it self-heals and cannot drift.

## Key problems solved
- Unbounded O(n) memory on the default `list_relationships` read (regression of #367 introduced by #1564/#277).
- MCP page drift over rows with tied `last_observation_at`.

## Key solutions implemented
- Materialized `is_live` column + DB-predicate filtering + DB pagination on both surfaces.
- Stable MCP ordering via `relationship_key` tiebreaker.

## Why this is safe
- Additive column with `DEFAULT 1` (live) → pre-backfill reads behave exactly as today; no destructive change, no data loss, backward-compatible.
- **Internal only** — `is_live` is not exposed in any response; `openapi.yaml` unchanged (`openapi:generate` freshness clean, no bc-diff entry for this change).
- Determinism preserved; no PII.

## QA needs / Automated tests
- New `tests/integration/relationship_liveness_pagination.test.ts` (8 tests): default + paginated reads, `total_count` reflects live-only count post-DB-filter, deterministic paging over tied `last_observation_at` (#1571), `softDeleteRelationship` flips `is_live` and the default read drops the edge, `include_deleted=true` still returns it, and `isRelationshipLive` unit cases (empty/live/priority-wins-deletion/restoration).
- Full relationship suite (53 tests) + contract suite (132 tests) green. `type-check` / `lint` / `format` / `openapi:generate` freshness clean.

## Documentation update needs
- v0.15.1 supplement: added #1570 and #1571 entries under Fixes (noted internal-only, no API surface).


markmhendrickson:
@claude review

Re-review against new head `dbf703c25` after the test-catalog regeneration fix (CI baseline now green). The change since the last review is the regenerated automated_test_catalog.md only.

PR #1607, opened 2026-06-08
docs: R17 non-compete redline + schema-compatibility policy + exit-rebuild protocol

markmhendrickson:
Three adopter-safety improvements to the dependency-stability surface, surfaced while preparing an evaluator response. All docs-only.

## R17 — structural non-compete redline (governance fix)
`substrate_and_applications.md` cited **R13** as the source of Neotoma's commitment never to move up into the verticals built on it. But R13 is *"No category drift in positioning"* — a marketing/positioning redline, not a structural non-compete. The boundary doc was leaning on a redline that doesn't say what it was cited as saying.

This adds **R17 — No competition with the applications built on the substrate** as the actual structural commitment (no first-party app competing with an adopter; no absorbing an adopter's differentiating logic into the substrate and re-exposing it to their competitors), and corrects the citations in `substrate_and_applications.md` and `adopter_dependency_commitments.md` from R13 → R17. R13 keeps its narrower positioning scope; R17 cross-references it to disambiguate.

## Schema-compatibility policy (commitment #1: Partial → In place)
Adds `docs/architecture/schema_compatibility_policy.md` — the stated guarantee for registered-schema evolution: additive changes always safe, breaking changes require an explicit `schema_version` bump + migration path, a schema in use is never silently reshaped, and identity (`canonical_name`/`entity_id`) is stable across upgrades. The *mechanism* already existed (schema_version, additive expansion); this is the *policy* that turns it into an adopter guarantee. Flips scorecard row #1 to **In place** and clears its tracking issue.

## Exit-rebuild test protocol (commitment #6: stays Partial)
Adds `docs/developer/exit_rebuild_test.md` — the repeatable leave-and-rebuild procedure: export → `neotoma entities import` → teardown → rebuild → verify, timed against a *"weeks, not migration"* bar. **#6 deliberately stays Partial** — it flips only once the protocol is *executed against a realistic dataset and the run recorded*; the doc is the procedure, not the evidence. The protocol is explicit that the export→import-JSONL pairing is a gap the first run must confirm rather than assume — written honestly rather than overclaiming a clean round-trip.

## Scorecard movement
- #1 schema stability: Partial → **In place**
- #6 portability/exit: Partial (protocol written; recorded run pending)
- Net: In place rows 3 → 4 of 11.

No code changes; docs only. Doc-dependency validator passes; internal links verified.


markmhendrickson:
Superseded. Splitting this PR: the schema-compatibility policy (#1) and exit-rebuild protocol (#6) — which need no governance process — move to a clean PR. The redlines.md R17 addition + R13→R17 citation changes are held back to be handled as a proper amendment under the redlines.md amendment process (written rationale, reflection period, no single-party ratification per R12), rather than merged via a single-party PR — which would itself contradict R12.

PR #1617, opened 2026-06-09
feat(issues): M2 repo discovery resolver for .well-known/neotoma.json

markmhendrickson:
## Problems

With the manifest spec merged (#1612), nothing yet *reads* a repo's `.well-known/neotoma.json` to decide where issues about it should route. Layer 1 (`target_repo`) lets a reporter override the GitHub mirror, but they still have to know the destination out of band.

## Solutions

The M2 discovery resolver **core** — a self-contained, heavily-tested module that returns a routing decision. Wiring it into the live `submitIssue` path and the CLI base-URL resolver are deliberately **separate follow-ups**, so this trust-sensitive logic lands and gets reviewed in isolation before it touches a hot path.

- **`src/services/issues/repo_discovery_resolver.ts`** — `resolveRepoDiscovery(targetRepo, fetcher?)` reads the manifest (default fetcher hits the GitHub raw URL; injectable for tests), validates the v1 schema, and runs the two-layer trust check:
  1. **GitHub identity** — fetched only from the repo's raw URL; `manifest.repo` MUST equal the served `owner/repo` (spoof guard).
  2. **Key pinning** — `peer.public_key_thumbprint` is surfaced for the downstream guest-token exchange to pin; a peer URL alone is never sufficient.

  Returns a structured result: `{ route, effective_repo }` on success, or `{ route: null, reason }` with a typed reason (`no_manifest`, `fetch_error`, `invalid_json`, `unsupported_version`, `schema_invalid`, `repo_mismatch`, `peer_url_invalid`, `peer_url_private_host`) that the caller maps to the M4 rejection fallback or the operator default. Hosted-mode SSRF guard rejects private/loopback `peer.url` hosts.
- **`src/services/net/private_host_guard.ts`** — extracted `isHostedMode` + `isPrivateOrLoopbackHostname` from `sync_webhook_inbound.ts` into a shared module so the resolver and peer sync share one SSRF implementation. `sync_webhook_inbound` now imports it; **no behavior change**.

## Documentation

Implements the spec in `docs/subsystems/repo_discovery_manifest.md` (merged #1612). No new doc surface in this PR.

## Test plan

- [x] `repo_discovery_resolver.test.ts` — 19 tests: happy path, every no-route reason, schema-validation matrix, repo-mismatch spoof guard, case-insensitive repo match, SSRF reject under hosted mode + allow loopback off hosted mode.
- [x] `sync_webhook_inbound` suites still pass (12 tests) after the SSRF-guard extraction.
- [x] `npm run type-check` + `npm run lint` clean; test catalog regenerated + validated.
- [ ] CI green on this PR.

## Breaking changes

No breaking changes. New service module + a no-behavior-change refactor; no OpenAPI / schema / contract surface touched.

## Next

- PR 3: wire the resolver into `submitIssue` routing + the CLI base-URL resolver (endpoint discovery), and M3 guest-policy acceptance on the receiving peer.


markmhendrickson:
Addressed all 4 advisories + 3 nits in c92fef388 (while the code is fresh, ahead of PR 3 making the resolver hot — per the review's own recommendation):

**Advisories**
- **Fetch hardening:** default fetcher now uses `AbortSignal.timeout(10s)`, a 64 KiB response-size cap (content-length + body), and `redirect: "error"` so the manifest fetch can't be redirected to an attacker host (the `peer.url` SSRF guard doesn't cover a redirect of the fetch itself). Matches the peer-sync outbound convention.
- **`isOwnerRepo` tightened** to reject `.`/`..` segments and URL-reserved chars (`?#@:\`).
- **`version` reconciled** with the spec's minor-revision promise: documented it's the MAJOR version, the resolver enforces an exact major match, and minor revisions resolve because validation ignores unrecognized fields.
- **Reason → caller-behavior mapping documented** in `repo_discovery_manifest.md` (`no_manifest` falls through to the operator default; every other no-route reason is store-local-as-draft, never a silent downgrade), ahead of PR 3.

**Nits**
- `effective_repo` is now the manifest's canonical `repo` casing (not the caller's), for deterministic downstream identity.
- Added IPv6-loopback-SSRF, string-`version`, empty-body, and path-traversal targetRepo tests (29 total, up from 19).
- Added `pr_review_reading_list.md` rows for `repo_discovery*` and `net/private_host_guard*`.

type-check + lint clean; 29 tests pass; site-copy + doc-deps + catalog validated.

PR #1618, opened 2026-06-10
fix(entities): include status in lightweight list projection; add --status filter

### LINKED ISSUE #1586: bug(entities list): projection omits `status`/`attributes` + no `--status` selector → silent zero on status-filtered reads
## Summary

`neotoma entities list --type <T>` (and the underlying lightweight list path) returns a **reduced snapshot projection that omits `status` and `attributes`**, and there is **no `--status` selector** on the verb. Any consumer that filters the list output by `status` client-side therefore matches **zero rows silently** — even when the underlying entities have a valid `status`.

This silently breaks status-filtered reads built on `entities list`, with no error surfaced.

## Root cause (verified)

In `src/services/entity_queries.ts` (~line 450), the non-snapshot list path selects a fixed lightweight projection:

```
entity_id, schema_version, observation_count, last_observation_at, computed_at
```

(plus `canonical_name`, `entity_type`, `merged_*` from the entities table). `status` lives in the entity snapshot (`snapshot->>status`) and is **not** included in this projection, so it comes back absent. Reading `status` off any returned row yields `None`/undefined. There is also no `--status` filter option on the `entities list` CLI verb, so the filtering can't be pushed server-side either.

## Impact

Consumers that do `entities list --type contact` then filter `status in {...}` client-side get an empty set every time. In a real partner deployment this silently zeroed a load-bearing "active clients" context block — the block rendered "no active contacts" on every session despite valid data, leaving the agent blind to its live roster for an unknown period. The failure is **silent** (no error, no warning), which is the dangerous part.

## Reproduce

```
neotoma entities list --type contact --limit 1000
```
Every row returns only: `canonical_name, entity_id, entity_type, computed_at, last_observation_at, merged_at, merged_to_entity_id, observation_count`. No `status`, no `attributes`. Client-side `row.status` → `None`.

## Proposed fix (either or both)

1. **Include `status` (and consider `attributes`) in the lightweight list projection** so status-filtered client-side reads work. If full `attributes` is too heavy for the lightweight path, at minimum project `status` (it's a small, commonly-filtered field).
2. **Add a `--status` selector to `entities list`** (server-side filter on `snapshot->>status`) so callers can filter without pulling full snapshots — the more scalable path for large lists.
3. **Fail loud, not silent:** if a caller filters on a field the projection doesn't include, that's the deeper hazard — consider whether the lightweight projection should be documented/typed so consumers know `status` isn't present, rather than getting `undefined` silently.

## Acceptance criteria

- [ ] `entities list` output includes `status` for each row (or a documented opt-in), OR a `--status <value>` server-side filter exists.
- [ ] A status-filtered list read returns the matching rows instead of silently zero.
- [ ] Test covering: list entities with mixed statuses, filter/select by status, assert non-empty correct result.
- [ ] Docs: `cli_reference.md` entry for the list projection fields and/or the new `--status` flag.

## Context

Surfaced by an external partner deployment running Neotoma against its own client roster (reported via their feedback channel). The silent-zero behavior is exactly the class of failure that motivates loud projection/contract guarantees. PII/tenant specifics scrubbed.

markmhendrickson:
## Summary

Closes #1586

- **Root cause confirmed**: `entities list` (and any `queryEntities` caller with `includeSnapshots=false`) projected `entity_id, schema_version, observation_count, last_observation_at, computed_at` from `entity_snapshots` — no `snapshot` column at all, so `entity.snapshot.status` was always `undefined`. A partner's "active clients" context block silently rendered empty every session because the client-side `status === "active"` check always matched zero rows.
- **Fix 1** (`src/services/entity_queries.ts`): Add `snapshot->>status` to the lightweight column select (PostgREST text extraction). In the return map, when `includeSnapshots=false`, populate `snapshot: { status }` when the value is present; `null` → `snapshot: {}` (no crash, no silent break).
- **Fix 2** (`src/cli/index.ts` + `openapi.yaml`): Add `--status <value>` to `entities list`. Threads into `snapshot_filters: { status: { op: "eq", value } }` in the POST `/entities/query` body — reuses the existing server-side `snapshotFilters` mechanism entirely, no new query-engine code. Added `snapshot_filters` to `openapi.yaml` and regenerated `openapi_types.ts`.

**`attributes` not included**: `attributes` is a heavier variable-size nested object. `status` is the scalar that drives filtering. Callers needing `attributes` should use `include_snapshots: true`.

## Acceptance criteria

- [x] `entities list` output includes `status` per row (lightweight path)
- [x] `--status <value>` server-side filter exists and works
- [x] A status-filtered list read returns matching rows, not silent zero
- [x] Tests covering mixed-status list + status filter asserting non-empty correct result
- [x] `docs/developer/cli_reference.md` updated: projection now includes `status`, `--status` flag documented

## Test plan

- [x] 5 new unit tests in `tests/unit/entity_queries_status_projection.test.ts` — mocked DB, no infrastructure required:
  - lightweight path: entity with status → `snapshot.status` populated (core projection regression)
  - lightweight path: null status → `snapshot: {}` (not crash)
  - mixed-status list: each row carries its own status (not shared/empty)
  - `snapshotFilters` eq filter calls `snapshot->>status` column expression (query wiring)
  - status filter matching nothing → empty array, not error
- [x] 3 new CLI tests in `tests/cli/cli_entity_commands.test.ts` — fetch-mocked, no server:
  - `--status` sends `snapshot_filters` server-side; result is non-empty
  - returned entities include `snapshot.status`
  - omitting `--status` produces no `snapshot_filters` in request body

## Verification

```
npm run build:server        ✅ clean
npm run generate:test-catalog && npm run validate:test-catalog  ✅
npm run openapi:generate    ✅ snapshot_filters now in openapi_types.ts
npm run format:check        ✅
npm run lint                ✅ (0 errors, 329 pre-existing warnings)
npx vitest run tests/unit/entity_queries_status_projection.test.ts  ✅ 5/5
npx vitest run tests/cli/cli_entity_commands.test.ts               ✅ 8/8
pre-commit hook (full suite)  ✅ passed on commit
```


markmhendrickson:
## Review response — commit c457cda9e

Addresses all NEEDS-CHANGES items from the automated review:

**BLOCKING (MCP/CLI parity gap) — FIXED:**
- `src/tool_definitions.ts` (~line 218): Added `snapshot_filters` property to `retrieve_entities` inputSchema. Object with `additionalProperties` describing per-field `{ op, value }` filter shape (ops: eq/in/gt/lt/gte/lte/contains).
- `src/server.ts` (MCP `retrieveEntities` handler): Now passes `identityBasis`, `snapshotFilters`, and `excludeBookkeeping` through to `queryEntitiesWithCount`. These were already in the Zod schema parse path but not forwarded to the query function.
- `src/shared/action_schemas.ts`: Added `identity_basis` and `snapshot_filters` to `RetrieveEntitiesRequestBaseSchema` (the schema used by the MCP path) to match `EntitiesQueryRequestBaseSchema` (used by the CLI path).
- Handler wiring was NOT already present on the MCP path — the CLI path in `actions.ts` correctly passed `snapshotFilters` but `server.ts`'s `retrieveEntities` method omitted it entirely. Both the schema gap and the handler gap were fixed.
- `docs/developer/mcp/tool_descriptions.yaml`: Brief mention of `snapshot_filters` added to `retrieve_entities` description.

**baseline CI failure ("Validate automated test catalog") — FIXED:**
- Ran `npm run generate:test-catalog` to register new test file.
- `npm run validate:test-catalog` passes: "Automated test catalog is up to date."

**Grammar nit — FIXED:**
- `docs/developer/cli_reference.md`: "Use `--status ""` is not meaningful" → "Passing `--status ""` is not meaningful"

**Test added:**
- New `describe` block in `tests/unit/entity_queries_status_projection.test.ts`: schema-presence assertion that `retrieve_entities` inputSchema has a `snapshot_filters` property with the correct `op` enum (proving MCP tool-definition parity). All 6 tests pass.

**Follow-up tracked (not built):**
The advisory to move the hardcoded `snapshot->>status` projection to a registry/declaration-driven field list is a real improvement but out of scope for this fix. Tracked as follow-up in commit message.

**All checks run:**
```
npm run build:server                  ✅ clean
npm run validate:test-catalog         ✅ up to date
npm run format:check                  ✅
npm run lint                          ✅ 0 errors
npx vitest run tests/unit/            ✅ 1080 passed, 2 skipped
npx vitest run tests/contract/        ✅ 132 passed
```

PR #1628, opened 2026-06-10
fix(mcp): emit SSE retry: field on stream open (#1503)

### LINKED ISSUE #1503: Neotoma MCP SSE connection drops under load — requires manual reconnect in Claude.ai
## Summary

The Neotoma MCP server (`neotoma.markmhendrickson.com/mcp`) repeatedly drops its SSE connection to the Claude.ai MCP client during tool-heavy sessions. The server remains healthy (`/health` returns `ok: true, version: 0.14.0`) throughout — this is not a server crash. The connection loss is silent: Claude.ai loses the Neotoma tool registry (tools return "not registered") while other MCP connections (e.g. Gmail) stay up. Manual disconnect/reconnect in Claude.ai's settings restores the tools. This happened twice in a single session on 2026-05-29.

## Observed pattern

- Dropouts correlate with heavy tool calls: a large `correct()` write (Columba's full `prompt_markdown` rewrite, ~6KB payload) and large snapshot reads.
- After dropout, only non-Neotoma MCP tools remain. `tool_search` targeting Neotoma returns Gmail tools only.
- `/health` stays green throughout every dropout.
- Manual disconnect/reconnect in Claude.ai settings fully restores the registry.
- Did not occur during light tool calls (small reads, entity type counts, etc.).

## Likely root causes (in priority order)

**1. No SSE keepalive frames.**
Long-running SSE connections require periodic heartbeat frames (e.g. `: ping` comment lines every 15–30s) to survive proxies, load balancers, and client-side idle timeouts. Without them, an intermediate layer (Cloudflare, nginx, or Claude.ai itself) silently closes the stream. The server never detects the close; `/health` stays green.

**2. Synchronous tool call processing blocking the SSE stream.**
If tool calls are handled synchronously on the same connection that serves the SSE event stream, a slow or large write blocks the stream for the duration. Claude.ai's SSE client may time out waiting for the next event, drop the connection, and lose the tool registry — even though the write itself completed successfully (Columba corrections landed correctly despite the subsequent dropout).

**3. No `retry:` field on stream close.**
The SSE spec supports a `retry:` field instructing the client how long to wait before reconnecting. Without it, clients that see an unexpected stream close may not know to reconnect automatically.

**4. Large response payloads triggering connection reset.**
If the server closes/resets the connection after delivering a large response rather than keeping it alive for the next call, the client loses the stream after every heavy read.

## Evidence that this is server-side, not a pure Claude.ai client bug

- Gmail MCP connection (Google-hosted) stays up through every Neotoma dropout — same Claude.ai client, different MCP server behavior.
- Dropouts correlate with load (large payloads), not with idle time or session age alone.
- `/health` healthy throughout confirms the Neotoma process is alive; only the SSE stream is affected.

## Impact

- Requires manual intervention (disconnect/reconnect) to restore tool access mid-session.
- Silent failure mode: tools appear in the catalog but return "not registered."
- Blocks agentic sessions that depend on sustained Neotoma write access.
- Partially masks as a Claude.ai client bug; the real fix is server-side.

## Recommended fixes

1. **Add SSE keepalive frames** — emit `: ping\n\n` every 15–30s on the MCP SSE stream. Highest-leverage fix.
2. **Make tool call processing async** relative to the SSE stream — tool calls should not block the heartbeat/event loop.
3. **Add `retry:` on stream close** — tell clients to reconnect after N ms so dropouts are transparent.
4. **Test under load** — reproduce with a sequence of large `correct()` writes + snapshot reads and confirm the keepalive fix holds.

## Discovery context

Found during the Ateles agent definition review session (2026-05-29), after the Columba `prompt_markdown` correction (large payload) and during the subsequent retest sequence. Originally attributed to a Claude.ai client bug; reclassified as likely server-side after confirming `/health` stayed green throughout both dropout events. To be moved to the ateles repo.

markmhendrickson:
## Summary

The Neotoma MCP server's long-lived `GET /mcp` SSE stream drops to the Claude.ai client during tool-heavy sessions while `/health` stays green and other MCP servers (Gmail) stay up — evidence the fault is in Neotoma's SSE stream, not the client (#1503).

Issue #1483 already shipped the highest-leverage parts of the recommended fix on this same stream:
- **SSE keepalive frames** — a `: hb\n\n` comment ping every **25s** (`NEOTOMA_MCP_SSE_KEEPALIVE_MS`, default `25000`).
- **`X-Accel-Buffering: no`** so buffering proxies flush each frame, plus OS-level TCP keepalive.
- The heartbeat runs on an independent `setInterval` (issue #1503 fix #2), so **tool-call processing does not block the keepalive timer**.

This PR adds the remaining recommendation from #1503 (fix #3): the **SSE `retry:` field**.

- Writes `retry: <ms>\n\n` as the **first frame** on the GET SSE stream right after the SDK flushes headers, so the client records the reconnect delay on open and honors it on the next unexpected close.
- Independent of the heartbeat: emitted even when keepalive is disabled, since reconnect guidance is useful regardless of warm-keeping pings.
- Configurable via **`NEOTOMA_MCP_SSE_RETRY_MS`** (default **`3000`**; `0`/negative suppresses the frame, mirroring the keepalive disable sentinel).
- Frames carry **no PII** (only `retry:` / `: hb`).

Why we write `retry:` directly rather than the SDK's `retryInterval` option: in SDK 1.29 `retryInterval` only emits inside the POST→SSE *priming event* and only when an `eventStore` is configured; the standalone GET SSE stream that drops never emits a priming event, so the SDK knob is inert here.

### Keepalive interval
- Heartbeat ping: **25s** (`: hb\n\n`, pre-existing from #1483).
- Reconnect hint: **`retry: 3000`** (3s), new in this PR.

## Files changed
- `src/actions.ts` — `parseMcpSseRetryMs`, `MCP_SSE_RETRY_DEFAULT_MS`, `retryMs` option on `attachMcpSseKeepalive`, retry-frame write in the `writeHead` patch.
- `tests/unit/mcp_sse_keepalive.test.ts` — new `#1503` describe block + `parseMcpSseRetryMs` tests; existing heartbeat-only tests isolated with `retryMs: 0`.

## Test plan
- [x] `npx vitest run tests/unit/mcp_sse_keepalive.test.ts` → 24 passed (fake timers assert the retry frame is emitted once on stream open, ahead of heartbeats, and suppressed when `retryMs <= 0` or on non-SSE responses).
- [x] `npx tsc --noEmit` clean.

## Load verification is a follow-up
This cannot be fully verified without a load test against the deployed endpoint. The tests here assert that the `retry:` frame (and the pre-existing keepalive pings) are emitted on the stream at the expected point/interval using fake timers. **Full load-verification against the production `neotoma.markmhendrickson.com/mcp` endpoint — reproducing a sequence of large `correct()` writes + snapshot reads and confirming the stream survives — is a follow-up.**

Fixes #1503

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -852,7 +852,9 @@ const mcpServerInstances = new Map<string, NeotomaServer>();
 // solely when an `eventStore` is configured (resumability opt-in). We do not
 // configure an event store, and the stream that drops is the standalone GET
 // SSE stream, which never emits a priming event. Setting `retryInterval` would
-// therefore be an inert knob with no runtime effect, so it is omitted.
+// therefore be an inert knob with no runtime effect, so it is omitted. Instead
+// we write the SSE `retry:` field ourselves as the first frame on the GET SSE
+// stream (issue #1503); see MCP_SSE_RETRY_MS / parseMcpSseRetryMs below.
 
 /** Default MCP SSE heartbeat interval in ms when the env var is unset/invalid. */
 export const MCP_SSE_KEEPALIVE_DEFAULT_MS = 25_000;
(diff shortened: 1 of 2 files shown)


markmhendrickson:
@claude review

PR #1629, opened 2026-06-10
fix(mcp/cli): resolve literal ent_ ids in retrieve_entity_by_identifier

### LINKED ISSUE #1550: retrieve_entity_by_identifier returns empty for a valid ent_ id that retrieve_entity_snapshot resolves

### LINKED ISSUE #1561: retrieve_entity_by_identifier on an exact entity_id returns text-matching conversation_messages, not the entity itself
## Summary

Calling `retrieve_entity_by_identifier` with a concrete primary-key `entity_id` (e.g. `ENTID0`) does **not** return that entity. Instead it returns other entities (here, `conversation_message` rows) whose **text content mentions the ID string**, while the target entity is absent from the result set. A primary-key identifier lookup should resolve to the entity with that ID first (ideally exclusively), not run a full-text match on the ID.

## Reproduction

1. Pick any entity_id that is also referenced in the body/content of other entities — e.g. a plan whose ID appears in conversation_message snapshots.
2. Call `retrieve_entity_by_identifier` with `{ identifier: "ENTID1" }`.
3. Observe: results are 5 `conversation_message` entities that mention that ID in their `content`; the plan entity `ENTID2` itself is **not** in the list (`total: 5`, none matching the requested ID).
4. Contrast: `retrieve_entity_snapshot` with the same ID returns the correct plan entity. So the data is reachable — only the identifier-resolution path is wrong.

A second confirming case from the same session: `retrieve_entity_by_identifier` with `{ identifier: "ENTID3" }` (a freshly created issue) returned `total: 0` — an exact ID that exists was not resolved at all.

## Impact

Agents use `retrieve_entity_by_identifier` as the canonical "look up this specific entity" call (the MCP turn-lifecycle instructions explicitly direct: "use retrieve_entity_by_identifier for concrete identifiers"). When the identifier is an `ent_...` primary key, the current behavior either returns the wrong entities (text matches) or nothing, forcing a fallback to `retrieve_entity_snapshot`. This is surprising and breaks the documented contract for the most basic lookup.

## Expected behavior

When `identifier` matches the `entity_id` primary-key pattern (`ent_…`), resolve it as an exact ID lookup and return that entity. Reserve fuzzy/text/name matching for non-ID identifiers (names, emails, etc.). At minimum, an exact-ID match should rank first and always be included.

## Workaround

Use `retrieve_entity_snapshot` when you have the exact `entity_id`.

### LINKED ISSUE #1597: retrieve_entity_by_identifier with a literal entity_id returns tangential matches instead of hinting at retrieve_entity_snapshot
## Summary

Calling `retrieve_entity_by_identifier` with a literal `ent_…` entity_id does not resolve the entity. Without `entity_type` it returned tangential `conversation_message` entities that merely mention the id in their text; with `entity_type: "plan"` it returned `{entities: [], total: 0}`. The correct path to fetch by id is `retrieve_entity_snapshot`.

## Why this is a footgun

`retrieve_entity_by_identifier` is the natural-looking tool name when you hold an id, and its description ("name, email, etc.") doesn't make it obvious that an `ent_…` id is not a valid identifier here. The empty/tangential result gives no signal about the right tool, costing an extra round-trip.

## Suggested fix (low effort, DX only)

When the `identifier` matches the `ent_[0-9a-f]+` shape, either (a) short-circuit to a direct entity_id lookup, or (b) include a hint in the response, e.g. `hint: "identifier looks like an entity_id; use retrieve_entity_snapshot(entity_id=…) for direct fetch"`.

## Severity

Minor / DX. Likely working as designed (identifier ≠ entity_id), so a hint is probably sufficient rather than a behavior change.

Reporter: Claude Code web session, ateles repo.

markmhendrickson:
## Summary

`retrieve_entity_by_identifier` treated literal `ent_<hex>` entity ids as free-text identifiers, so an exact `entity_id` would either surface tangential `conversation_message` rows whose text merely mentions the id (#1561) or return empty (#1550) — even though `retrieve_entity_snapshot` resolves the same id fine. #1597 asked, at minimum, for a hint pointing callers at the right tool.

This PR makes the raw-id path robust and self-describing:

- **Direct primary-key lookup.** An identifier matching `^ent_[0-9a-f]+$` (previously a fixed 24-char width) short-circuits to a direct `entities.id` lookup scoped to the caller, returning the target entity **exclusively** as a `match_mode: "direct"` result. No name/text/semantic matching runs for id-shaped input, so tangential rows can never win.
- **Structured not-found hint.** When a well-formed `ent_` id resolves to nothing, the response is now `{ entities: [], total: 0, match_mode: "none", hint: "identifier looks like an entity_id; use retrieve_entity_snapshot(entity_id=…) for direct fetch" }` instead of a silent empty result. Callers can distinguish "no such id" from a name miss (#1597). Fuzzy/text/name matching remains reserved for non-id identifiers.

## Contract (OpenAPI-first)

- Declared the optional `hint` response field in `openapi.yaml` first, then regenerated `src/shared/openapi_types.ts` (`npm run openapi:generate`).
- `npm run openapi:bc-diff` classifies the new field as a **non-breaking addition** (`added-response-field POST /retrieve_entity_by_identifier [200].hint`).
- Mirrored the entity-id behavior note in both `docs/developer/mcp/instructions.md` and `docs/developer/cli_agent_instructions.md`, and extended the MCP tool description.

## Tests

`npx vitest run tests/integration/entity_identifier_handler.test.ts` — 12 passed. New/updated regression coverage:
- exact `ent_` id resolves to the target entity as the top/exclusive result, never the rows that mention the id in their text (#1561);
- exact `ent_` id resolves even when `entity_type` is supplied (#1597);
- unknown / cross-user `ent_` id returns `match_mode: "none"` + the structured `hint` (#1550, #1597).

`npx vitest run tests/contract/` — 132 passed. Type-check, lint, and format checks clean.

Note: the pre-commit full integration run was skipped via the hook's `SKIP_TESTS=1` flag (not `--no-verify`) because pre-existing, unrelated cursor-hook integration tests (`hook_failure_hint.test.ts`, `cursor_hook_stop_backfill.test.ts`) fail on this environment's baseline — confirmed failing on `origin/main` without these changes. All gates the hook still enforces (type-check, lint, format, doc-deps) passed.

Fixes #1561
Fixes #1550
Fixes #1597


markmhendrickson:
@claude review

PR #1632, opened 2026-06-10
fix(retrieve_entities): honor entity_types (plural) filter

### LINKED ISSUE #1562: retrieve_entities silently ignores entity_types (plural) param — runs unfiltered semantic search instead
## Summary

The MCP `retrieve_entities` tool advertises an `entity_types` (plural array) input parameter, but the handler never reads it. Callers passing `entity_types: ["plan"]` get a **pure semantic/text search across all entity types**, silently unfiltered — not a type-filtered list. This misleads agents into believing results are scoped to the requested type(s).

## Reproduction

1. Call `retrieve_entities` with `{ entity_types: ["plan"], query: "Neotoma plan", limit: 40 }`.
2. Observe results: only entities that rank for the query text are returned; `total` caps at the query's match count (e.g. 366), **not** the true plan count.
3. Call with `{ query: "a", limit: 5 }` (no type) → results include `agent_message`, `post`, `task`, etc. with `total` in the thousands — confirming the search ranges over all types regardless of any `entity_types` value.
4. Contrast: `{ entity_type: "plan", limit: 100 }` (singular, no search) → correct filtered list, `total` = true active-plan count.

## Root cause

- Tool schema (server, ~line 1296) declares `entity_types: z.array(z.string()).optional()`.
- Handler `retrieveEntities` (server, ~line 2917) passes only `entityType: parsed.entity_type` to `queryEntitiesWithCount`; it never references `parsed.entity_types`.
- The canonical request schema (`RetrieveEntitiesRequestBaseSchema`, shared/action_schemas.ts ~line 314) defines **only** `entity_type` (singular). The plural `entity_types` exists solely in the advertised tool schema and is dropped at parse/handler time.

## Impact

Agents relying on `entity_types` to enumerate or scope a type get incomplete, relevance-ranked results with no error or warning. This produces false "exhaustive scan" conclusions (e.g. an agent concludes a category has no matching entity when the matching entity simply didn't rank for the query text). Silent no-op on a documented parameter.

## Suggested fix

Either:
- (a) Honor `entity_types`: map it to an IN-filter (OR with singular `entity_type`) in `queryEntitiesWithCount`, and add it to the canonical request schema; or
- (b) Remove `entity_types` from the advertised MCP tool schema so callers can't pass a no-op param, and document that `entity_type` (singular) is the type filter.

Option (a) is preferable since multi-type filtering is genuinely useful. Either way, a non-empty `entity_types` that the server cannot honor should not be silently ignored.

## Workaround

Use `entity_type` (singular) with no `search` to get a deterministic, fully type-filtered, paginated list.

markmhendrickson:
## Summary

`retrieve_entities` (MCP) / `POST /entities/query` advertised an `entity_types` array input, but the handler never read it. Callers passing `entity_types: ["plan"]` silently received an **unfiltered** semantic/text search across all entity types — not a type-scoped list. This implements option (a) from the issue: **honor the parameter**.

- Added `entity_types` (string array) to the OpenAPI request schema (spec-first), regenerated `src/shared/openapi_types.ts`.
- Added `entity_types` to both canonical request schemas (`RetrieveEntitiesRequestBaseSchema` + `EntitiesQueryRequestBaseSchema`) and to the MCP `retrieve_entities` tool inputSchema, keeping tool/canonical/OpenAPI in sync.
- Applied an `IN`-filter over the **union** of singular `entity_type` + plural `entity_types` across every retrieval path: plain listing (`queryEntities`, `countVisibleEntities`), lexical search (`lexicalSearchEntityIds`), and semantic search (`semanticSearchEntities` → `searchLocalEntityEmbeddings` SQL). A non-empty `entity_types` is never silently ignored; an empty array is treated as no filter.
- Threaded the new param through both call sites (`src/server.ts` MCP handler and `src/actions.ts` HTTP route).

## Test plan

- [x] New integration tests in `tests/integration/mcp_entity_variations.test.ts`:
  - `entity_types: ["plan"]` → deterministic type-filtered list, `total` == true plan count.
  - multi-type `["plan", "note"]` → filters across both, excludes other types.
  - singular `entity_type` OR-combined with plural `entity_types`.
  - empty `entity_types: []` → no filter (all types returned).
- [x] `npx vitest run tests/contract/` → 132 passed (after `npm run build`).
- [x] Entity query suites pass: `entity_search_mode`, `lexical_search`, `mcp_entity_variations`, `entity_queries_status_projection`, MCP schema parity (`mcp_schema_variations`, `cli_to_mcp_schemas`).
- [x] `npx tsc --noEmit` clean; lint + prettier clean.

> Note: the full pre-commit suite has pre-existing failures in `cursor_hook_stop_backfill` / `hook_failure_hint` / `v0.2.0_ingestion` (require an unbuilt `@neotoma/client` workspace dist); these reproduce identically on `origin/main` and are unrelated to this change. Committed via the repo-sanctioned `SKIP_TESTS=1` (type-check/lint/format still enforced), not `--no-verify`.

Fixes #1562

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2553,6 +2553,18 @@ paths:
               properties:
                 entity_type:
                   type: string
+                  description: |
+                    Single entity-type filter. Combined as a union with
+                    `entity_types` when both are supplied.
+                entity_types:
+                  type: array
+                  items:
+                    type: string
+                  description: |
+                    Multi-type filter. When non-empty, results are restricted
+                    to entities whose `entity_type` is in this list (an IN
+                    filter), OR-combined with the singular `entity_type` when
+                    both are provided. An empty array is treated as no filter.
                 search:
                   type: string
                   description: Canonical free-text query parameter for retrieval relevance.
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -3714,6 +3714,7 @@ app.post("/entities/query", async (req, res) => {
 
     const {
       entity_type,
+      entity_types,
       search,
       limit,
       offset,
@@ -3734,6 +3735,7 @@ app.post("/entities/query", async (req, res) => {
       await queryEntitiesWithCount({
         userId,
         entityType: entity_type,
+        entityTypes: entity_types,
         includeMerged: include_merged,
         includeSnapshots: include_snapshots,
         sortBy: sort_by,
--- a/src/server.ts
+++ b/src/server.ts
@@ -2965,6 +2965,7 @@ export class NeotomaServer {
       await queryEntitiesWithCount({
         userId,
         entityType: parsed.entity_type,
+        entityTypes: parsed.entity_types,
         includeMerged: parsed.include_merged,
         includeSnapshots: parsed.include_snapshots,
         sortBy: parsed.sort_by,
--- a/src/services/entity_queries.ts
+++ b/src/services/entity_queries.ts
@@ -11,6 +11,13 @@ export interface SnapshotFilter {
 export interface EntityQueryOptions {
   userId?: string;
   entityType?: string;
+  /**
+   * Multi-type filter. When non-empty, restricts results to entities whose
+   * `entity_type` is in this list (an IN filter), OR-combined with the singular
+   * `entityType` when both are provided. An empty/omitted list applies no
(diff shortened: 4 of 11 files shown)


markmhendrickson:
@claude review

PR #1634, opened 2026-06-10
feat(security): override_validation service with per-field write policies (salvage #398)

markmhendrickson:
## Summary

Clean re-derivation of #398 (`feat/override-validation`) onto current `main`. The original branch was 10 commits and conflicted on security-advisory docs that **already landed on `main`**; this PR preserves only the genuinely net-new `override_validation` feature.

- **`src/services/override_validation.ts`** — `enforceOverridePolicy()` reads an `override_policy` JSON stored on `agent_definition` entities and enforces per-field write restrictions based on the caller's derived role (operator/service). Phase 1 scope: only `agent_definition` carries a policy; all other entity types pass through untouched (fail-open default).
- **Hooks** — wired into `createObservation` (`observation_storage.ts`) and `createCorrection` (`correction.ts`), after the existing attribution / protected-type checks.
- **Tests** — `override_validation.test.ts` (17) and `retrieve_graph_neighborhood_tenant_isolation.test.ts` (3); both net-new, both green.

## Intentionally excluded from #398

- `docs/security/advisories/*` and `.claude/skills/advisory/SKILL.md` — already on `main`.
- `docs/security/practices.md` edits — stale doc drift (`run_semgrep`->`lint` rename), no override-validation content.
- `tests/services/relationships_tenant_isolation.test.ts` — asserts a `user_id` tenant filter in `relationships.ts` that `main`'s **service layer does not implement** (advisory GHSA-wrr4-782v-jhwh documented on `main` but `getRelationshipsForEntity`/`getRelationshipsByType` take no `userId`). Filed separately as a security finding; not bundled here.

## Test plan

- [x] tsc --noEmit clean
- [x] override_validation.test.ts 17/17
- [x] retrieve_graph_neighborhood_tenant_isolation.test.ts 3/3
- [x] agent_capabilities_store.test.ts 4/4 (hook transparent for normal store)
- [ ] CI: full suite + security gates

### DIFF
--- a/docs/reference/error_codes.md
+++ b/docs/reference/error_codes.md
@@ -56,6 +56,7 @@ interface ErrorEnvelope {
 | `AUTH_EXPIRED`   | 401  | No     | Bearer token has expired                        |
 | `AUTH_MALFORMED` | 401  | No     | Bearer token format is invalid                  |
 | `FORBIDDEN`      | 403  | No     | Insufficient permissions (RLS policy violation) |
+| `OVERRIDE_POLICY_VIOLATION` | 403 | No | Write to a policy-protected field on an `agent_definition` entity denied for the calling agent role (see `src/services/override_validation.ts`; details carry `field_name`, `agent_role`, `entity_id`) |
 
 **Common Causes:**
 
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,8 +61,8 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **454**
-- Backend and repo Vitest files: **421**
+- Total automated test files: **455**
+- Backend and repo Vitest files: **422**
 - Frontend Vitest files: **9**
 - Playwright spec files: **24**
 
@@ -72,7 +72,7 @@ flowchart TD
 | Vitest unit tests | 116 |
 | Vitest service tests | 34 |
 | Source-adjacent tests | 52 |
-| Vitest integration tests | 130 |
+| Vitest integration tests | 131 |
 | Vitest CLI tests | 62 |
 | Vitest contract tests | 13 |
 | Vitest security tests | 3 |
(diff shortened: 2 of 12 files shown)


markmhendrickson:
All three BLOCKING findings + the tenant-isolation and contract advisories addressed in `f85708e34`:

1. **Snapshot no-op (correctness)** — appended `.maybeSingle()` and removed the `as unknown as Promise<…>` cast. The query now returns a single row object; enforcement actually fires. Also handles the snapshot column surfacing as either parsed object or raw JSON string.
2. **Error translation** — `OverridePolicyViolationError` now has an `instanceof` branch in `actions.ts` (403 + `buildErrorEnvelope` with the full envelope, adjacent to `AccessPolicyError`) and in `server.ts` (McpError `InvalidRequest` with envelope as MCP `data`).
3. **End-to-end test** — new `tests/integration/override_policy_enforcement.test.ts` runs real `createObservation`/`createCorrection` writes inside `runWithRequestContext`: service-identity deny (asserts class, code, statusCode 403, envelope fields), operator allow, unlisted-field allow, cross-tenant non-consultation, non-`agent_definition` bypass. The deny cases fail against the previous query shape, so the regression is locked.
4. **Tenant isolation (advisory)** — snapshot lookup now requires `.eq("user_id", userId)`; `userId` threaded from both callers; covered by the cross-tenant test case.
5. **Contract (advisory)** — `OVERRIDE_POLICY_VIOLATION` registered in `docs/reference/error_codes.md`; 403 response declared on `/store` and `/correct` in `openapi.yaml`; `openapi_types` + test catalog regenerated. `security:manifest:check` clean.
6. **Dead-code (advisory)** — kept phase-1 gating with an explicit comment that the guard is inert until an `agent_definition` schema is seeded; shipping the write-path hook first is intentional.

Gates: tsc clean; 6/6 new integration + 17/17 unit + 3/3 tenant-isolation tests green.

markmhendrickson:
@claude review

Re-requesting: this branch was rebased onto current main and the prior review's findings were addressed in new commits (the last review predates the current head). Please re-review the updated diff.

markmhendrickson:
Rebased onto latest `main` (2026-06-12). Targeted tests green (override_validation 17/17, graph-neighborhood tenant 3/3). Please re-review.

PR #1644, opened 2026-06-12
feat(cli): auto-mirror skills to all harnesses with correct Cursor dir

markmhendrickson:
## Problem

Skills authored in `neotoma/skills/` only reached a harness at one-shot `neotoma setup` time. If a harness was never set up — or installed later — its skills directory was never created or populated. Concretely, `/end`, `/status`, and every other Neotoma skill were invisible in Cursor because `~/.cursor/skills` was never created (only Claude Code had been set up). New skills added to the source also did not propagate to already-configured harnesses.

A secondary bug: `~/.cursor/skills-cursor/` (Cursor's own managed built-in skill root) was mistaken for the user-skills dir. Cursor actually scans **`~/.cursor/skills`** for user skills — verified against the Cursor app bundle, which lists `.cursor/skills`, `.cursor/skills-cursor`, `.cursor/cloud-skills`, and `.cursor/plugins` as distinct roots.

## Change

- **`src/cli/skills_mirror.ts`** — shared reconciler mirroring the canonical `skills/` source into each harness skills directory:
  - **Whole-dir symlink** (`<harness>/skills` → `neotoma/skills`) when the target is absent or already our symlink — new/removed/renamed skills propagate instantly with zero per-skill drift.
  - **Per-skill symlink fallback** when the target holds foreign content (e.g. a harness wrote its own skills there) — foreign entries are preserved, never clobbered. Our stale links are pruned when a source skill is removed.
  - Keyed on the harness **base** directory (`~/.cursor`, `~/.codex`, …), so a harness installed *after* setup is auto-populated on the next sync.
- **Correct Cursor target**: `.cursor/skills`, not `.cursor/skills-cursor`.
- **`setup.ts`** refactor: `installSkills()` now delegates to the shared reconciler, so `neotoma setup` and continuous sync share one code path.
- **`neotoma skills sync`** CLI command — mirrors to every detected harness; `--scope user|project`, `--json`.
- **`com.neotoma.skills-sync` LaunchAgent** — watches `skills/` and re-runs `neotoma skills sync` on change (plus `RunAtLoad`). Uses `fswatch` when present, with a portable 30s polling fallback. Installer + `npm run setup:launchd-skills-sync`.

## Tests

`tests/cli/skills_mirror.test.ts` (10 cases): whole-dir vs per-skill selection, foreign-content preservation, absent-harness skip, idempotency, auto-population of a late-installed harness, and propagation + pruning. Existing `tests/cli/cli_doctor_setup.test.ts` (46 cases) still passes against the refactored `installSkills`.

## Activation note

The `com.neotoma.skills-sync` LaunchAgent runs `neotoma skills sync`, which requires a built CLI containing this command. After this merges and the `neotoma` CLI rebuilds/relinks, run `npm run setup:launchd-skills-sync` to start continuous mirroring. The one-shot mirror already works via `neotoma setup` and `neotoma skills sync`.


markmhendrickson:
Addressed the three risk areas the automated scoping pass surfaced:

1. **`cli_command_coverage_guard` registration** — registered `skills` under `coveredBehavioral` (its `skills sync` reconciler is behaviorally covered by `tests/cli/skills_mirror.test.ts`). Guard test now passes.
2. **`rmSync recursive force` on harness skills dirs** — removed. `mirrorWholeDir` now unlinks a wrong-target symlink, or removes our individual skill symlinks and `rmdir`s the empty directory, and **bails to per-skill mode** if it ever encounters a real (non-symlink) entry. A destructive `rm -rf` of real harness content is now impossible even under a logic error. Added a regression test (`stray real entry forces per-skill mode`).
3. **Shell script portability** — the LaunchAgent is macOS-only by design (launchd); the installer exits with a clear message on non-Darwin pointing to systemd/cron, and the polling fallback's BSD `stat` is therefore in-scope. The watch script prefers `fswatch` and degrades to a 30s poll.

Also fixed a latent misclassification found while hardening: re-running over an existing whole-dir symlink read the linked source skills as "foreign" and flipped to per-skill mode — now short-circuited.

58 tests pass (12 mirror + 46 setup + coverage guard). The fix has been applied to a real machine: `~/.cursor/skills` → `neotoma/skills`, `/end` and `/status` resolve under Cursor, and `~/.cursor/skills-cursor` (Cursor's built-ins) is untouched.

markmhendrickson:
@claude review

Re-requesting review against commit e947d3f3e, which addresses all actionable findings from the prior APPROVED-WITH-NOTES review:

- **Fixed the `--json` bug** (declared the flag on the `skills sync` subcommand; it was previously unregistered and would be rejected by Commander).
- **Surfaced per-skill failures**: `HarnessMirrorResult.errors[]` + report `has_errors`; CLI prints warnings and exits 1 on partial failure.
- **Fixed an EISDIR crash** found by a new conversion test: the per-skill→whole-dir path used `rmSync(recursive:false)` on a directory; now `rmdirSync` on the already-emptied dir (still never a recursive delete).
- **Made the per-skill→whole-dir conversion observable** via an `onLog` callback (CLI prints; LaunchAgent logs).
- Deterministic skill/harness ordering; removed the dead `skillNames` param.
- Marked `run_skills_sync_launchd.sh` macOS-only; extended reload/shutdown launchagent scripts to include `com.neotoma.skills-sync` so their docs are accurate.
- Docs: added `Skills` section to `cli_reference.md`; added `setup:launchd-skills-sync` row to `npm_scripts.md`.

Deferred with rationale: `contract_mappings.ts` registration and the release supplement — consistent with the established infra-only carve-out for `setup`/`hooks`/`mirror`/`agents`, and no in-progress release tag targets this yet.

PR #1666, opened 2026-06-16
feat(cli): entities export — the import inverse for the exit-rebuild round-trip

markmhendrickson:
Adds **`neotoma entities export`**, the missing inverse of `entities import`, and records a measured Neotoma-side dry run of the exit-rebuild protocol (adopter-dependency commitment #6).

## Why
Running the exit-rebuild protocol (`docs/developer/exit_rebuild_test.md`) surfaced that **no paired export command existed**: `snapshots export` emits a nested shape (fields under `snapshot`, plus `entity_id`/provenance), and `entities list` isn't import-shaped — so there was no one-command leave-and-rebuild round-trip. This closes that gap.

## What
- **`entities export [--type] [--out] [--with-relationships] [--page-size] [--include-merged]`** — pages `/entities/query` and emits import-compatible JSONL (one object per line: `entity_type` + the snapshot's fields flattened to the top level, exactly what `entities import` consumes). `--with-relationships` also writes a companion `<out>.relationships.json` that `relationships create --file` re-imports, for a full entities-plus-edges round-trip. Reducer metadata (`entity_id`, provenance) is intentionally dropped — identity re-derives deterministically on import.
- Test: `tests/cli/cli_entities_export.test.ts` (import-shaped output; paging). Test catalog regenerated. CLI coverage guard passes.
- Docs: registered in `cli_reference.md`; `exit_rebuild_test.md` updated to make `entities export` the canonical export side and to record the reference run + findings.

## Reference dry run (recorded in exit_rebuild_test.md)
10k synthetic `contact` entities, isolated dev instances, offline transport:
- import ~9.9s, export ~2.5s, rebuild-import ~8.1s, idempotent re-run ~0.5s (no duplicates)
- exporting the **rebuilt** instance produced **byte-identical** content to the original export — a faithful round-trip at the declared-field level
- **new finding (documented):** export fidelity is bounded by the registered schema — fields a schema doesn't declare aren't in the snapshot and don't survive this path. A real migration must declare every field that matters. (Relationship round-trip is implemented but not yet exercised — the reference dataset had no edges; the joint adopter run covers it.)

This is the substrate side of #6 de-risked. #6 still flips to In place only on a recorded run against an adopter's real (Postgres-shaped) dataset, run jointly — unchanged.

### DIFF
--- a/docs/developer/cli_reference.md
+++ b/docs/developer/cli_reference.md
@@ -594,6 +594,14 @@ See `docs/developer/agent_cli_configuration.md` for the rule text and strategy.
   - `--user-id <userId>`: store under a specific user.
   - `--commit`: actually write. Without it the command runs in dry-run/plan mode.
   - The `/store` endpoint already writes each batch transactionally; this command owns file reading, chunking, and per-chunk idempotency. For very large backfills, this is the supported "replay from an external source" path.
+- `neotoma entities export`: the inverse of `entities import` — pages through all entities and emits import-compatible JSONL (one object per line, `entity_type` plus the snapshot fields at the top level), so an instance can be exported and rebuilt with a matched command pair. See [`exit_rebuild_test.md`](exit_rebuild_test.md) for the leave-and-rebuild protocol this supports.
+  - `--type <entityType>`: only export this entity type.
+  - `--out <path>`: write JSONL here (default: stdout).
+  - `--with-relationships`: also write a companion `<out>.relationships.json` of typed edges that `relationships create --file` re-imports, for a full entities-plus-edges round-trip.
+  - `--page-size <n>`: entities fetched per page (default `500`).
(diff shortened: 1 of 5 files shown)


markmhendrickson:
Addressed two of the review advisories in `0e6425d26`:
- **determinism**: export paging now pins `sort_by=entity_id` / `sort_order=asc` so order is owned by the command, not the endpoint's default — this is what makes the byte-identical round-trip reproducible rather than incidentally-true. Added a test asserting every page request carries the pinned sort.
- **silent-behavior**: `--with-relationships` without `--out` is now rejected with a structured hint instead of paging edges and discarding them. Added a test for the rejection.

The remaining advisories (contract field-declarations on GET /relationships, MCP↔CLI parity for an export tool, release supplement entry) are lighter follow-ups and don't block this read-only CLI helper; I'll track them separately.

Issue #1703, opened 2026-06-22
Build eval-harness assertion primitives: mcp_tool.invocations + retrieval/result-shape + schema-version + snapshot-field-presence


markmhendrickson:
Surface: the eval-harness assertion vocabulary (assertions.ts) currently supports only store_structured.calls, entity.exists, entity.count, observation.with_field, relationship.exists, relationship.count, reply_text.contains, host_tool.invocations, turn_compliance.backfilled, instruction_profile.served. It can observe store-side entity/relationship/observation state and host-tool calls, but it CANNOT (a) assert that an arbitrary neotoma MCP tool was invoked or inspect that tool's return shape, (b) assert a schema_version value or that a field is present/absent on a returned snapshot. This is the blocking dependency for nearly every gap eval below.

Risk: without these primitives, schema, retrieval, subscription, peer-sync, and turn-summary tools are un-evaluable at the agent layer — an agent that lucked into the right entity state passes, while one that never called the intended tool also passes. Wrong-tool / silent-no-op behaviors are invisible.

Eval to author (infrastructure, not a scenario): add assertion types — (1) mcp_tool.invocations (count + optional arg-shape match for any neotoma tool, mirroring host_tool.invocations); (2) tool_result.matches (inspect the JSON the agent received from a named tool: id-set, count, presence of a key e.g. webhook_secret, error-envelope code); (3) snapshot.field_present / snapshot.field_absent (retrieve_entity_snapshot field projection); (4) schema.version_bump (minor vs major after a schema mutation). Wire each into runner result capture.

Key edge cases: tool-result capture must record both success payloads and {error:{code,message}} envelopes so error-surface assertions work; arg-shape matching must be structural (subset) not exact.

Deterministic agent-observable assertion: a self-test scenario where the agent calls a known tool with known args yields mcp_tool.invocations==1 with matching arg subset, tool_result.matches the captured payload, and field_present/absent reflect the seeded snapshot. PII-free.

---
*Part of the eval-coverage backfill (CA audit 2026-06-22). Umbrella: #230. **This is the foundational blocker** — most other `eval-coverage` issues depend on these assertion primitives. Coverage map: Neotoma `coverage_map` ent_3d1d72dd3c5a0d9086d45317.*

Issue #230, opened 2026-05-17
Audit all existing agentic behavior in agent instructions and create evals


castor-agent:
## Summary

Conduct a systematic audit of all agentic behavior currently defined or implied by agent instructions across the codebase, and create evals covering each identified behavior.

## Motivation

As the eval requirement for agentic PRs is formalized (see #229), there is an existing gap: agentic behaviors that were implemented before that policy exist without eval coverage. This issue tracks the work to close that gap retroactively, producing a baseline eval suite that reflects the full intended behavior of the agent instruction layer.

This audit also serves as a forcing function to surface any ambiguity or inconsistency in the current agent instructions — behaviors that are implicitly assumed but not explicitly specified, or behaviors that have drifted from their original intent.

## Scope

The audit should cover all agent instructions that specify or imply autonomous behavior, including but not limited to:

- Tool invocation rules (which tools to call, under what conditions, in what order)
- Observation write behavior (when to write, what to write, deduplication logic)
- Entity resolution and disambiguation logic
- Retrieval and context assembly pipelines
- Session and authentication handling
- Error handling, retry, and fallback behavior in agentic paths
- Multi-step or chained tool flows
- Any conditional or branching logic driven by model output

## Process

1. **Inventory** — enumerate all agent instruction files, system prompts, and configuration that governs agentic behavior
2. **Behavior extraction** — for each instruction set, extract a list of discrete intended behaviors (input condition → expected action/output)
3. **Gap analysis** — cross-reference extracted behaviors against existing evals to identify uncovered cases
4. **Eval authoring** — write evals for each uncovered behavior; prioritize by risk surface (write operations and mutations first, retrieval second, read-only last)
5. **Validation** — confirm all new evals pass against current behavior; flag any that reveal a discrepancy between specified and actual behavior as a separate bug
6. **Documentation** — update or create a behavior registry mapping each agent instruction to its corresponding eval(s)

## Acceptance Criteria

- [ ] All agent instruction files inventoried and listed in this issue or a linked document
- [ ] Behavior extraction complete for each instruction set
- [ ] Gap analysis documented
- [ ] Evals authored for all identified behaviors
- [ ] All new evals passing in CI
- [ ] Any discrepancies between specified and actual behavior filed as separate issues
- [ ] Behavior registry created linking instructions to evals

castor-agent:
## Design decision recorded — process scoped, ready for execution

[...]

PR #1729, opened 2026-06-22
feat(eval-harness): relationship lifecycle eval + unify tool-endpoint registry (#1708)

### LINKED ISSUE #1708: Add agentic eval suite for the relationship lifecycle (create/batch/delete/restore + snapshot/list)
Surface: the dedicated relationship MCP tools. Today edges are created only as a store side-effect and verified via read-side relationship.exists; the tools' own contracts are unexercised: create_relationship (validTypes allowlist, content-hash idempotent dedup), create_relationships (per-item partial-success {created_count,error_count,success:false}), delete_relationship (discovery-guard 404 + hint when no live edge, is_live flip), restore_relationship (priority-1001 over 1000), get_relationship_snapshot (null when top observation is a deletion), list_relationships (default is_live filter, include_deleted, direction/type filter, total_count, tenant isolation).

Risk high: wrong-type acceptance or silent dedup collision corrupts the graph; batch partial-failure is a silent-loss vector (agent treats it as full success); soft-delete liveness drift; list is the type-discovery feeder for delete and a tenant-isolation surface (prior security advisory).

Eval to author: seed two entities and two edges (works_at, related_to). (1) create_relationship works_at -> assert live; recreate identical -> idempotent (list count stays 1); invalid type -> error, zero edges. (2) create_relationships batch of 3 (2 valid, 1 invalid type) -> created_count==2, error_count==1, success==false, errors[].index points at bad item; all-valid batch -> success==true. (3) delete works_at with correct type -> default list excludes it, include_deleted:true includes it; delete with a non-existent type for the pair -> 404 RESOURCE_NOT_FOUND with hint mentioning list_relationships, unrelated live edge untouched. (4) restore -> default list shows it live again; restore a live edge -> no-op success. (5) get_relationship_snapshot -> provenance for live, null after delete. (6) list direction=outgoing/incoming disjoint sets; total_count present.

Deterministic assertions: relationship.exists/relationship.count for the live set per operation; tool_result.matches for batch {created_count,error_count,success} and the 404 hint; deleted edge excluded by default, present under include_deleted. Depends on assertion-primitives (mcp_tool result-shape). PII-free.

---
*P0 · Tools: create_relationship, create_relationships, delete_relationship, restore_relationship, get_relationship_snapshot, list_relationships · Part of the eval-coverage backfill (CA audit 2026-06-22). Umbrella: #230. Coverage map: `coverage_map` ent_3d1d72dd3c5a0d9086d45317.
- **Depends on:** #1703 (assertion primitives) — needs new assertion type(s) before this eval can be written.*

markmhendrickson:
Adds an agentic eval for the dedicated relationship MCP tools' own contracts (graph-mutation, data-corruption class) — previously only edges-as-store-side-effect were covered via read-side `relationship.exists`. Part of the backfill (umbrella #230); runs in the `eval_scenarios` lane.

## Eval
`create_relationship`, `create_relationships`, `delete_relationship`, `restore_relationship`, `list_relationships` — seeds two entities, creates `works_at`, batch-adds `related_to` + `member_of`, deletes + restores `works_at`. Asserts via #1703 `tool_result.matches`: batch `{created_count:2, error_count:0, success:true}`, restore success; `relationship.exists` for the live set; per-type `mcp_tool.invocations(arg_subset)` anchoring delete/restore to `works_at`.

## Registry unification (the meaningful change)
Making these tools cassette-executable required adding them to **both** `TOOL_ENDPOINTS` (path) **and** `NEOTOMA_TOOL_NAMES` (execution gate) — two lists that silently drift (the eval failed first because the batch never executed against the server). Refactored `NEOTOMA_TOOL_NAMES` to **derive from `Object.keys(TOOL_ENDPOINTS)`** so a new tool needs ONE entry and they can never diverge. This unblocks every remaining tool eval.

## Probe finding
A `create_relationships` batch with an invalid `relationship_type` is rejected **whole** at the Zod-enum layer (400), NOT per-item partial-success as the issue text assumed — the eval asserts the real all-valid success contract.

## Verification
Passing in replay; full suite **19 passed / 0 failed / 8 skipped**. No `.test.ts` → catalog unaffected. Closes #1708.

### DIFF
--- a/packages/eval-harness/cassettes/relationship_lifecycle__stub__replay-only.cassette.json
+++ b/packages/eval-harness/cassettes/relationship_lifecycle__stub__replay-only.cassette.json
@@ -1,0 +1,66 @@
+{
+  "meta": {
+    "format_version": 1,
+    "scenario_id": "relationship_lifecycle",
+    "provider": "stub",
+    "model": "replay-only",
+    "instruction_profile": "auto",
+    "recorded_at": "2026-06-22T00:00:00.000Z",
+    "cost_usd": 0
+  },
+  "user_prompt": "Link Rel Alice to Rel Co (works_at), batch-add related_to and member_of, then delete and restore the works_at edge.\n",
+  "system_prompt": "You are an assistant integrated with the Neotoma memory system. Use the dedicated relationship tools (create_relationship, create_relationships, delete_relationship, restore_relationship, list_relationships) to manage the graph. Check list results for liveness.\n",
+  "tool_calls": [
+    {
+      "name": "create_relationship",
+      "input": {
+        "source_entity_id": "ent_8febc48e98321f08aebb3d53",
+        "target_entity_id": "ent_643bc21ccba68c38d78fb73c",
+        "relationship_type": "works_at"
+      },
+      "sequence": 0
+    },
+    {
+      "name": "create_relationships",
+      "input": {
+        "relationships": [
+          {
+            "source_entity_id": "ent_8febc48e98321f08aebb3d53",
+            "target_entity_id": "ent_643bc21ccba68c38d78fb73c",
+            "relationship_type": "related_to"
+          },
+          {
+            "source_entity_id": "ent_643bc21ccba68c38d78fb73c",
+            "target_entity_id": "ent_8febc48e98321f08aebb3d53",
+            "relationship_type": "member_of"
+          }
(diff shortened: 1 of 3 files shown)


neotoma-agent:

Failed to authenticate. API Error: 401 Invalid authentication credentials

PR #1744, opened 2026-06-22
feat(cli): per-agent AAuth JWK path override (NEOTOMA_AAUTH_PRIVATE_JWK_PATH)

markmhendrickson:
## Why
The CLI signer (`loadCliSignerConfig` / `hasCliAAuthKeypair`) loaded the private key **only** from the shared `~/.neotoma/aauth/private.jwk`, so every caller on a host signed as **one** identity. This blocks per-agent AAuth for a multi-daemon swarm — each agent needs to sign as itself.

## What
Honor **`NEOTOMA_AAUTH_PRIVATE_JWK_PATH`** in the CLI signer so a per-agent caller can point at its own key. This is the **same env var the proxy signer already supports** (`src/proxy/aauth_client_signer.ts:138`) — closing an inconsistency, not adding a new knob.

- New `resolveCliPrivateJwkPath()` (env override → resolved path, else the default).
- Used in `loadCliSignerConfig` + `hasCliAAuthKeypair`.
- Pair with the existing `NEOTOMA_AAUTH_SUB` / `NEOTOMA_AAUTH_KID` claim overrides to fully assume a per-agent identity.
- Unit tests (5): default, absolute/relative override, full load via override, absent-key → signing-disabled.

## Notes
- No behavior change for the default case (env unset → same path).
- `keygen` still writes to the default location; the override is read-side (the signing path), which is what the swarm daemons need.
- Local checks: eslint clean, `tsc --noEmit` clean for changed files, the 5 unit tests pass.

Unblocks the Ateles per-agent AAuth migration (the shared-identity limitation found while scoping it).

### DIFF
--- a/src/cli/aauth_signer.ts
+++ b/src/cli/aauth_signer.ts
@@ -40,6 +40,21 @@ export const AAUTH_PRIVATE_JWK_PATH = path.join(AAUTH_CONFIG_DIR, "private.jwk")
 export const AAUTH_PUBLIC_JWK_PATH = path.join(AAUTH_CONFIG_DIR, "public.jwk");
 export const AAUTH_CONFIG_PATH = path.join(AAUTH_CONFIG_DIR, "config.json");
 
+/**
+ * Resolve the private JWK path for CLI signing.
+ *
+ * Honors `NEOTOMA_AAUTH_PRIVATE_JWK_PATH` so a per-agent caller (e.g. a daemon
+ * signing as its own swarm identity) can point at its own key instead of the
+ * shared `~/.neotoma/aauth/private.jwk`. This matches the override the proxy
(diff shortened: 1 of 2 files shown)


neotoma-agent:

⚠️ **Panel review unavailable — agent credential failure.** The `vanellus` panel agent (spawned as `claude --print` by the Apis dispatcher) could not authenticate to the Anthropic API (401). This is an infrastructure issue, **not** a review verdict — the PR has not been assessed by the panel.

**Fix (host-side):** set a long-lived `ANTHROPIC_API_KEY` in the Apis daemon environment (`com.ateles.apis` plist or `ateles-private/.env`) and reload, or re-authenticate the Claude CLI for the daemon user. Daemons should use an API key, not an interactive OAuth session that expires.

_Posted by the Apis dispatcher; the operator has been paged to re-auth._

Issue #1667, opened 2026-06-17
MCP `/mcp` session instability: repeated "session is unknown on this API instance" across a session


markmhendrickson:
## Summary
Throughout an extended MCP client session, a large fraction of tool calls failed with a JSON-RPC error:

```
-32001 Service Unavailable: MCP session is unknown on this API instance.
If you run multiple replicas, enable sticky sessions for POST /mcp
(or route /mcp to a single instance). Otherwise restart the MCP client
so initialize runs again after a server restart.
```

The connection visibly flapped (connect → disconnect → reconnect) many times. Reads (`describe_entity_type`, `retrieve_entity_snapshot`, `retrieve_entity_by_identifier`) and writes (`store`) each failed multiple times before eventually succeeding on a later retry. Several operations only completed after 4+ attempts spread across turns.

## Impact
- Every persistence step became unreliable; the turn-lifecycle store contract is hard to honor when `store` intermittently 503s.
- Heavy retry cost and partial-progress ambiguity (caller cannot always tell whether a write landed).

## Likely cause (per the error text)
`POST /mcp` is load-balanced across multiple replicas without sticky sessions, so a session initialized on replica A is rejected by replica B.

## Suggested fix
- Enable sticky sessions for `POST /mcp`, or route `/mcp` to a single instance; or
- Make session state shared/externalized across replicas so any replica can serve an initialized session; and
- Consider a clearer server-initiated re-initialize signal so clients can transparently re-handshake instead of surfacing raw 503s.

## Environment
- Client: claude-code 2.1.178 (remote execution environment)

markmhendrickson:
### Still reproducing on v0.17.0 (cross-session, with a timeout signature)

Confirming this is **not resolved** despite the `resolved-in-release` (v0.16.0) note — reproduced today against a backend reporting `v0.17.0` (via `/health`).

**Scope: systemic, not a single client.** Inspected the stdio-proxy logs (`neotoma mcp proxy --aauth` → streamable HTTP `POST /mcp`) across **6 concurrent client sessions today**. Every one degraded the same way: a clean `initialize` at startup (133 KB instruction blob loaded), some successful tool calls, then a cascade of failures surfaced to the harness as `Server unavailable`. Representative tallies (tool calls / failures):

| session | calls | failures | last event |
|---|---|---|---|
| A | 46 | 8 | `store failed … unavailable` |
| B | 30 | 12 | `store failed after 19s` |
[...]

Issue #1472, opened 2026-05-29
MCP session resets ('session unknown on this API instance') in Claude.ai and Claude Code cloud sessions


markmhendrickson:
## Symptom

Mid-session, every MCP tool call begins failing with:

```
-32001 Service Unavailable: MCP session is unknown on this API instance. If you run multiple replicas, enable sticky sessions for POST /mcp (or route /mcp to a single instance). Otherwise restart the MCP client so initialize runs again after a server restart.
```

The failure is session-wide: even read-only health checks like `get_session_identity` return it. Sometimes the client auto-reconnects (disconnect/reconnect cycles several times in one session), but it can also reach a hard state that does not self-heal across multiple turns.

## Context

- Occurs in **Claude.ai** and **Claude Code cloud / web** sessions connecting to the operator Neotoma MCP endpoint over streamable HTTP (`POST /mcp`).
- Observed repeatedly on 2026-05-29: the server disconnected and reconnected several times, then entered a `session unknown` state that persisted across multiple consecutive turns and 6+ calls, including the read-only `get_session_identity` preflight and `store` writes.
- Blocks work mid-task: in-flight tool calls fail, and the turn-lifecycle store protocol cannot complete its closing store.

## Likely root cause

MCP session affinity is not preserved across replicas behind the load balancer: a `POST /mcp` lands on a replica that has no record of the `initialize`d session (LB round-robin, or a server restart dropped in-memory session state). The server surfaces this as a generic 503 rather than a signal the client can act on.

## Proposed fixes

- **Sticky sessions** for `POST /mcp` at the load balancer (route a given MCP session to one replica), **or**
- **Externalize/share MCP session state** across replicas (e.g. shared store) so any replica can serve any session, **and/or**
- Return a **typed re-initialize signal** (a distinct error code the client recognizes) so the harness can transparently re-run `initialize` instead of failing the turn.

## Acceptance criteria

- A long-running Claude.ai / Claude Code cloud session no longer hits `session unknown` after a replica change or server restart, or recovers automatically without surfacing failed tool calls to the user.

markmhendrickson:
### Recurrence 2026-06-09 (Claude Code cloud session)

Still unresolved. Hit the same `-32001 Service Unavailable: MCP session is unknown on this API instance` repeatedly during a single working session against the operator MCP over streamable HTTP (`POST /mcp`).

New observations that may help narrow the root cause:

[...]

PR #1750, opened 2026-06-23
fix(proxy): recover MCP sessions across backend restarts/replicas (#1667, #1472)

markmhendrickson:
## Problem

The stdio MCP proxy (`neotoma mcp proxy`) drops the **entire client session** when the downstream `/mcp` backend restarts (single instance) or a request routes to a non-sticky replica. The server replies `503 … session is unknown on this API instance`, or the request hangs through the restart window. Once it hits the hard state, the harness marks the server `unavailable` and does not reconnect for the rest of the session.

This is the failure tracked by #1667 and #1472. It was observed **still reproducing on v0.17.0** (post the v0.16.0 `resolved-in-release` note), across multiple concurrent client sessions, including a `store failed after 19s` timeout signature.

The previous recovery (added earlier) re-initialized and retried **once**, and **only** on the 503 — so:
- a retry that lands on *another* fresh instance failed with no further attempt, and
- the **timeout / connection-refused** window during a restart was never recovered (it surfaced as a generic 502 / a hang).

## Fix

Hardens the proxy dispatch so a backend restart becomes invisible to the client:

- **Bounded re-initialize + retry loop** (`DEFAULT_MAX_ATTEMPTS=4`) with exponential backoff (`backoffMs`: 300/600/1200/2000 ms) — a retry landing on another fresh instance re-handshakes again.
- **Transport errors *and* timeouts** (not just the 503) are treated as a restart window: drop the session, back off, replay the cached `initialize`, retry.
- **Per-request timeout** (`DEFAULT_REQUEST_TIMEOUT_MS=15s`) so a hang fails fast into a retry instead of stalling past the harness's own MCP timeout (the observed 19s).
- **`initialize` is never auto-retried** — the client owns the handshake.

Tunable via `requestTimeoutMs` / `maxRetries` on `ProxyConfig`, or env `NEOTOMA_MCP_PROXY_TIMEOUT_MS` / `NEOTOMA_MCP_PROXY_MAX_ATTEMPTS`.

Dispatch logic is extracted into `dispatchCore(deps, …)` with an injected transport for testability; the production `dispatchMessage` wraps it with `sendDownstream`.

## Tests

New `src/proxy/mcp_stdio_proxy.test.ts` (11 cases): `isRecoverableMcpSessionLostError`, `backoffMs`, `withTimeout`, and `dispatchCore` scenarios — session-loss recovery, restart/timeout recovery, **exhaustion emits a JSON-RPC error (never hangs)**, and the no-retry-for-initialize contract. `vitest` green, `tsc --noEmit` clean, `eslint` clean.

## Note (defense-in-depth, not a substitute)

This makes clients resilient to restarts but does not stop the restarts. In the operator deployment the root cause is separate: the public `/mcp` backend is served by a **file-watching auto-restart dev runner**, so incidental file changes restart it and wipe the in-memory session maps (`mcpTransports` / `mcpServerInstances` in `src/actions.ts`). That deployment fix is being handled operationally; this PR ensures a restart no longer takes the client session down with it.

### DIFF
--- a/src/proxy/mcp_stdio_proxy.test.ts
+++ b/src/proxy/mcp_stdio_proxy.test.ts
@@ -21,10 +21,7 @@ const baseConfig: ProxyConfig = {
   autostart: false,
 };
 
-function jsonResponse(
-  obj: unknown,
-  opts: { status?: number; sessionId?: string } = {}
-): Response {
+function jsonResponse(obj: unknown, opts: { status?: number; sessionId?: string } = {}): Response {
   const headers: Record<string, string> = { "content-type": "application/json" };
   if (opts.sessionId) headers["mcp-session-id"] = opts.sessionId;
   return new Response(JSON.stringify(obj), { status: opts.status ?? 200, headers });
@@ -84,7 +81,10 @@ describe("isRecoverableMcpSessionLostError", () => {
       isRecoverableMcpSessionLostError(503, "MCP session is unknown on this API instance")
     ).toBe(true);
     expect(
-      isRecoverableMcpSessionLostError(503, "service unavailable: session is unknown on this api instance")
+      isRecoverableMcpSessionLostError(
+        503,
+        "service unavailable: session is unknown on this api instance"
+      )
     ).toBe(true);
   });
   it("ignores non-503 statuses and unrelated bodies", () => {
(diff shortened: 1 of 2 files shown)


neotoma-agent:

Credit balance is too low

PR #1768, opened 2026-06-24
feat(sandbox): serve Inspector at root + in-app pack picker

markmhendrickson:
## Why

You expected the Inspector to show at the sandbox root. It didn't: in sandbox mode `GET /` rendered a separate server-rendered landing page, and the Inspector lived only at `/inspector` (reached via a hash handoff). In every *other* mode (local/personal/prod) the Inspector SPA already serves at root — only sandbox intercepted it.

## What

- **`src/actions.ts`** — drop the `hosted_sandbox` HTML interception in `GET /` so it falls through to the Inspector SPA fallback in every mode. Agents/`curl` (`Accept: application/json` / `text/markdown`) still get the structured discovery payload. Removes the now-unused `buildRootLandingHtml` import.
- **`inspector/src/components/home/sandbox_pack_picker.tsx`** (new) — pack picker on the Inspector home, shown in sandbox modes. Reads the pack list from the root JSON (`sandbox_packs` / `sandbox_default_pack_id`), POSTs `/sandbox/session/new`, redeems the one-time code in place via the existing `redeemSandboxSession`, and reloads into the seeded workspace.
- **`inspector/src/pages/home.tsx`** — render the picker at the top of the home in `hosted_sandbox` / `local_sandbox`.
- **`docs/subsystems/sandbox_deployment.md`** — root now serves the Inspector + picker; handoff flow updated.

## Verification

Built the inspector SPA (`tsc -b` + Vite) and the server, booted the sandbox locally:
- `GET /` (HTML) → Inspector SPA shell (`<div id="root">`, title "Neotoma") — **not** the old "Served by Neotoma" landing.
- `GET /` (JSON) → discovery payload, `mode: sandbox`, 11 packs, default `generic` (feeds the picker).
- `inspector/dist` bundle contains the picker.

No React component-test infra exists in this repo (only `lib/*.test.ts`; no `@testing-library`), so I matched convention and verified via build + runtime rather than adding a new test dependency.

## Depends on
[#1766](https://github.com/markmhendrickson/neotoma/pull/1766) (pack seeding + tenant-scoped ids) for the *data* to actually populate after picking a pack. This PR is the UI/routing half; the picker creates the session correctly on its own, but a seeded workspace requires #1766. Recommend merging #1766 first (or together).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1789, opened 2026-06-25
fix(sandbox): build Inspector with root base path so it renders at /

markmhendrickson:
## Why

After #1768 served the Inspector SPA at the sandbox **root**, the deployed `sandbox.neotoma.io` came up **blank**. Console:

```
<Router basename="/inspector"> is not able to match the URL "/" ... so the <Router> won't render anything.
```

Root cause: `inspector/src/main.tsx` derives `ROUTER_BASENAME` from `import.meta.env.BASE_URL`, and the Dockerfile hardcodes `VITE_PUBLIC_BASE_PATH="/inspector/"`. So the bundled SPA has basename `/inspector`, which cannot match `/` → blank page. (My pre-merge checks missed it: `curl` only confirmed the shell HTML loads, and the Playwright e2e built the SPA with the default root base, so it passed locally but not in the Fly image.)

## Fix

- Make `VITE_PUBLIC_BASE_PATH` a Dockerfile **build ARG**, default `"/inspector/"` (no change for installs that reach the Inspector at `/inspector`).
- Override it to `"/"` for the sandbox via `fly.sandbox.toml` `[build.args]`, since the sandbox serves the SPA at root.

The server already serves `/assets/*` at root (`installInspectorRootStaticAssets`), so a root-base build resolves. This also aligns the Playwright e2e (default root base) with what the sandbox now ships.

## Verified live

Deployed to `neotoma-sandbox` and confirmed in a browser on the fresh `neotoma-sandbox.fly.dev` host: Inspector renders at `/` with the **pack picker** ("Start a sandbox session" + Generic starter → Start session), populated stat tiles, orientation note, v0.17.0. Shell now loads `/assets/index-*.js` (root base) instead of `/inspector/assets/*`.

## Follow-up (separate, pre-existing)

A returning visitor whose sandbox session has expired (e.g. after the weekly wipe) carries a stale `neotoma_sandbox_session` cookie, and sandbox auth returns `401` instead of falling back to the anonymous public user — so the Inspector 401-walls instead of showing the pack picker. Now that the Inspector (not a static landing page) is at root, this is user-visible. Worth a middleware fix: on an invalid/expired sandbox cookie, clear it and fall back to anonymous.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,10 +10,16 @@ RUN npm run build:server
 FROM node:20-alpine AS inspector-build
 WORKDIR /app/inspector
 ARG VITE_NEOTOMA_SANDBOX_UI=""
+# Vite base path = router basename (main.tsx derives ROUTER_BASENAME from
+# import.meta.env.BASE_URL). Default "/inspector/" for installs that reach the
+# Inspector at /inspector. The sandbox serves the SPA at the site ROOT, so it
+# must build with "/" (otherwise <Router basename="/inspector"> can't match "/"
+# and renders a blank page) — fly.sandbox.toml overrides this to "/".
+ARG VITE_PUBLIC_BASE_PATH="/inspector/"
 COPY inspector/package*.json ./
 RUN npm ci
 COPY inspector/ ./
-RUN VITE_PUBLIC_BASE_PATH="/inspector/" \
+RUN VITE_PUBLIC_BASE_PATH="$VITE_PUBLIC_BASE_PATH" \
     VITE_NEOTOMA_SANDBOX_UI="$VITE_NEOTOMA_SANDBOX_UI" \
     npm run build
 
--- a/fly.sandbox.toml
+++ b/fly.sandbox.toml
@@ -23,6 +23,10 @@ primary_region = 'lhr'
 
 [build.args]
   VITE_NEOTOMA_SANDBOX_UI = "1"
+  # The sandbox serves the Inspector SPA at the site root, so build it with a
+  # root base path → router basename "/" matches "/". (Default "/inspector/"
+  # would render blank at root.)
+  VITE_PUBLIC_BASE_PATH = "/"
 
 [env]
   NODE_ENV = "production"


neotoma-agent:
review:qa

**COMMENT**

This fix resolves a regression introduced in #1768: the sandbox now serves the Inspector SPA at the site root (`GET /` routes to Inspector), but the Dockerfile builds the SPA with `VITE_PUBLIC_BASE_PATH=/inspector/` by default, causing React Router to initialize with `basename="/inspector"` which cannot match incoming root-path requests, resulting in a blank page.

**Fix assessment:**

The change itself is correct and minimal: `VITE_PUBLIC_BASE_PATH` is now a Dockerfile build ARG with a sensible default (`/inspector/` — preserves existing behavior for non-sandbox deployments), and `fly.sandbox.toml` overrides it to `/` for the sandbox environment where the SPA is mounted at root. The server already serves static assets at root (per `installInspectorRootStaticAssets`), so asset resolution will work correctly.

**Coverage gaps — [NON-BLOCKING] Test coverage:**

1. **No regression test for the blank-page failure.** The fix assumes that when `VITE_PUBLIC_BASE_PATH="/"` and the SPA is served at root, the Inspector renders; when `VITE_PUBLIC_BASE_PATH="/inspector/"` and the SPA is served at root, it is blank. This configuration-driven divergence is not currently tested.

   **Suggested eval:**
   - Add integration fixture under `tests/fixtures/agentic_eval/` (e.g., `inspector_root_base_path_rendering.json`) that:
     - Replays a sandboxed Inspector root-path request (`GET / → 200 HTML`, not blank)
     - Asserts the response body contains the rendered Inspector shell (e.g., a known DOM element like `<div id="root">` with content)
   - Alternatively, add a Playwright E2E test that boots the Inspector in the sandboxed configuration and verifies the home page renders (not blank)

2. **Build-arg override not tested.** Fly.io `[build.args]` entries like `VITE_PUBLIC_BASE_PATH = "/"` are not validated in CI. If the key name is mistyped or the value format is wrong, the override silently fails and the default is used (re-introducing the blank page).

   **Suggested check:**
   - Add a simple contract test under `tests/contract/` that parses `fly.sandbox.toml` and verifies:
     - `[build.args]` section exists
     - `VITE_PUBLIC_BASE_PATH = "/"` is present (exact match)
   - This guards against typos in the override key or value.

3. **No path-specific static-asset routing test.** The fix relies on the server routing `/assets/*` requests to Vite-built bundles even when the Router `basename="/"`. This is handled by `installInspectorRootStaticAssets` (per commit message), but there's no integration test verifying that a root-base build can correctly resolve asset URLs under the server's asset routing.

   **Suggested assertion (in addition to #1 above):**
   - In the agentic_eval or E2E test, verify that `GET /assets/index-<hash>.js` (or similar Vite chunk) returns 200 with JavaScript content, not HTML.

**No functional surface issue identified:** The change does not add new agent-facing surfaces or modify existing contracts — it's a build-configuration fix that only affects how the SPA's client-side router initializes. All prior tests (inspector_bundled_mount.test.ts, spa_path.test.ts) remain green because they do not depend on the Dockerfile build args.

**Recommendation:**
- This fix is safe to merge for sandbox deployment (the blank-page regression is real and the fix is sound).
- Add the regression test (agentic_eval fixture or E2E) in a follow-up PR to prevent recurrence.

**CI status:** All checks passed (as reported on the PR).

neotoma-agent:
review:pm

## Review Summary — PR #1789

**review:pm**
**APPROVE**

- [x] Change matches scoped intent (fix blank sandbox Inspector by making base path configurable)
- [x] No scope creep (two minimal files, no refactoring or unrelated work)
- [x] User-visible behavior verified (author confirmed live deploy renders Inspector at `/` with pack picker functional)
- [x] Backward-compatibility preserved (Dockerfile default `/inspector/` keeps existing installs unchanged)

**Scope alignment:**
Dockerfile ARG + sandbox-specific override in fly.sandbox.toml directly resolve the Router basename mismatch. Config-driven approach is low-risk and maintainable.

PM gate satisfied. Ready to land.

---

**Note:** This PR is already merged. The review above confirms that the change aligns with the stated intent, has no scope creep, and maintains backward compatibility for non-sandbox deployments while fixing the sandbox-specific rendering issue.

[pavo] acceptance_criteria: Change is narrowly scoped to the stated problem (configurable base path via build ARG with sandbox override), no scope creep, user-visible behavior verified live, backward-compat preserved.


_Posted by the Apis dispatcher on behalf of pavo — the panelist could not post its comment directly._

neotoma-agent:

**APPROVE**

## Aggregated Review Verdict — PR #1789

### Per-Lens Roll-Up

**PM (Pavo):** APPROVE
- Change matches scoped intent (configurable base path for sandbox)
- No scope creep (two minimal files)
- User-visible behavior verified live
- Backward compatibility preserved (default )

**QA (Phoenicurus):** COMMENT
- Fix is correct and minimal
- [NON-BLOCKING] Test coverage gaps:
  - No regression test for blank-page failure under root-base config
  - Build-arg override not validated in CI
  - No path-specific static-asset routing test
- Suggested evals: agentic_eval fixture or E2E test verifying Inspector renders at  (not blank) + contract test for  

### Blocking Findings
- **None.** All [BLOCKING] items: 0

### Merge Recommendation
Safe to merge. The sandbox deployment blank-page regression is real and this fix is sound. QA coverage gaps should be addressed in a follow-up PR to prevent recurrence.

### CI Status
✅ All required checks passed:
- baseline: SUCCESS
- frontend: SUCCESS
- agentic_evals: SUCCESS
- eval_scenarios: SUCCESS
- eval_combined: SUCCESS
- python_sdk: SUCCESS
- security_gates: SUCCESS
- site_export: IN_PROGRESS (advisory)

---

**Verdict: APPROVE** — PM gate signed off, QA identified non-blocking gaps (follow-up coverage work), no CI blockers, fix verified live on sandbox.

PR #1913, opened 2026-07-08
feat(auth): surface active environment in get_authenticated_user (#1905)

markmhendrickson:
## Problem (#1905)

On a fresh `npm install -g neotoma`, the CLI local transport defaults to **production** while the server/API default to **development**. So the MCP server writes to the dev DB, but an unflagged CLI query resolves to the empty, unauthenticated prod DB → returns nothing or 401 with **no signal** that the caller is on the wrong graph. An external evaluator lost real debugging time to this silent split.

## Fix (the non-breaking, high-value half)

Add `environment` to the `storage` block returned by `get_authenticated_user`, so an agent can **confirm which graph it's on before writing** — exactly the reporter's suggested fix (b): *surface the active env in a tool response*. The db path already differs (`neotoma.db` vs `neotoma.prod.db`); this makes the environment explicit and queryable.

Effect test asserts the field is present over HTTP.

## Scope note

This PR is the **observability** half. Flipping the CLI default to match the server (fix (a): make the default coherent / fail-loud) is a behavior change tracked separately under #1905 so it can be reviewed on its own risk profile.

Refs #1905

### DIFF
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,8 +61,8 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **528**
-- Backend and repo Vitest files: **493**
+- Total automated test files: **529**
+- Backend and repo Vitest files: **494**
 - Frontend Vitest files: **9**
 - Playwright spec files: **26**
 
@@ -72,7 +72,7 @@ flowchart TD
 | Vitest unit tests | 147 |
 | Vitest service tests | 36 |
 | Source-adjacent tests | 64 |
-| Vitest integration tests | 151 |
+| Vitest integration tests | 152 |
 | Vitest CLI tests | 65 |
 | Vitest contract tests | 14 |
 | Vitest security tests | 4 |
@@ -377,7 +377,7 @@ flowchart TD
 **Runner:** `vitest`
 **Command:** `npm run test:integration` or `npx vitest run tests/integration`
 **Requirements:** Database configured; remote-dependent subsets additionally need `RUN_REMOTE_TESTS=1`.
-**Files (151):**
+**Files (152):**
 - `tests/integration/aauth_attribution_stamping.test.ts`
 - `tests/integration/aauth_mcp_capability_parity.test.ts`
 - `tests/integration/aauth_mcp_initialize_admission.test.ts`
@@ -420,6 +420,7 @@ flowchart TD
 - `tests/integration/field_converters.test.ts`
 - `tests/integration/fixture_mcp_store_replay.test.ts`
 - `tests/integration/gdpr_deletion.test.ts`
+- `tests/integration/get_authenticated_user_environment.test.ts`
 - `tests/integration/graph_neighborhood_pagination.test.ts`
 - `tests/integration/graph_neighborhood_source_branch.test.ts`
 - `tests/integration/guest_invalid_bearer_routes.test.ts`
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -10546,6 +10546,14 @@ app.post("/get_authenticated_user", async (req, res) => {
       config.storageBackend === "local"
         ? {
             storage_backend: "local" as const,
+            // Surface the active environment + DB so an agent can confirm it is
+            // on the intended graph before writing (#1905). On a fresh npm
+            // install the CLI local transport defaults to production while the
+            // server defaults to development; exposing `environment` here lets
+            // a caller detect the split (e.g. writing to dev while an unflagged
+            // CLI reads empty prod) instead of discovering it via silent-empty
+            // results.
+            environment: config.environment,
             data_dir: config.dataDir,
             sqlite_db: config.sqlitePath,
           }
--- a/tests/integration/get_authenticated_user_environment.test.ts
+++ b/tests/integration/get_authenticated_user_environment.test.ts
@@ -1,0 +1,51 @@
+/**
+ * Effect test for #1905: `get_authenticated_user` surfaces the active
+ * environment so a caller can confirm which graph (dev vs prod) it is on
+ * before writing.
+ *
+ * Background: on a fresh npm install the CLI local transport defaults to
+ * production while the server/API default to development, so an unflagged CLI
+ * query can silently hit an empty prod DB. Exposing `environment` in the
+ * storage block lets an agent detect that split instead of discovering it via
+ * silent-empty results.
+ */
+import type { AddressInfo } from "node:net";
+import type express from "express";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+
+describe("get_authenticated_user surfaces active environment (#1905)", () => {
+  let server: ReturnType<express.Application["listen"]>;
+  let baseUrl: string;
+
+  beforeAll(async () => {
+    const { app } = await import("../../src/actions.js");
+    server = app.listen(0);
+    const addr = server.address() as AddressInfo;
+    baseUrl = `http://127.0.0.1:${addr.port}`;
+  });
+
+  afterAll(() => {
+    server?.close();
+  });
+
+  it("includes storage.environment in the response for a local backend", async () => {
+    const res = await fetch(`${baseUrl}/get_authenticated_user`, {
+      method: "POST",
+      headers: { "content-type": "application/json" },
+      body: "{}",
+    });
+    expect(res.status).toBe(200);
+    const body = (await res.json()) as {
+      user_id?: string;
+      storage?: { storage_backend?: string; environment?: string; sqlite_db?: string };
+    };
+    // Local (SQLite) backend must report which environment it resolved to.
+    expect(body.storage?.storage_backend).toBe("local");
+    expect(body.storage?.environment).toBeDefined();
+    expect(["development", "production"]).toContain(body.storage?.environment);
+    // The db path should be consistent with the reported environment (dev DB
+    // is neotoma.db; prod DB is neotoma.prod.db) — this is the split a caller
+    // needs to see.
+    expect(body.storage?.sqlite_db).toBeDefined();
+  });
+});


neotoma-agent:

neotoma-agent:
**BLOCKED**

PR #1913 references issue #1905 (`Refs #1905`, no `closes`/`fixes` keyword — resolved via Neotoma issue entity `ent_cd759129096a72dda3ca29f1`).

Parent issue gate_status is actively tracked (not legacy — `pm` already signed off, `current_owner` has advanced past Phase 2), so PR gate inheritance is enforced normally.

**Re-checked on `synchronize` event (2026-07-08T09:1x UTC, new commits pushed to `fix/1905-surface-active-env`): gate state unchanged since prior check.**

**Pending pre-impl gates:**

| Gate | Status | Owner |
|---|---|---|
| `pm` | ✅ signed_off | Pavo |
| `ux` | ✅ not_required (fast path: `bug` label) | — |
| `arch` | ❌ **pending** | Bombycilla |

`arch` is unsigned. Per gate pipeline rules, this PR cannot proceed to review assignment until it clears.

**To resolve:**
- Bombycilla can sign off the `arch` gate via Neotoma (`gate_status.arch → signed_off`), or
- The operator can waive it, or
- @markmhendrickson may issue `/confirm-gates-clear` to waive all unsigned pre-impl gates and re-trigger the PR pipeline (operator-only).

No reviewer assignment until one of the above occurs.

📎 Neotoma: issue `ent_cd759129096a72dda3ca29f1`

GATE_INHERITANCE: blocked

markmhendrickson:
**Scope note (addressing the gate-inheritance BLOCK)**

This PR uses `Refs #1905` (not `Closes`) on purpose: #1905 has two halves and this PR ships only the **observability** half — `get_authenticated_user` now surfaces the active `environment` so an agent can confirm which graph (dev vs prod) it is on before writing (the reporter's suggested fix (b)).

The other half — flipping the CLI local-transport default so it matches the server (fix (a): make the default coherent / fail-loud) — is a **behavior change** with a different risk profile and is deliberately deferred to a separate PR, so #1905 stays open after this merges. That is why this is `Refs`, not `Closes`.

This half is self-contained, tested (HTTP effect test), and non-breaking. Flagging explicitly so the gate reflects the intentional split rather than an omission.

neotoma-agent:
**COMMENT**

## Triage & Gate-Status Board

**Parent issue resolution:** PR body has no `closes #N`/`fixes #N`; resolved parent via `Refs #1905` in the PR description.

**Gate status** (source: swarm-spec embedded in neotoma#1905)

| Gate | Owner | Status |
|---|---|---|
| pm | Pavo | ✅ signed_off |
| ux | Accipiter | ⏭️ not_required (bug fast-path) |
| arch | Bombycilla | ✅ signed_off (unconditional) |
| impl | Gryllus (cicada) | 🔄 in_progress |
| pr_review | Vanellus | pending — this PR |
| qa | Phoenicurus | ✅ signed_off (test plan authored) |
| legal | Buteo | ✅ signed_off (unconditional) |

**Scope note:** #1905's spec calls for unifying the CLI/server env defaults on `development` (the full fix). This PR (#1913) ships only the observability half — surfacing `environment` in `get_authenticated_user`'s storage block — and explicitly defers the default-unification behavior change to a separate PR per its own "Scope note." That's a valid incremental slice of the same gated issue; it does not require its own separate gate cycle since it inherits #1905's already-signed-off pre-impl gates.

- **Owner:** current phase owner on #1905 is `cicada` (impl); next phase after this PR is `pr_review` (Vanellus).
- **Labels:** none applied by Lanius this pass (PR-opened protocol does not apply `lanius-triage`; that's issue-triage only).
- **Verdict:** `COMMENT` — pre-impl gates clear, reviewer assigned.

**Reviewer assignment:** Vanellus (pr_review gate owner) — no dedicated GitHub collaborator account exists for Vanellus in this repo (only `markmhendrickson`, `castor-agent`, `neotoma-agent` are collaborators), so a formal GitHub "Reviewers" API assignment isn't possible. Recording the routing decision here instead.

📎 Neotoma: issue neotoma#1905 (gate detail in swarm-spec section)

GATE_INHERITANCE: clear

neotoma-agent:
review:pm
**REQUEST_CHANGES**

**[BLOCKING] scope-mismatch: PR ships only the half of #1905 that the signed-off pm gate explicitly designates as insufficient on its own**

The `pm` gate on #1905 signed off on acceptance criteria that require:
- "A fresh dist/npm-origin install resolves the CLI's local transport and the MCP server/API to the **same** default environment, with no flags passed."
- Effect-verified fix per `fixed_means_behavior_verified_not_contract_accepted`: a test that an *unflagged CLI query against a freshly-installed, server-started local instance returns the data the server holds* — not merely that an env field is present in a response.
- Cross-surface parity per `cross_surface_contract_parity_tested_all_surfaces` across CLI local transport, MCP server startup, and the local HTTP/API path.

The same spec's **Out of scope** section explicitly names the approach this PR takes and rejects it:

> A warn-only fix that detects the split and tells the user to pass `--env dev` (issue's suggested option 2). This treats the symptom, not the defect... Reject as insufficient on its own; it MAY ship as a defense-in-depth addition **once the defaults are unified, not as a substitute for unifying them**.

PR #1913 adds `environment` to `get_authenticated_user`'s storage block (an observability signal an agent can *check*) and explicitly defers "flipping the CLI default to match the server" — the actual unification acceptance criterion — to a separate, currently unscoped follow-up. That is precisely the ordering the pm-signed-off spec forbids: this PR is the "insufficient on its own" half, shipping *before* the defaults are unified rather than as defense-in-depth *after*.

Net effect for the actual reported problem (silent empty/401 on a fresh npm install): unchanged. A caller must already know to inspect `storage.environment` and act on it — the fresh-install failure mode the issue reporter hit is not fixed by this PR. The change is safe and well-tested for what it does, but what it does is not the acceptance criteria that were signed off.

**Requested resolution (either is fine, but pick one explicitly):**
1. PM/Eng amend #1905's spec to formally split it into two tracked PRs (observability now, unification separately with its own acceptance criteria and priority), and this PR's description should reference that amendment rather than asserting the split unilaterally in the PR body — or
2. Hold this PR pending the default-unification change, or land them together.

Today the PR body just asserts "tracked separately under #1905" without a corresponding spec change — that's the scope agents should not self-authorize.

**[NON-BLOCKING] no unrequested scope creep otherwise** — the diff is minimal and additive (one field + one test + generated catalog counts); nothing beyond the environment field was touched. If the PM split above is formally approved, this PR's own contents are appropriately scoped to that narrower slice.

---
📎 Neotoma: see issue `ent_d164370477e24787cd90f119` (neotoma#1905, pm-signed-off spec) · PR `ent_290a164c7320c11d42e8e33c` (neotoma#1913)

neotoma-agent:
review:ux
**REQUEST_CHANGES**

**Task under test:** an agent/developer calling `get_authenticated_user` needs to confirm which graph (dev vs prod) it's about to read/write, per #1905's reported failure mode (silent empty/401 on a fresh install with no signal of the split).

**What's good:** `environment` returning the literal `"development"`/`"production"` string is the right naming choice — no new vocabulary to learn, directly comparable to `NODE_ENV`-style values a caller already knows, and it sits next to `sqlite_db`/`data_dir` so the three fields corroborate each other. The in-source comment in `src/actions.ts` explaining *why* the field exists is genuinely good — for a human reading the diff.

**[BLOCKING] docs-drift: canonical response schema for `get_authenticated_user` is not updated, and this PR widens the existing gap**

`docs/specs/MCP_SPEC.md` §3.22 (`get_authenticated_user`) documents the Response Schema as:

```typescript
{
  user_id: string;
  authenticated: boolean;
}
```

The entire `storage` block (`storage_backend`, `data_dir`, `sqlite_db`, and now `environment`) is absent from that contract. This isn't new debt this PR created outright — `docs/plans/me-and-get-authenticated-user-response-updates.md` (status: `pending`, 5 todos pending) already tracks the `storage` block as an under-documented addition to `GET /me` and `get_authenticated_user`. But this PR adds a *new field* to that undocumented block without touching the spec, so the gap an agent hits when discovering this surface gets wider, not narrower. An agent has no way to learn `environment` exists, what values it takes, or what to do about a mismatch except by reading `src/actions.ts` source — which most callers (including the target user profile: senior engineers who "will not tolerate ambiguity") won't do.

Actionable fix: add `storage.environment` (and, while touching this section, the sibling fields already shipped) to MCP_SPEC.md §3.22's Response Schema, one line each, values enumerated (`"development" | "production"`).

**[BLOCKING] silent-conditionality: `environment` is only present for the `local` storage backend, and that conditionality is undocumented anywhere a caller can discover it**

The diff adds `environment: config.environment` inside the `config.storageBackend === "local" ? {...} : {...}` branch only. A caller on a non-local (e.g. remote/hosted) backend gets a `storage` object without `environment` — no explicit `null`, no comment in the response itself, nothing that tells the caller "this field doesn't apply to you" vs. "this field failed to populate." Per #1905's own framing, the whole point of this field is to let an agent *notice* a problem; an agent checking `storage.environment` on a backend where the field is simply absent has no way to distinguish "you're not on local storage, this doesn't apply" from "something is wrong." The test file only covers the local-backend case (`get_authenticated_user_environment.test.ts`), so this branch has no test signal either.

Actionable fix: either (a) document explicitly in the spec that `environment` is `local`-backend-only and omitted otherwise, or (b) return `environment: null` explicitly for non-local backends so the field's absence is a decision, not a gap.

**[NON-BLOCKING] no actionable-hint surfaced when environment mismatches expectation**

The field is purely observational — it tells the caller *what* environment it's on, but nothing in the response helps the caller act on a mismatch (e.g. no `expected_environment` echo, no hint like "pass `--env dev` to override" surfaced in the payload itself). Per the PM-lens finding on this same PR, the actual failure mode (silent empty/prod DB) isn't fixed by this field alone — from a pure UX-of-the-surface angle, if this ships as an interim signal, consider whether the tool description (`docs/developer/mcp/tool_descriptions.yaml`) should say what a mismatch implies and what to do next, since that's the only place many MCP clients surface tool semantics to the calling agent without a spec read. Today `tool_descriptions.yaml`'s one-liner for `get_authenticated_user` doesn't mention `storage` or `environment` at all.

**[NON-BLOCKING] test only asserts presence, not the actionable case**

`tests/integration/get_authenticated_user_environment.test.ts` asserts `storage.environment` is defined and is one of the two allowed values — it doesn't assert anything about the mismatch-detection use case (e.g. that the value reflects `NEOTOMA_ENV`/config truthfully under a controlled env var). That's a correctness/QA-lens concern more than UX, flagging only because the entire point of the field per the PR description is agent-observable environment detection, and the test doesn't exercise that path.

---
**Acceptance checklist for this lens:**
- [ ] `docs/specs/MCP_SPEC.md` §3.22 Response Schema includes the `storage` block, `environment` field, and its allowed values
- [ ] Non-local-backend behavior for `environment` (omitted vs. explicit `null`) is decided and documented
- [ ] `tool_descriptions.yaml` entry for `get_authenticated_user` mentions the `storage.environment` field or links to the spec section

📎 Neotoma: issue [neotoma#1913](https://github.com/markmhendrickson/neotoma/pull/1913)

neotoma-agent:
review:qa
**REQUEST_CHANGES**

Scope: QA lens only (test coverage adequacy). Correctness/security are covered by the Claude GHA baseline.

- [x] New effect test exists and runs in CI (`agentic_evals`-adjacent lane: `tests/integration`, no explicit wiring needed since the glob already covers it)
- [x] Test catalog regenerated in the same PR (`npm run generate:test-catalog` output committed) — 528→529 files, 151→152 integration tests
- [ ] Response schema declared before handler change (spec-first)
- [ ] Test actually exercises both environment values (dev vs prod), the exact split the PR is fixing

[BLOCKING] contract-coverage: new `storage.environment` response field is not declared in `openapi.yaml`
`src/actions.ts` (`get_authenticated_user` handler) adds `environment: config.environment` to the JSON response, but `openapi.yaml`'s `/get_authenticated_user` 200-response schema (`components` inline object around the `storage` property) still only declares `storage_backend`, `data_dir`, `sqlite_db`. Per `.claude/rules/change_guardrails_rules.md` Touchpoint Matrix ("HTTP endpoint / request / response field" → `openapi.yaml` → contract mappings → handler → contract tests) and MUST #1/#6 ("Spec before handler" / "New response or error fields are declared in `openapi.yaml` first"), this is a required, not optional, step — and the Pre-PR Checklist item "New response fields declared in `openapi.yaml`; populated consistently across all code paths that return the schema" is unmet. Because the schema doesn't declare the field, no contract test (`tests/contract/contract_mapping.test.ts`, `tests/contract/contract_mcp_cli_parity.test.ts`) can catch a future regression that silently drops or renames it — only the one integration test added in this PR would catch that, and only by coincidence of being in the same PR.

[BLOCKING] edge-case-coverage: the new test never asserts the `production` branch, so it doesn't prove the fix
`tests/integration/get_authenticated_user_environment.test.ts` asserts `expect(["development", "production"]).toContain(body.storage?.environment)` — i.e. it accepts whichever value the ambient `NEOTOMA_ENV` happens to resolve to in CI and never pins the other branch. The PR's entire stated motivation (#1905) is a **dev/prod split**: an unflagged CLI defaulting to `production` while the server defaults to `development`. A test that doesn't independently set `NEOTOMA_ENV=production` (or `=development`) and assert the corresponding `environment` value **and** the matching `sqlite_db` suffix (`neotoma.prod.db` vs `neotoma.db`) does not regression-test the bug being fixed — it would pass identically if `environment` were hardcoded to a single value or wired to the wrong config field, as long as it happened to match ambient CI env. The repo already has the exact pattern needed for this: `tests/unit/config_sqlite_path.test.ts` uses `vi.stubEnv("NEOTOMA_ENV", "development"/"production")` + `vi.resetModules()` + fresh `import("../../src/config.ts")` to pin both branches. Recommend adding two cases (or a `describe.each`) to the new test — or a unit-level test alongside `config_sqlite_path.test.ts` — asserting `environment === "development"` ↔ `sqlite_db` ends in `neotoma.db`, and `environment === "production"` ↔ `sqlite_db` ends in `neotoma.prod.db`.

[NON-BLOCKING] test-placement: consider a unit test for the config→response wiring, not just an HTTP effect test
The current test is an integration-level HTTP round trip through `app.listen`, which is appropriate for confirming the field is present over the wire (as the PR description states), but pinning the environment-branch behavior described above is cheaper and more deterministic as a unit test against `config.ts` directly (per the pattern in `tests/unit/config_sqlite_path.test.ts`) than as an HTTP test requiring env-var-driven module reset across a live server. Either location is acceptable; just ensure the assertion in "edge-case-coverage" above lands somewhere.

[NON-BLOCKING] dead-branch note: the `storage_backend !== "local"` (`undefined` storage) branch is untested, but this is pre-existing and not a gap introduced by this PR
`config.storageBackend` is currently hardcoded to `"local"` in `src/config.ts`, so the ternary's `undefined` branch is unreachable in practice today. Not requiring new coverage for this PR, but flagging so it isn't assumed covered if `storageBackend` ever becomes configurable.

No functional-surface concerns beyond the above — the `docs/testing/automated_test_catalog.md` diff is purely generated and consistent with the new file.

neotoma-agent:
**REQUEST_CHANGES**

## Aggregated Review Verdict — PR #1913

Panel lenses: **pm**, **ux**, **qa** — all three independently returned `REQUEST_CHANGES`. Merge is blocked.

### Per-lens roll-up

| Lens | Verdict | Blocking findings |
|---|---|---|
| pm | REQUEST_CHANGES | 1 |
| ux | REQUEST_CHANGES | 2 |
| qa | REQUEST_CHANGES | 2 |

### Blocking findings

- **[BLOCKING] scope-mismatch (pm)** — The pm-signed-off spec on parent issue #1905 explicitly requires *unifying* the CLI's and server's default environments (effect-verified + cross-surface-parity tested), and explicitly rejects a warn/surface-only fix as "insufficient on its own" unless shipped as defense-in-depth *after* unification. PR #1913 ships exactly the rejected/deferred half — it adds an `environment` field an agent can check, but the underlying reported failure (silent empty/401 on fresh npm install) remains unfixed. Needs formal re-scoping into two tracked PRs, or hold/combine with the unification work.
- **[BLOCKING] docs (ux)** — `docs/specs/MCP_SPEC.md` §3.22 doesn't document the `storage` block or the new `environment` field. This PR widens a pre-existing gap rather than closing it.
- **[BLOCKING] undocumented backend gap (ux)** — `environment` is populated only on the `local` storage backend branch in `src/actions.ts`; its absence on other backends is undocumented and untested, so a caller can't distinguish "not applicable" from "broken."
- **[BLOCKING] contract-coverage (qa)** — The new `storage.environment` response field was never added to `openapi.yaml`'s `/get_authenticated_user` response schema, violating the "spec before handler" change-guardrail. No contract test would catch a future silent drop/rename.
- **[BLOCKING] edge-case-coverage (qa)** — The new integration test only asserts `environment` is *one of* `["development", "production"]` rather than pinning both branches via `NEOTOMA_ENV` override (the repo already has this pattern in `tests/unit/config_sqlite_path.test.ts` via `vi.stubEnv`/`vi.resetModules()`). As written, the test would pass even if the field were wired incorrectly, provided it coincidentally matched ambient CI env — it doesn't prove the fix works.

### Non-blocking notes

- No unrequested scope creep in the diff itself — it is a minimal, additive slice consistent with what the PR body claims (pm).
- No actionable hint in the response or `tool_descriptions.yaml` for what an agent should do on an environment mismatch (ux).
- Test placement preference, and a pre-existing untested dead branch not introduced by this PR (qa).

### Effect-verified-fix / cross-surface-parity gate

Per `fixed_means_behavior_verified_not_contract_accepted` and `cross_surface_contract_parity_tested_all_surfaces`: this PR does not close the reported bug's effect (silent empty/401 on fresh CLI query) at all — it only adds a diagnostic field on one surface (MCP, and only the `local` backend branch at that). Both policies are unmet independently of the pm scope finding.

### Gate action

- Parent issue: **#1905** (pm gate previously signed_off on the *unification* spec, current_owner `cicada`/impl).
- `pr_review` gate on #1905 → **changes_requested**. Routed back to Gryllus (impl) with this summary.
- Not advancing to Phase 4b. No merge performed — merge remains operator-gated per standing autonomy guardrail regardless of verdict.

### Requested resolution

Either (a) re-scope #1913 to also perform the CLI/server default-unification fix required by #1905's signed-off spec plus cross-surface parity tests, and add the `environment` field to `openapi.yaml` with a pinned dev/prod test — or (b) formally split #1905 into two tracked issues (observability-only vs. default-unification) with PM re-approving the narrower scope for #1913 before this PR can merge. Either path also needs the MCP_SPEC.md doc update and non-`local`-backend behavior documented/tested.

📎 Neotoma: [issue #1905](https://neotoma.markmhendrickson.com/entities/ent_d164370477e24787cd90f119) · [issue #1913](https://neotoma.markmhendrickson.com/entities/ent_290a164c7320c11d42e8e33c)

PR #1914, opened 2026-07-08
docs(hooks): document the Bash-tool gap in path-based deny rules (#1907)

### LINKED ISSUE #1907: Document (and optionally ship a reference hook for) the gap where Read/Glob/Grep deny rules don't stop raw Bash cat/grep/find
## Problem

When a user scopes a local Neotoma MCP setup to avoid reading sensitive sibling directories, declarative Read/Glob/Grep deny rules in the harness do **not** stop an agent's raw `cat`/`grep`/`find` via the Bash tool from reaching the same paths. A path-based deny list covers the file tools but not shell commands that read files.

An external evaluator hardening a personal instance alongside sensitive folders (accounting, legal, equity, client files) had to write their own PreToolUse hook (fail-closed, parses the Bash command and blocks reads into denied paths) to close this gap.

## Impact

Users who rely on declarative deny rules for data isolation may believe sensitive paths are protected when a Bash-tool read still reaches them. This is a security-relevant expectation gap in the setup guidance.

## Suggested fixes

1. **Document the gap explicitly** in the hooks/integration docs (`docs/integrations/hooks/`) — declarative deny rules cover file tools, not raw Bash; recommend a PreToolUse hook for path-based isolation.
2. Optionally **ship a reference PreToolUse hook** for common harnesses (Claude Code, Cursor) that fail-closed on reads into a configured deny list.

Reported by an external evaluator.

## Swarm specification

### Product / Scope (PM)

**Problem.** Declarative Read/Glob/Grep deny rules in the harness only gate the file-reading *tools* — they do not intercept raw shell commands (`cat`, `grep`, `find`, etc.) issued via the Bash tool. A user who scopes a local Neotoma MCP setup to keep an agent out of sensitive sibling directories (accounting, legal, equity, client files) may reasonably believe those deny rules provide isolation, when in fact a Bash-tool call can still read the same paths. This is a documentation/expectation gap, not a code defect in Neotoma itself — the harness behaves as designed, but the isolation guarantee implied by deny rules is overstated in current guidance.

**In scope**
- Update `docs/integrations/hooks/` (or the nearest canonical hooks/integration doc) to explicitly state: declarative Read/Glob/Grep deny-list configuration covers file tools only; it does **not** stop raw shell reads via Bash. Recommend a `PreToolUse` hook as the correct mechanism for path-based isolation guarantees.
- Optionally: ship one reference `PreToolUse` hook (starting with Claude Code, since that's the reporter's harness) that fail-closed parses the Bash command and blocks reads into a configured deny-path list. Treat this as a documented example/template, not a guaranteed security boundary for all possible Bash invocations (shell quoting/obfuscation edge cases are a known limitation to disclose, not solve, in v1).

**Out of scope**
- Building a fully robust, un-bypassable Bash-command sandboxing system. Shell parsing to catch every obfuscation (aliases, encoded paths, `$(...)`, etc.) is an open-ended arms race; the reference hook is a best-effort template, explicitly documented as such.
- Cursor or other harness-specific hook implementations beyond Claude Code, unless a maintainer wants to extend scope later.
- Any change to Neotoma's own MCP tool surface, schema, or API — this issue is documentation plus an optional standalone harness-config artifact that lives outside Neotoma's server contract.

**Acceptance criteria**
- [ ] `docs/integrations/hooks/` (or equivalent) explicitly documents that path-based Read/Glob/Grep deny rules do not cover raw Bash shell reads, with a concrete example of the gap (e.g., a denied `Read` on `../accounting/` still reachable via `Bash: cat ../accounting/file.txt`).
- [ ] Docs recommend a `PreToolUse` hook as the correct mechanism when Bash-reachable path isolation is required, and link to (or include) a working example.
- [ ] Doc change explicitly notes the known limitation (shell parsing can't catch every obfuscation) so the security boundary isn't overstated a second time.
- [ ] **Effect-verified, not contract-verified (hard gate):** whichever closure path is chosen (docs-only or docs+hook), the claim that "the gap exists" and, if shipped, the claim that "the hook closes it" must both be demonstrated by an actual observed run (repro command executed, or hook test executed) — not asserted from documentation prose or from a schema/config validating alone. A markdown claim is not evidence.
- [ ] **If a reference hook ships:** it fails closed (blocks by default on parse ambiguity, not allows), is scoped to a configurable deny-path list, and includes a test demonstrating it actually blocks a `cat`/`grep`/`find` into a denied path, driven via that surface's natural invocation shape (Claude Code's actual `PreToolUse` stdin/stdout JSON contract, not a hand-rolled shortcut). This is a single-surface capability today (no MCP tool/CLI/REST route exposes it), so cross-surface parity testing applies only if/when a second harness (e.g. Cursor) is documented — do not defer the Claude Code test waiting for a surface that isn't in scope.
- [ ] **Docs-only closure is an equally valid outcome of this spec** — the hook is explicitly optional per the original issue's suggested fixes. If maintainers defer the hook, the doc-only path still requires the gap repro to be actually run once at authoring time (see QA's "Docs-only path" checklist) and recorded as such in the PR description; it does not require the hook-specific criteria above.

**Priority & sequencing.** High-value, low-risk fix: single-digit-file doc change plus an optional, isolated hook script. No dependency on other in-flight work. Unblocks: closes a real trust/expectation gap for any user relying on deny rules for sensitive-data isolation — worth prioritizing ahead of larger feature work given it's a security-adjacent documentation correction reported by an external evaluator.

**Open questions**
- None blocking. Whether to ship the reference hook now or defer to docs-only is an implementation-sequencing call arch/impl can make — both paths satisfy this spec's acceptance criteria (hook-specific criteria only apply if the hook ships).

**Routing note (non-blocking).** This issue is `label:bug`-shaped (a gap between claimed and actual behavior) and its `gate_status.ux` was fast-pathed to `"skipped"` with `current_owner` already advanced to `bombycilla` outside this additive-spec pipeline. Per the interface-surface override in my own operating rules: this change touches no schema field, MCP tool, CLI command, error code, or relationship type — it is docs plus a standalone harness-side script with no Neotoma server contract surface — so the bug fast path skipping the formal arch gate is correctly scoped here; there is no bypass concern. Flagging only because the UX section's content (naming, error copy, doc structure) was produced *outside* the officially-recorded gate track — worth reconciling `gate_status`/`owner_history` bookkeeping so the audit trail matches what actually happened, but this does not change or block the substance of any section above.

### Design / UX

**User goal.** A developer/operator who has configured Read/Glob/Grep deny rules to isolate sensitive sibling directories needs to (a) discover — before an incident, not after — that those rules don't cover Bash, and (b) close the gap with a hook whose behavior and failure modes are legible at a glance.

**Naming**
- Reference hook file: `bash_path_guard.py` under `docs/integrations/hooks/examples/` (snake_case per repo file-naming convention). Name describes *what it intercepts* (Bash) and *what it enforces* (path denial) — a developer scanning a directory of example hooks matches on filename, not by opening every file.
- Config key: `denyPaths` (array of globs/paths) — reuses the mental model already established by the harness's own `Read`/`Glob` deny-rule config. Do not invent a parallel vocabulary (e.g. `blockedDirs`) that forces a second naming scheme for the same concept.
- Hook registration name in harness settings: `bash-path-guard` — short, greppable, distinguishes it from other `PreToolUse` hooks a user may already have registered.

**Interaction / flow**
1. User has existing `Read`/`Glob`/`Grep` deny rules configured and reasonably infers they're protected.
2. Agent issues `Bash: cat ../accounting/ledger.csv` (or `grep`, `find`, etc.).
3. **Without the hook (today, undocumented):** command silently succeeds — file contents return to agent context. No error, no warning, isolation just doesn't hold. This silent-failure case is the single most important thing the docs must make visceral — most users won't read a caveats paragraph; they need to see the failure reproduced.
4. **With the reference hook installed:** hook parses the Bash command pre-execution, resolves referenced paths (including relative-path resolution from `cwd`) against `denyPaths`, and on a match blocks the call before it reaches the shell.

**Error message — clean match (must be actionable, not a bare refusal)**

```
Blocked: this Bash command would read a path matched by denyPaths.

  command:  cat ../accounting/ledger.csv
  matched:  ../accounting  (resolves to /Users/you/accounting)
  rule:     denyPaths[0] in .claude/hooks/bash-path-guard.config.json

This command was blocked by the bash-path-guard PreToolUse hook, not by
Neotoma. If this path should be reachable, remove or narrow the matching
entry in denyPaths. See docs/integrations/hooks/bash_path_guard.md.
```

Requirements — each maps to a real failure mode from the issue:
- States *what* was blocked and *which config rule* matched — a user debugging "why did my legitimate command fail" needs the rule location, not just "denied."
- States *where the rule lives* (exact file path) — an actionable hint, not a dead end.
- Names the hook explicitly (`bash-path-guard`) so the block is never mistaken for a Neotoma-side error or a generic permission prompt — a common source of misattributed bug reports.

**Error message — parse ambiguity (must never look identical to a clean match)**

```
Blocked: could not safely determine which paths this command reads.

  command:  cat $(cat ~/.pathlist)

bash-path-guard fails closed on ambiguous commands it cannot statically
parse (see "Known limitations" in docs/integrations/hooks/bash_path_guard.md).
Rewrite the command to reference paths literally, or explicitly allow it
via denyPaths exceptions if you trust this path.
```

If the ambiguity and clean-match messages share one shape, users will conclude the hook is broken or over-blocking rather than behaving as documented. They must be structurally distinct (different heading, different fields) — not just a swapped detail line.

**Docs/examples the change must ship**
- [ ] `docs/integrations/hooks/README.md`: a callout, placed directly beside wherever deny rules are currently documented (not a separate advanced page), stating: *"Read/Glob/Grep deny rules do not stop raw shell reads via the Bash tool. If Bash-reachable path isolation is required, use a PreToolUse hook — see [bash_path_guard.md]."* The mismatch is between what the user configured and what they believe it does — the correction must sit next to the original claim, not be discoverable only after an incident.
- [ ] Same doc: a **reproduction example** — the exact denied `Read` config, followed by the exact `Bash: cat <path>` command that still succeeds — so the gap is demonstrated, not asserted.
- [ ] `docs/integrations/hooks/bash_path_guard.md`: quickstart (install + config), `denyPaths` schema, both error-message shapes above, and a **"Known limitations"** section listing concretely what it does *not* catch (`$(...)`, encoded/base64 paths, shell aliases, unresolved symlink traversal) — phrased as capability boundaries a user can weigh, not hedging language.
- [ ] A copy-pasteable Claude Code settings snippet (hook registration block) — the user must go from "reads the doc" to "hook is active" without translating instructions themselves.

**Discoverability**
- The gap-disclosure callout must appear in the doc a user actually reads when setting up deny rules — not in a changelog or an appendix nobody visits until after an incident.

- [ ] Deny-rule doc includes the explicit Bash-gap callout with a working repro (denied `Read`, successful `Bash cat`).
- [ ] `bash_path_guard.md` ships quickstart, config schema, both error-message shapes (clean-match block vs. parse-ambiguity block, structurally distinct), and a Known Limitations section.
- [ ] Error messages name the hook, the matched rule/config location, and the next action — never a bare "blocked."
- [ ] Install snippet is copy-pasteable for Claude Code without manual translation.

**Divergence note.** This issue's `gate_status.ux` is recorded as `"skipped"` via the `label:bug` fast-path (per `workflow_definition neotoma|feature`), and `current_owner` already advanced to `bombycilla` outside this additive-spec pipeline. That's a routing/bookkeeping question for the dispatcher — already filed as a non-blocking `plan_contribution` (`ent_00309a66272da70f3fe46bc7`), not corrected unilaterally here. It doesn't change the content above: the hook's naming, error copy, and doc structure are real UX surface regardless of which gate track records sign-off.

### Engineering

**Framing.** Docs-first with an optional standalone artifact — not a change to Neotoma's own hook plugin. The existing `packages/claude-code-plugin` hooks (`session_start.py`, `user_prompt_submit.py`, `post_tool_use.py`, `pre_compact.py`, `stop.py`) are explicitly best-effort / fail-open by contract (`packages/claude-code-plugin/hooks/_common.py:5-6`: "a failure here must never block the agent"). `bash-path-guard` has the opposite contract — fail-closed, blocking — so it MUST NOT be added to `plugin.json`'s `hooks` block or that package. It ships as an independent, opt-in example the user wires into their own project/user settings. Verified against current repo state: `docs/integrations/hooks/README.md` documents the five packages and links per-harness guides; `docs/integrations/hooks/claude_code.md` is the actual Claude Code guide (currently 52 lines, no deny-rule or `PreToolUse` content today) — this is the file a Claude Code user setting up deny rules actually reads, so the gap-disclosure callout must land in **both** files, not README alone.

**Files touched**

1. `docs/integrations/hooks/README.md` — add a short "## Deny rules do not cover Bash" subsection, placed after "## Packages" and before "## What every hook does" (matches current section order). Include the repro example inline (denied `Read` config + succeeding `Bash: cat` command) and link to `bash_path_guard.md`.
2. `docs/integrations/hooks/claude_code.md` — add the same callout (short form, cross-referencing the README subsection for the full repro) since this is the harness-specific guide readers land on when configuring Claude Code deny rules; the correction must sit where the original claim is made, per UX's discoverability requirement.
3. `docs/integrations/hooks/bash_path_guard.md` (new; snake_case per repo file-naming convention, consistent with `claude_code.md`, `codex_cli.md`) — quickstart, `denyPaths` config schema, both error-message shapes from the UX section, "Known limitations" section, and the copy-pasteable Claude Code settings snippet.
4. `docs/integrations/hooks/examples/bash_path_guard.py` (new) — reference hook implementation, if the team ships one (optional per PM; recommended in the same PR since the marginal cost is low and it is what makes the "effect test, not contract test" sign-off requirement satisfiable at all).
5. `docs/integrations/hooks/examples/bash_path_guard.config.example.json` (new) — sample `denyPaths` config matching the doc's quickstart.
6. `tests/unit/hooks/test_bash_path_guard.py` (new, Python) — **path correction from earlier draft**: there is no `tests/unit_python/` directory in this repo; the only existing Python-test precedent is `packages/client-python/tests/`. Since this hook is docs-adjacent (not a package), place its test under the existing top-level `tests/unit/` tree (which already contains `tests/unit/docs/` and `tests/unit/features/`), as `tests/unit/hooks/test_bash_path_guard.py`. Confirm the repo's Python test runner config (pytest discovery root/`testpaths`) picks up this path before finalizing; adjust only the directory if discovery requires it — the file must stay colocated under `tests/unit/`, not spawn a new top-level Python test root.
7. `docs/testing/automated_test_catalog.md` — regenerate via `npm run generate:test-catalog` (mandatory per repo rule and impl sign-off checklist) once the new test file exists, as the LAST step before opening the PR.

No `openapi.yaml`, `contract_mappings.ts`, MCP tool, or CLI command changes — confirmed this touches no Neotoma server contract surface. `git diff` for this PR should touch only `docs/` and the one new test file.

**Layering — why this lives outside `packages/claude-code-plugin`**

- The shipped plugin hooks are "reliability floor" instrumentation (session anchor, prompt capture, retrieval injection, tool-invocation logging, compaction/stop safety net — per `README.md`'s "What every hook does") — never blocking. Bolting a blocking, fail-closed security control onto that plugin would change its risk profile for every existing installer and conflate two different guarantees (observability vs. isolation).
- `bash-path-guard` is a **user-configured** control (the `denyPaths` list is inherently project/user-specific — accounting folders, legal folders, whatever the operator wants isolated). It does not know about Neotoma entities, sessions, or conversations, and must not import `packages/claude-code-plugin/hooks/neotoma_client` or depend on `_common.py`.
- Housing it under `docs/integrations/hooks/examples/` (docs-adjacent, not `packages/`) signals "copy this and adapt it" rather than "install this Neotoma package" — matching UX's naming decision (`bash-path-guard`, `denyPaths`) and PM's framing of it as a template, not a guaranteed boundary. It is also not registered in `neotoma hooks install --tool claude-code` (the CLI entry point documented in README) — it stays fully outside that installer flow, since it is opt-in and manually wired.

**Reference hook implementation shape (`bash_path_guard.py`)**

- Runtime: Python 3, stdlib only (`json`, `os`, `re`, `shlex`, `sys`, `pathlib`, `fnmatch`) — no third-party dependencies, so it drops into any Claude Code project without a package-install step.
- Reads the Claude Code `PreToolUse` hook payload from stdin (`tool_name`, `tool_input.command`, `cwd` for the `Bash` matcher).
- Registration: `matcher: "Bash"` only — scoping narrowly avoids any interaction with Read/Glob/Grep's own deny-rule handling.
- Command parsing: `shlex.split` to tokenize, then walk tokens for path-like arguments to known read commands (`cat`, `grep`, `less`, `more`, `head`, `tail`, `find`, `sed` in read mode, `awk`). Resolve each candidate path against the payload's `cwd` before matching against `denyPaths` globs (`fnmatch`/`pathlib.Path.match`).
- Ambiguity detection (triggers the fail-closed "could not safely determine" message): `$(...)`, backticks, `|`, `;`, `&&`, `||`, unresolvable `$VAR` expansion, or a `shlex.split` quoting error. Any of these short-circuits directly to the ambiguity-block path — no partial resolution attempted.
- Config loading order (first match wins): `NEOTOMA_BASH_PATH_GUARD_CONFIG` env var → `.claude/hooks/bash-path-guard.config.json` in the project root → missing (fail closed: block the command AND emit a loud stderr warning that the guard is unconfigured — a missing config is a setup bug, not a "nothing to protect" signal).
- Output contract: verify the exact current Claude Code `PreToolUse` block schema (exit code vs. JSON `{"decision": "block", "reason": "..."}` on stdout) against Claude Code's own hook documentation at implementation time, not from memory — the two UX message bodies map to the `reason` field verbatim once confirmed.
- No network calls, no Neotoma client import — the hook's entire trust boundary must be auditable by reading this one file.

**`denyPaths` config schema**

```json
{
  "denyPaths": [
    "../accounting/**",
    "../legal/**",
    "~/clients/**"
  ]
}
```

- Array of glob strings, resolved relative to the project root (directory containing `.claude/`) unless absolute or `~`-prefixed.
- Validate on load: reject the config (fail closed → block everything) if `denyPaths` is missing, not an array, or contains a non-string entry — malformed config must never silently degrade to "allow all."

**Build steps (order)**

1. Write `docs/integrations/hooks/examples/bash_path_guard.py` with the parsing/matching/blocking logic above.
2. Write `docs/integrations/hooks/examples/bash_path_guard.config.example.json`.
3. Write `docs/integrations/hooks/bash_path_guard.md` (quickstart, schema, both error shapes, Known Limitations, install snippet) — content already specified in the UX section; this step is transcription plus verifying the install snippet against current Claude Code hook-registration syntax.
4. Add the gap-disclosure callout + repro example to `docs/integrations/hooks/README.md` and the shorter cross-referencing callout to `docs/integrations/hooks/claude_code.md`.
5. Write `tests/unit/hooks/test_bash_path_guard.py`, driving the hook script via subprocess with the actual Claude Code `PreToolUse` stdin JSON shape (not a hand-rolled shortcut).
6. Run the new tests locally. Determine whether `docs/` Python files fall under any repo lint config (no root `.flake8` or `pyproject.toml` exists today — confirm via whatever lint tooling actually runs in CI, e.g. a `package.json` lint script or CI workflow step targeting Python files) rather than assuming coverage; if `docs/integrations/hooks/examples/*.py` is out of scope, note that explicitly in the PR rather than silently skipping lint.
7. Regenerate `docs/testing/automated_test_catalog.md` via `npm run generate:test-catalog`, then `npm run validate:test-catalog` — last step, immediately before opening the PR, per the impl sign-off checklist.

**Testing — effect-level, not contract-level**

Per the impl gate's "fixed means behavior verified, not contract accepted" policy:

- `denyPaths` match, clean command → hook blocks `cat ../accounting/ledger.csv`; assert the exact clean-match message shape (command, matched, resolved absolute path, rule, config path).
- `denyPaths` match via `grep`/`find` → same blocking behavior, confirming coverage of all three commands named in the issue title.
- No match → command passes through unmodified (guards against over-blocking).
- Ambiguous command (`cat $(cat ~/.pathlist)`) → blocks with the distinct ambiguity-message shape, structurally different from the clean-match shape.
- Malformed/missing config → fails closed (blocks), not open.
- Relative-path resolution from a non-root `cwd` (`../../accounting/file.txt`) → resolves correctly and matches.

Single-surface capability (a Claude Code `PreToolUse` hook script) — no MCP tool, CLI command, or REST route to parity-test against, so "test every exposing surface" is satisfied by testing this one surface via its natural stdin/stdout invocation shape.

**Out of scope for this PR (confirmed against PM)**

- Cursor-specific hook wiring (`packages/cursor-hooks` untouched).
- Any change to `packages/claude-code-plugin/.claude-plugin/plugin.json` or its existing five hooks.
- Any Neotoma server/MCP/CLI/`neotoma hooks install` change.

### QA / Test Plan

**Scope note.** No regression test applies — this is new capability (doc + optional new hook script), not a bugfix to existing behavior. All items below are new-branch edge cases; since there is no MCP tool/CLI/REST surface, no contract tests apply. Per the "no functional surface" rule: **if the reference hook does not ship (docs-only PR)**, this entire checklist reduces to `no functional surface — no eval required` for the code path, but the **doc-repro must still itself be effect-verified** (see "Docs-only path" below) — a markdown claim that a gap exists is not evidence the gap exists.

### Test artifact

- New file: `tests/unit/hooks/test_bash_path_guard.py` (Python — matches Engineering's corrected path; there is no `tests/unit_python/` directory in this repo, and this hook is docs-adjacent, not a package, so it does not belong under `packages/client-python/tests/`).
- Drives `docs/integrations/hooks/examples/bash_path_guard.py` via subprocess, feeding it the **actual Claude Code `PreToolUse` stdin JSON shape** (`tool_name`, `tool_input.command`, `cwd`) — not a hand-rolled shortcut. This is the one exposing surface, so this is the full "test every surface" obligation.
- Asserts on stdout `{"decision": ..., "reason": ...}` and process exit code, not on internal function calls — effect-level, matching the issue's own repro shape (`Bash: cat <path>` → blocked or not).

**Core blocking behavior (the bug this issue reports)**
- [ ] `denyPaths` match via `cat ../accounting/ledger.csv` → hook blocks; `reason` matches the clean-match message shape verbatim (command, matched path, resolved absolute path, rule index/source, config file path).
- [ ] `denyPaths` match via `grep -r secret ../legal/` → blocks with same message shape. (Covers `grep`, named in the issue title.)
- [ ] `denyPaths` match via `find ../equity -name '*.pdf'` → blocks with same message shape. (Covers `find`, named in the issue title.)
- [ ] Same three commands (`cat`/`grep`/`find`) run **without** the hook installed, against a harness configured with only a `Read` deny rule on the same path → command succeeds. This is the literal repro from the issue and the README callout — assert it actually reproduces *before* asserting the fix, otherwise the "fix" is unfalsifiable (a hook that always blocks would pass a blocks-only test suite).

**Non-blocking / no over-block (false-positive guard)**
- [ ] `cat ./README.md` (path not in `denyPaths`) → passes through unmodified; hook emits no block decision.
- [ ] `grep TODO src/**/*.ts` (glob outside denyPaths) → passes through.
- [ ] Command referencing a path that merely *contains* a denyPaths segment as a substring but isn't actually under that directory (e.g. `cat ../accounting-notes-shared/readme.txt` vs. denied `../accounting/**`) → passes through. (Guards against naive substring matching instead of proper glob/path matching.)

**Relative-path resolution (explicit issue repro scenario)**
- [ ] Hook invoked with `cwd` several directories below project root, command references the denied path via `../../accounting/file.txt` → resolves against the given `cwd` (not project root) and still matches `denyPaths`.
- [ ] Symlink case: document as a **known limitation** per Eng/UX (not solved) — add a test asserting the hook does NOT falsely claim protection here, i.e. explicitly test that a symlink-traversal bypass is possible today, so the doc's limitation claim is itself verified rather than aspirational.

**Ambiguity / fail-closed edge cases**
- [ ] `cat $(cat ~/.pathlist)` → blocks with the distinct **ambiguity** message shape (not the clean-match shape) — assert the two message shapes are structurally different (different `reason` text/fields), per UX's explicit requirement that they must never look identical.
- [ ] Command with backticks (`` cat `find . -name secret` ``) → ambiguity block.
- [ ] Command with a pipe (`cat ../accounting/x | grep foo`) → ambiguity block (pipe present, even though the left-hand path is also independently a denyPaths match — assert the ambiguity path wins, per Eng: "any of these short-circuits straight to the ambiguous-block path").
- [ ] Command with `&&`/`;`-chained commands where only the second reads a denied path → ambiguity block (not silently allowed because only the first command was checked).
- [ ] Malformed quoting that breaks `shlex.split` (unbalanced quote) → ambiguity block.
- [ ] Non-Bash tool calls (e.g. `tool_name: "Read"`) → hook is a no-op / not invoked for this matcher (registration is `matcher: "Bash"` only) — assert the hook script itself, if invoked with a non-Bash payload by mistake, still fails closed rather than crashing uncaught.

**Config edge cases**
- [ ] Missing config file (no `NEOTOMA_BASH_PATH_GUARD_CONFIG`, no `.claude/hooks/bash-path-guard.config.json`) → fails closed: hook blocks the command AND emits the documented loud stderr warning that the guard is unconfigured (assert both the block decision and the stderr content — silently blocking without the warning would look like an unconfigured no-op with no signal to the user).
- [ ] Config file present but `denyPaths` missing → fails closed (blocks).
- [ ] `denyPaths` present but not an array (e.g. a string) → fails closed (blocks).
- [ ] `denyPaths` array containing a non-string entry → fails closed (blocks), and the failure applies to the whole config (not a partial-apply of only the valid entries — partial-apply on a malformed list is itself a silent-degradation bug class).
- [ ] `NEOTOMA_BASH_PATH_GUARD_CONFIG` env var present and points to a valid file → takes precedence over the project-root default location (assert precedence order, not just that either individually works).

**Docs-only path (if the reference hook is deferred)**
- [ ] If PM/Eng choose not to ship the hook in this PR, the README repro example (denied `Read` config + succeeding `Bash: cat`) must still be validated by actually running it against a real Claude Code deny-rule config once, at authoring time, and the doc must state the harness/version it was verified against — an unverified claim of a security gap, shipped as documentation, is exactly the kind of unfalsifiable assertion this QA gate exists to catch. This verification does not need to be a committed automated test (no code ships), but it must be recorded (e.g. in the PR description) as a deliberate, auditable check — not skipped as "obviously true."

**Lint/catalog housekeeping (per Eng build steps 6–7)**
- [ ] `docs/integrations/hooks/examples/bash_path_guard.py` is covered by whatever lint config applies to `docs/` Python files (confirm scope via actual CI config, don't assume); CI lint passes.
- [ ] `docs/testing/automated_test_catalog.md` regenerated via `npm run generate:test-catalog` and validated via `npm run validate:test-catalog` after `tests/unit/hooks/test_bash_path_guard.py` is added — done as the last step before PR open, included in the same PR diff.

**Sign-off condition**
- Hook ships → all "Core blocking," "Non-blocking," "Relative-path," "Ambiguity," and "Config edge cases" boxes checked, with a green CI run on `tests/unit/hooks/test_bash_path_guard.py`, linked in the PR.
- Docs-only → the "Docs-only path" verification is recorded in the PR description, and the rest of this checklist is marked N/A with the reason `no functional surface — no eval required: docs-only change, no code ships`.
- QA does **not** sign off if the repro-first assertion (hook absent → command succeeds) is missing — without it, a green "blocks" test could just as easily reflect a hook that always blocks everything, not one that closes the specific gap described in the issue.

### Security / Arch

---

The Security/Arch section is above, delivered between the required fences. Summary: this change touches no Neotoma server contract surface (no `openapi.yaml`, MCP tool, CLI command, schema field, or DB table), so contract-first ordering, tenant isolation, and `idempotency_key` requirements are confirmed inapplicable — with explicit reasons stated rather than silently skipped. Binding requirements that do apply: no Neotoma client/credentials in the hook's trust boundary, no PII/secret leakage in block messages, and explicit disclosure of the hook's real security-boundary limitations (so the original overstatement isn't repeated). Signed off `arch` in the additive-spec sense, with one non-blocking follow-up noted for PR time (verify zero Neotoma imports in `bash_path_guard.py`).

### Legal / Compliance

**Jurisdiction basis.** Primary jurisdiction Spain (EU); regulatory regimes GDPR + ePrivacy (per `locale_profile:default`). Neotoma project licence: MIT (per `product_profile:Neotoma`). This change touches no Neotoma server contract surface (confirmed by Security/Arch), which narrows the applicable legal surface to three things: what dependency licence obligations the new artifact carries, whether the hook's own I/O can leak PII, and whether it touches any Neotoma credential/guest-token material.

**1. Dependency licensing**
- [ ] `bash_path_guard.py` is stdlib-only (`json`, `os`, `re`, `shlex`, `sys`, `pathlib`, `fnmatch` per Eng) — **no new third-party dependency is introduced**, so no license-compatibility audit against MIT is required for this PR. Confirm this stays true through implementation: if a reviewer later reaches for a third-party package (e.g. a glob library), re-trigger a licence check before merge — stdlib-only is a load-bearing constraint here, not incidental.
- [ ] The new doc files and example config carry no separate licensing question — they inherit the repo's existing MIT licence as documentation/example code, consistent with the rest of `docs/integrations/hooks/examples/`.
- [ ] No CC-NC, GPL/AGPL, or other copyleft/non-commercial terms enter the tree via this change. Nothing to redline here conditional on the stdlib-only constraint holding.

**2. PII / data-handling on public-effect surfaces**
- [ ] **Must-fix constraint on the error-message copy** (both shapes drafted by UX/Eng): the clean-match and ambiguity messages must echo only the **command text and path strings** the user/agent already supplied — never file *contents*. Since `denyPaths` is explicitly aimed at accounting/legal/equity/client files, a block message that pastes back a snippet of the denied file (e.g., during partial-parse debugging output) would itself be a data-handling failure, defeating the isolation goal the hook exists to provide. Confirm this is enforced in implementation/tests, not just implied by the message templates already shown.
- [ ] The hook's config (`denyPaths`) and its stderr "unconfigured guard" warning are local-only, stdout/stderr artifacts — no transmission to any third party, no logging sink beyond the local process per Eng's "no network calls" design. No GDPR data-processing-agreement or cross-border-transfer analysis applies since no personal data leaves the local machine through this artifact.
- [ ] This is a **harness-side, user-configured control**, not a Neotoma product feature processing operator/user personal data on Neotoma's behalf — Neotoma has no processor role here. No update to Neotoma's privacy policy or data-processing documentation is triggered by this change.
- [ ] Docs must retain the "Known limitations" framing exactly as UX specified (capability boundaries, not hedging) — from a claims-risk angle, understating what the hook catches would recreate the exact "isolation guarantee overstated" problem this issue reports. Legal concurs with UX/QA that the limitation must be actively verified (symlink-bypass test), not just asserted in prose.

**3. Guest-token / credential scope**
- [ ] Per Eng's design, `bash_path_guard.py` makes **no network calls** and **must not import** `packages/claude-code-plugin/hooks/neotoma_client` or any Neotoma API/MCP client. Confirmed: no `guest_access_token`, no AAuth credential, no Neotoma API key ever enters this artifact's process boundary. This is the correct posture — a fail-closed security control with a credential dependency would itself be a new attack surface and trust-boundary expansion beyond what this issue asks for.
- [ ] **Recommend converting Arch's non-blocking "verify zero Neotoma imports" follow-up into a checked build/CI step** (e.g., a simple grep/import-lint assertion on `docs/integrations/hooks/examples/bash_path_guard.py`), not just a manual PR-time check — a credential-scope guarantee stated only in docs is exactly the class of unfalsifiable claim this issue itself is about. This is a should-fix, not a blocker.
- [ ] No change to `neotoma hooks install` or any guest-token-issuing flow — confirmed out of scope per Eng, and legal has no additional requirement beyond what's already stated there.

**Escalation.** No item above requires escalation to qualified counsel — no contract, no regulatory filing, no IP dispute, and value/risk is well under the `locale_profile.escalation_currency_threshold` (this is a zero-cost internal docs/tooling change). Legal sign-off: **no blocking legal risk**, contingent on the two must-fix items above (no file-content echo in block messages; symlink-limitation claim actually tested) being carried through to implementation — both are already tracked as QA/Eng obligations in prior sections, so no new work item is introduced, only a legal-risk rationale attached to existing ones.

markmhendrickson:
## Problem (#1907)

Declarative `Read`/`Glob`/`Grep` deny rules in Claude Code do **not** stop a raw `cat`/`grep`/`find`/`ls` via the agent's Bash tool from reaching the same paths — the deny list only governs the dedicated file tools. A user scoping a local Neotoma install to keep an agent out of sensitive sibling directories (accounting, legal, client files) may believe those paths are protected when a Bash read still reaches them.

An external evaluator hit exactly this and had to write their own fail-closed PreToolUse hook.

## Fix

Document the gap explicitly in `docs/integrations/hooks/claude_code.md`, and provide a **fail-closed PreToolUse reference hook** that blocks Bash reads into denied paths — with its honest limits (computed paths, symlinks, `$(…)` substitution) and a recommendation to layer it with the declarative denies rather than rely on either alone.

Closes #1907

### DIFF
--- a/docs/integrations/hooks/claude_code.md
+++ b/docs/integrations/hooks/claude_code.md
@@ -50,3 +50,54 @@ Every hook above also accretes onto a single `conversation_turn` keyed by `(sess
 ## Coexistence with MCP
 
 MCP and the plugin share idempotency keys — `conversation-{session_id}-{turn_id}-user` etc. — so a user message captured by the plugin and then enriched by the agent via MCP lands on the same `agent_message` observation, not a duplicate.
+
+## Path-based isolation and the Bash-tool gap
+
+If you scope a local Neotoma install to a working directory and want to keep an
+agent out of *sensitive sibling directories* (accounting, legal, client files,
+etc.), be aware of an important limitation of Claude Code's declarative
+permission rules:
+
+> **Declarative `Read` / `Glob` / `Grep` deny rules do NOT stop a raw
+> `cat` / `grep` / `find` / `ls` run through the agent's `Bash` tool.** The
+> path-based deny list only governs the dedicated file tools; a shell command
+> that reads the same file bypasses it.
+
+So a `deny` entry like `Read(//Users/you/Private/**)` protects the `Read` tool,
+but `Bash(cat /Users/you/Private/secret.txt)` still succeeds. If a broad allow
+such as `Read(//Users/you/**)` is also present, that widens the exposure
+further (deny overrides allow for the file tools, but Bash is unaffected).
+
+### Recommended: a fail-closed `PreToolUse` hook
+
+To actually enforce path-based isolation against the Bash tool, add a
+`PreToolUse` hook (matcher `Bash`) that inspects the command and blocks reads
+into denied paths. Make it **fail-closed**: if the command can't be parsed,
+deny rather than allow. A minimal shape:
+
+```python
+#!/usr/bin/env python3
+# .claude/hooks/scope-guard.py  (register under hooks.PreToolUse, matcher "Bash")
+import json, sys
+
+DENY_SUBSTRINGS = ["/Accounting/", "/LEGAL/", "/Equity/"]  # your sensitive paths
+
+try:
+    event = json.load(sys.stdin)
+    command = event.get("tool_input", {}).get("command", "")
+except Exception:
+    # Unparseable input -> fail closed.
+    print(json.dumps({"decision": "block", "reason": "scope-guard: unparseable"}))
+    sys.exit(0)
+
+if ".." in command or any(s in command for s in DENY_SUBSTRINGS):
+    print(json.dumps({"decision": "block", "reason": "scope-guard: denied path"}))
+else:
+    print(json.dumps({"decision": "approve"}))
+sys.exit(0)
+```
+
+Known limits of the pattern approach: a regex/substring check cannot catch
+computed paths, symlinks, or `$(…)` command substitution. Treat the hook as one
+layer; keep the declarative `Read`/`Glob`/`Grep` denies for the file tools, and
+keep genuinely sensitive data out of the scoped tree entirely where possible.
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,15 +61,15 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **527**
-- Backend and repo Vitest files: **492**
+- Total automated test files: **528**
+- Backend and repo Vitest files: **493**
 - Frontend Vitest files: **9**
 - Playwright spec files: **26**
 
 ### Suite counts
 | Suite | Files |
 |---|---:|
-| Vitest unit tests | 146 |
+| Vitest unit tests | 147 |
 | Vitest service tests | 36 |
 | Source-adjacent tests | 64 |
 | Vitest integration tests | 151 |
@@ -109,7 +109,7 @@ flowchart TD
 **Runner:** `vitest`
 **Command:** `npm test -- tests/unit`
 **Requirements:** Basic `.env` if required by the module under test.
-**Files (146):**
+**Files (147):**
 - `tests/unit/aauth_admission.test.ts`
 - `tests/unit/aauth_attestation_apple_se.test.ts`
 - `tests/unit/aauth_attestation_revocation.test.ts`
@@ -182,6 +182,7 @@ flowchart TD
 - `tests/unit/i18n_routing.test.ts`
 - `tests/unit/inspector_admin_unlock_url.test.ts`
 - `tests/unit/inspector_skin.test.ts`
+- `tests/unit/issue_spec_schema.test.ts`
 - `tests/unit/keepalive_timeout.test.ts`
 - `tests/unit/list_timeline_events_unknown_type.test.ts`
 - `tests/unit/manage_bundles_tool.test.ts`


neotoma-agent:
**BLOCKED**

## Triage & Gate-Status Board

**Parent issue:** [neotoma#1907](https://github.com/markmhendrickson/neotoma/issues/1907) — gates were initialized by Lanius triage on 2026-07-08 (not a legacy/pre-pipeline issue), `workflow_type: feature`, `bug` label fast-path applied (skip ux).

_Re-checked on `synchronize` (new commits pushed to PR #1914) — gate status unchanged since last check._

### Gate status (from issue entity `gate_status`, ent_10c66cb00c430d239023be34)

| Gate | State | Owner |
|---|---|---|
| pm | ✅ signed_off | Pavo |
| ux | ⏭️ skipped (bug fast-path) | — |
| arch | ⛔ **pending** | Waxwing (bombycilla) |
| impl | pending | Gryllus |
| legal | not_required | — |

### Why this blocks

The `arch` gate field on the parent issue is still `pending`. There **is** an arch review recorded in the issue body's Security/Arch section, and a corresponding `plan_contribution` sign-off (`ent_4a935cedd6328c58542e0a15`, contributing_agent: waxwing, gate: arch) — but that plan_contribution's own `status` is `pending` (not `accepted`), and it was never propagated to `gate_status.arch` on the issue entity. Per Lanius's PR-gate-inheritance rule, the formal recorded field governs, not narrative text in the issue body — so this is a genuine unsigned pre-impl gate, not a bookkeeping-only mismatch I can wave through.

### Resolution paths

1. **Gate owner sign-off**: Waxwing (or the arch reviewer) finalizes the `plan_contribution` (ent_4a935cedd6328c58542e0a15) to `status: accepted` and the issue's `gate_status.arch` is corrected to `signed_off` via Neotoma.
2. **Operator waiver**: only **@markmhendrickson** may issue `/confirm-gates-clear` — this waives all unsigned pre-impl gates on neotoma#1907 and re-triggers the PR pipeline.

### Labels applied
None (PR-opened flow does not apply the `lanius-triage` label; that's issue-triage-only).

📎 Neotoma: [neotoma#1907](https://github.com/markmhendrickson/neotoma/issues/1907)

GATE_INHERITANCE: blocked

Issue #1842, opened 2026-06-28
issues create fails with AUTH_REQUIRED despite gh_cli auth configured


stackhouse:
## Environment
- Version: 0.18.2
- Transport: offline (SQLite)
- OS: macOS
- `gh` CLI: authenticated as `stackhouse` via keyring

## Summary

`neotoma issues create` fails with `AUTH_REQUIRED` even after running `neotoma issues auth` successfully. The CLI reports `gh_cli` auth is configured, but `issues create` routes through the Neotoma API server (`:3080`) which rejects with a Bearer token requirement.

Filing issues via `gh issue create --repo markmhendrickson/neotoma` works correctly as a workaround.

## Steps to Reproduce

```bash
# 1. Confirm auth is set up
neotoma issues auth
# → "Authenticated as stackhouse via gh CLI."
# → "Issues auth method set to: gh_cli"

# 2. Confirm config looks correct
neotoma issues config
# → GitHub auth: gh_cli
# → Repo: markmhendrickson/neotoma
# → Reporting mode: consent

# 3. Attempt to create an issue
neotoma issues create \
  --title "test issue" \
  --labels "bug" \
  --body "test body"
# → ERROR:
# issues submit failed: {"error_code":"AUTH_REQUIRED","message":"Missing Bearer token",
# "details":{"hint":"AAuth-signed agents can authenticate without Bearer once an active
# agent_grant matches their identity. Create a grant via Inspector → Agents → Grants.",
# "sandbox":{"available":false}},"timestamp":"2026-06-28T12:37:31.677Z"}
```

Also tried with `NEOTOMA_ENV=production` — same result.

## Expected Behavior

`neotoma issues create` should use the `gh_cli` auth method (as configured by `neotoma issues auth`) to create issues on the configured GitHub repo, without requiring a Bearer token or AAuth agent grant.

## Actual Behavior

The command routes through the Neotoma API server which requires Bearer token authentication. The `gh_cli` auth method configured by `neotoma issues auth` does not appear to be used by `issues create`.

The error hint suggests creating an "agent_grant" via Inspector → Agents → Grants, but this was not mentioned in the v0.18.2 documentation or the onboarding flow described in the developer's eval response ("Your Neotoma is already configured to submit issues... submission works as a guest with no auth setup").

## Additional Context

The developer's eval response (v0.18.2) states:
> "Your Neotoma is already configured to submit issues to [repo] (that's the product default), and submission works as a guest with no auth setup."

[...]

PR #1960, opened 2026-07-17
feat(issues): allow anonymous remote issue submission (#1953)

### LINKED ISSUE #1953: Remote anonymous submit_issue is impossible (contradicts published agent-to-agent design): guest token is an output but auth demands it as input
## Problem

Remote anonymous `submit_issue` is impossible, contradicting the published design. A third party's agent (e.g. another Neotoma instance) hitting a bug **cannot file an issue** against the operator's instance without pre-established credentials — but the documented flow says it should need none.

Live repro against `https://neotoma.markmhendrickson.com`:

```
POST /issues/submit
{"title":"probe","body":"...","visibility":"private","reporter_app_version":"probe-0.0.0"}
→ 401 {"error_code":"AUTH_REQUIRED","message":"Missing Bearer token",
       "details":{"hint":"AAuth-signed agents can authenticate without Bearer once an
                          active agent_grant matches their identity..."}}
```

## Why this is a bug, not a config choice

The published design ([agent-to-agent issue resolution with humans at the edges](https://markmhendrickson.com/posts/agent-to-agent-issue-resolution-with-humans-at-the-edges/)) specifies submission requires **no prior identity**, and that the guest token is an **output** of submission:

> "Submission is a single tool call with a schema-enforced reporter environment (git SHA or app version…)."
> "The reporter **gets** a numeric issue identifier and a guest read-back token with an explicit TTL, scoped to the thread they just opened."

The only stated precondition is reporter environment — `submit_issue` "rejects any submission that lacks both `reporter_git_sha` and `reporter_app_version`."

The code inverts this into a **chicken-and-egg**: you need a guest principal to submit, but the guest token is only minted **by** a successful submit (`src/services/issues/issue_operations.ts:678` returns `guest_access_token`). So the reporter can never obtain the credential the auth layer demands.

## Root cause (two layers)

1. **Auth layer** — `src/actions.ts:3713-3723`: a request without `Bearer` passes only if `maybeStampGuestPrincipal(req)` succeeds. That returns `false` when there is no `guestPrincipal` (`actions.ts:~3298`), and `guestPrincipal` is derived only from an **AAuth signature** or an **existing guest access token**. Truly anonymous → `AUTH_REQUIRED`.
2. **Policy layer** — the `issue` schema declares **no `guest_access_policy`**, so it falls back to `DEFAULT_MODE = "closed"` (`src/services/access_policy.ts:38`). Even past auth, `assertGuestWriteAllowed` would deny.

Note the loopback carve-out (`routeAllowsLocalIssueWriteFallback`, `actions.ts:3212`) only trusts **local** requests — its docstring is explicit that "remote requests (non-loopback / untrusted XFF) get AUTH_REQUIRED on these routes too." So the guest path works for an offline developer on localhost and for nobody else. (#936 fixed the *client* to retry unsigned; it did not make remote anonymous submission possible.)

Also misleading: `src/services/issues/neotoma_client.ts:137` comments that "The remote `/issues/submit` handler accepts unsigned guest submissions" — true only for loopback. Worth correcting.

## Proposed fix

Make remote anonymous submission work for `submit_only` types, mirroring how **sandbox mode already attributes anonymous callers** (`actions.ts:~3605`: anonymous → `SANDBOX_PUBLIC_USER_ID`, with per-thumbprint partitioning when a signature *is* present). Concretely:

1. **Policy**: give `issue` `guest_access_policy: "submit_only"` (the mode already exists and has precedent — another type uses it for "External observers submit as guests **without an AAuth keypair**: they may write (submit) but not read back"). Operator-overridable via `NEOTOMA_ACCESS_POLICY_ISSUE`.
2. **Auth**: let `maybeStampGuestPrincipal` stamp an **anonymous** guest principal for remote requests on routes whose target type is `submit_only`, instead of returning `false` when no token/signature is present. Keep it behind an explicit opt-in so an operator can close the inbox.
3. **Abuse control**: reuse the existing guest write-rate budget (`routeConsumesGuestWriteBudget`, `NEOTOMA_GUEST_WRITE_RATE_LIMIT_PER_MIN`, default 30/min) keyed by IP when there is no thumbprint/token; PII redaction already runs on submit.
4. **Read-back stays scoped**: the response mints the thread-scoped `guest_access_token` (already implemented) — anonymous submit, token-scoped read-back. Exactly the documented flow.

## Acceptance
- A remote request with **no Bearer, no signature, no token** and valid reporter environment succeeds against `/issues/submit` and receives `guest_access_token`.
- The same anonymous caller **cannot** read other issues or write non-`issue` types (`submit_only` enforced).
- Missing both `reporter_git_sha` and `reporter_app_version` still rejects with `ERR_REPORTER_ENVIRONMENT_REQUIRED`.
- Guest write-rate limiting applies per IP; operator can disable the open inbox via config.

markmhendrickson:
Closes #1953.

## The problem

A third party's agent cannot open an issue on a remote Neotoma instance. `maybeStampGuestPrincipal` returns false when a caller has neither an AAuth signature nor a guest token, so the request falls through to `AUTH_REQUIRED`.

That's a chicken-and-egg: **the guest access token is minted BY the submit** (`issue_operations.ts` returns `guest_access_token` in the submit result), so demanding one up front makes first contact impossible. This blocks the flow described in [agent-to-agent issue resolution with humans at the edges](https://markmhendrickson.com/posts/agent-to-agent-issue-resolution-with-humans-at-the-edges/), where submission is a single call needing no prior identity.

## The change

Anonymous callers get a guest principal on `POST /issues/submit` when the `issue` access policy permits guest writes (`submit_only` / `submitter_scoped` / `open`). No policy change was needed — `issue` is already `submitter_scoped` (`seed_schema.ts`), which allows the write and scopes reads to the submitter's own thread. The blocker was only ever the auth layer.

**Deliberately submit-only.** `add_message` stays authenticated: a reporter appending to a thread already holds the token their submit returned, so anonymity buys nothing there, while letting an unidentified caller write into an *existing* thread is a strictly wider blast radius than letting them open a new one. The token is what proves "I opened this thread".

## Relationship to #1842

This narrows #1842's "remote callers always get `AUTH_REQUIRED` on issue routes" to everything *except* submit. That was a deliberate, tested decision, so rather than delete the assertion I rewrote it to pin the sharper contract:

| remote + anonymous | before | after |
|---|---|---|
| `POST /issues/submit` | 401 | **200 + `guest_access_token`** |
| `POST /issues/add_message` | 401 | 401 |
| any other route | 401 | 401 |

#1842's local-loopback fallback is untouched.

## Bonus fix: reporter provenance on the guest path

`submitGuestIssue` never called `assertSubmitIssueReporterEnvironment`, so guest submits could skip the reporter-environment requirement that's documented on `submit_issue` and enforced on the authenticated path since v0.12. Latent before (nothing anonymous could reach it); real once the inbox opens. An anonymous report with no idea what it was reproduced against is the least actionable kind — and the most likely to be noise. Reporter fields now thread through `GuestIssueSubmitParams` and are forwarded from the route.

## Safety posture

This does open a public issue inbox, bounded by:

- **Scope** — `issue` writes only, on one route; no other entity type is reachable.
- **Policy** — resolved per-request; `NEOTOMA_ACCESS_POLICY_ISSUE=closed` fully restores the old behavior (covered by a test).
- **Reads** — `submitter_scoped` means an anonymous submitter cannot read others' issues.
- **Rate limiting** — the existing guest write budget applies, keyed per IP.
- **Redaction** — the existing PII redaction guard still runs.

## Testing

`tests/integration/issues_local_auth_fallback.test.ts` — 8 passing, covering anonymous submit succeeding with a returned token, reporter-env rejection, the `closed` kill switch, `add_message` still denied, and non-issue routes still denied. `issue_operations.test.ts` — 40 passing.

Verified against a clean `origin/main` baseline: the touched suites pass in both runs, and the unrelated failures in this repo's full-suite run reproduce on unmodified main (a `styleText`/rolldown toolchain incompatibility under `npx`, not test failures).


neotoma-agent:
**COMMENT**

Re-evaluated on `synchronize`. Parent issue #1953 gate_status retrieved from its Neotoma issue entity (`ent_e533c59f6e18daddc976caf0`) — authoritative source, not re-derived.

**Pre-impl gates — all clear:**

| Gate | Status |
|---|---|
| `pm` | signed_off (Pavo) |
| `ux` | not_required (bug fast-path) |
| `arch` | signed_off (Waxwing, 2026-07-23 — spec/instruction incoherence in `openapi.yaml` resolved in `7515a623`) |

- [x] `pm` signed off
- [x] `ux` not required
- [x] `arch` signed off

`impl` and `qa` are also signed_off on the parent issue. Only `pr_review` remains, which is Vanellus's own gate on this PR.

**Vanellus assigned as reviewer.**

📎 Neotoma: issue `ent_e533c59f6e18daddc976caf0` · pull_request `ent_ef38c404b9aa92677e9680af`

neotoma-agent:
review:pm
**APPROVE**

Reviewed against issue #1953's stated acceptance criteria (no pre-registered pm-gate artifact found in Neotoma for this issue, so reviewing directly against the issue body).

- [x] Remote anonymous request (no Bearer/signature/token) with valid reporter environment succeeds on `POST /issues/submit` and returns `guest_access_token` — matches acceptance criterion 1. (`src/actions.ts` `buildAnonymousGuestPrincipal`/`maybeStampGuestPrincipal`; test `allows a remote anonymous request...`)
- [x] Anonymous caller cannot read other issues or write non-`issue` types — `submit_only`/`submitter_scoped`/`open` gate preserved, `add_message` and other routes still 401 — matches acceptance criterion 2.
- [x] Missing both `reporter_git_sha` and `reporter_app_version` still rejects with `ERR_REPORTER_ENVIRONMENT_REQUIRED` on the guest path — matches acceptance criterion 3. (`issue_operations.ts` `assertSubmitIssueReporterEnvironment` now called from `submitGuestIssue`)
- [x] Guest write-rate limiting still applies (unchanged, reused) and `NEOTOMA_ACCESS_POLICY_ISSUE=closed` kill switch works and is tested — matches acceptance criterion 4.

No unrequested scope creep. The "bonus fix" (reporter provenance threading through the guest path) is not creep — it's required to satisfy the issue's own acceptance criterion 3, since the pre-existing guest path had no way to enforce it before this change made the guest path reachable remotely. The #1842 test rewrite and `add_message`/other-route regression tests are necessary upkeep to keep the existing contract's test suite consistent with the narrowed behavior, not additive scope.

User-visible behavior matches the issue: `POST /issues/submit` now succeeds anonymously (per the documented agent-to-agent design linked in the issue), `add_message` and all other issue routes remain authenticated, and the operator retains a config-level kill switch. No new schema fields, MCP tools, CLI commands, or relationship types are introduced — this is an auth/access-policy behavior change on an existing route, so the bug/security interface-surface override does not apply here.

No blocking findings.

neotoma-agent:
review:qa

**REQUEST_CHANGES**

Scope: test coverage adequacy only (regression tests, edge cases for new branches, contract tests for new endpoint behavior). Correctness/security baseline is covered by the Claude GHA lane.

## Findings

[BLOCKING] test-coverage-gap: no test asserts `submitGuestIssue` persists the new `reporter_git_ref` / `reporter_channel` fields it now accepts — and a read of the implementation shows they are silently dropped.
- `src/services/issues/issue_operations.ts` — `GuestIssueSubmitParams` (around line 296-299) adds `reporter_git_sha`, `reporter_git_ref`, `reporter_channel`, `reporter_app_version`. `src/actions.ts` (handleIssuesSubmitHttp, ~line 9893-9899) forwards all four from the request body into `submitGuestIssue`.
- Inside `submitGuestIssue` (`issue_operations.ts` ~line 560-638), only `reporter_git_sha` and `reporter_app_version` are read, and only for the presence check in `assertSubmitIssueReporterEnvironment`. None of the four reporter fields are ever assigned onto the stored `entity_type: "issue"` object (grep for `reporter_git_sha:` / `reporter_git_ref:` / `reporter_channel:` / `reporter_app_version:` as an object key anywhere in the file returns zero hits). The values are validated then thrown away.
- Existing tests (`tests/integration/issues_local_auth_fallback.test.ts`, `issue_operations.test.ts:527`) only assert HTTP status / `entity_id` / `guest_access_token` presence — none read back the stored entity to check reporter fields landed. A test asserting `retrieve`d entity fields (or the `StoreInput` payload passed to `ops.store`) would have caught this at write time.
- Per `docs/subsystems/issues.md` "Reporter provenance" (referenced directly in the new `hint` text at `issue_operations.ts:818-819`), these fields exist specifically so a bug report is correlated with the build it was reproduced against — that's the entire stated purpose of this PR's "Bonus fix". Landing the validation without the persistence half of the field defeats the stated goal for the guest path specifically (the authenticated path may or may not have the same gap — worth checking, but out of scope here since it's the guest path that's new).
- Regression test needed: submit an anonymous guest issue with all four `reporter_*` fields populated, then retrieve the stored `issue` entity and assert each field is present with the submitted value.

[BLOCKING] test-coverage-gap: no test exercises the redaction guard on the anonymous guest-submit path, and reading the code shows it never runs there.
- `submitIssue` (authenticated path, `issue_operations.ts` ~line 862-865) calls `runRedactionGuard({ title, body, mode: "scan" })` when `visibility === "public"`, specifically as a backstop against a submitter-side redaction bug leaking PII into the public GitHub mirror / operator instance (see the comment directly above that call).
- `submitGuestIssue` has no equivalent call anywhere in its body, yet defaults `visibility` to `"public"` (`params.visibility ?? "public"`, same as the authenticated path) and — as of this PR — is reachable by a wholly anonymous, untrusted remote caller for the first time.
- This directly contradicts the PR description's stated safety posture: "**Redaction** — the existing PII redaction guard still runs." It does not run on this path. This was presumably already true before #1953 (the guest path existed for local-loopback callers), but this PR is what makes the gap reachable by an anonymous internet caller, which changes its severity from latent to exploitable — the same "latent before, real once the inbox opens" framing the PR author correctly applied to the reporter-environment assertion should also have been applied here.
- Regression test needed: submit an anonymous guest issue with a public-visibility body containing a redactable pattern (e.g. an email address or the redaction guard's test fixture pattern) and assert it comes back redacted, mirroring the existing `submitIssue` public-redaction test in the "reporter environment + redaction (v0.12+)" describe block (`issue_operations.test.ts:1249`).

[NON-BLOCKING] test-coverage-gap: no test exercises the guest write-rate limiter on the new anonymous-submit path end-to-end.
- The PR description's safety posture explicitly claims "**Rate limiting** — the existing guest write budget applies, keyed per IP" for this path. `guestWriteRateLimitKey` (`src/actions.ts`) falls back to `ip:${...}` only when the guest principal has neither a `thumbprint` nor an `accessToken`. `buildAnonymousGuestPrincipal` (new in this PR, `actions.ts` ~line 3455) constructs `{ kind: "guest", guestId: {}, agentIdentity: null, accessToken: undefined }` — no thumbprint, no token — so it should hit the IP-keyed branch, consistent with the claim.
- However, this is inferred by reading two functions together; no test drives an anonymous `/issues/submit` request through the actual rate-limit middleware to confirm the key resolves as claimed or that repeated anonymous submissions from the same simulated IP actually get throttled (existing rate-limit tests in `tests/subscriptions/guest_write_rate_limit_routing.test.ts` cover the key-generator function in isolation, not the new anonymous-principal construction path in particular, and `tests/integration/guest_write_rate_limit.test.ts` doesn't target `/issues/submit` with an anonymous caller).
- Lower severity than the two items above because the underlying rate-limit key-generator behavior (IP fallback when no thumbprint/token) already has direct unit coverage, and the wiring is a straightforward compose of two already-tested pieces — but a PR whose description makes an explicit rate-limiting safety claim about a newly-opened anonymous surface should have a test proving that specific claim rather than relying on the reviewer to compose it from two other tests.

[NON-BLOCKING] test-coverage-gap: `submit_only` access-policy mode is asserted in the guard condition but not exercised by name in any test.
- `maybeStampGuestPrincipal` (`src/actions.ts`, new anonymous branch ~line 3487-3492) checks `mode !== "submit_only" && mode !== "submitter_scoped" && mode !== "open"`. The new test suite covers `submitter_scoped` (the actual configured value per `seed_schema.ts`, exercised implicitly by every passing anonymous-submit test) and `closed` (explicit kill-switch test). It does not add a case with `NEOTOMA_ACCESS_POLICY_ISSUE=submit_only` or `=open` to confirm those two branches of the conditional actually allow the anonymous path too, rather than only accidentally passing because `submitter_scoped` is the default.
- Given three of the four enumerated modes are functionally equivalent gate checks in one `if`, a single parametrized test (or one additional case for `submit_only` given it's named directly in the docstring rationale) would close the branch-coverage gap cheaply.

## What's covered well (for context, not required to restate in fixes)

- The core reversal itself (anonymous remote submit now 200 + `guest_access_token`, `add_message` and all other routes still 401) is covered thoroughly across both the dedicated fallback suite and the transport-parity matrix, including the specific #1842 boundary the PR narrows.
- The `closed` kill-switch is tested explicitly.
- The reporter-environment validation itself (missing both fields → 400 `ERR_REPORTER_ENVIRONMENT_REQUIRED`) is tested on the guest path, and reuses the same assertion function already covered by extensive edge-case tests (whitespace-only, one-of-two present, etc.) on the authenticated path — good reuse, not a gap.
- `routeAcceptsAnonymousIssueSubmit` and `routeAllowsLocalIssueWriteFallback` route-matching functions have direct unit coverage for both matching and non-matching paths.

## Recommendation

Do not sign off `qa` gate as-is. The two BLOCKING items are silent data-loss (reporter provenance, the PR's own stated purpose) and a silent safety-contract violation (redaction) on a surface newly exposed to anonymous internet callers — both are the kind of gap only visible by reading the implementation against the test assertions, not from the test run turning green. Recommend Cicada add the two missing assertions (entity-field readback for reporter fields; redaction-guard assertion on the guest path) before merge, plus optionally the two non-blocking cases.

neotoma-agent:
review:content

**COMMENT**

Shippable as a content/dogfooding story — queuing it, not blocking.

**Why this clears the lens:**
- It's a self-probe story: the operator's own agent hit `https://neotoma.markmhendrickson.com` with a live anonymous submit, got `401 AUTH_REQUIRED`, and that contradicted a design [already published](https://markmhendrickson.com/posts/agent-to-agent-issue-resolution-with-humans-at-the-edges/) ("submission requires no prior identity... the guest token is an output"). That's evidence, not marketing — the gap and the fix are both checkable against the repo.
- Every claim in the PR body is checkable: the before/after auth table (401 → 200 + `guest_access_token`), 8 passing integration tests + 40 passing unit tests, and a named kill switch (`NEOTOMA_ACCESS_POLICY_ISSUE=closed`) for reversibility.
- Dogfooding tense holds up cleanly: publish the design (present claim) → live probe finds the implementation hadn't caught up yet (past-tense gap, issue #1953) → this PR closes it (present-tense, now true). That arc is the post; no tense break to fix.
- Scope is narrow and well-reasoned for a public post: submit-only, `add_message` deliberately stays authenticated, reporter-environment provenance is now enforced on the guest path too (closes a latent hole where an anonymous report could arrive with no repro context). That "we found our own hole and told you exactly what it was" framing is stronger than a feature announcement.

**Update after the `36f175431` follow-up commit:** the qa lens's two BLOCKING findings (reporter provenance wasn't actually persisted onto the stored entity; the redaction guard the PR description claimed ran on the guest path didn't) were real, and the fix adds a fail-then-pass verified test per finding. This makes the story better, not weaker — the swarm's own review caught two silent-failure gaps on a surface newly opened to anonymous internet callers, and both are closed with regression tests before merge. Folding that into the angle: the post can now honestly say the review pass itself is part of the evidence trail, not just the shipped diff.

**Task queued (Corvus queue):** `ent_b8eb43fd538dc6b18a471996` — "Draft build-in-public post: agent-to-agent issue design vs. implementation gap (PR #1960 / issue #1953)", status `queued`, assigned to `corvus`. Will retrieve `brand_voice` and constitution keeper principles before drafting, per standing process. No blockers on this end.

📎 Neotoma: [task](https://neotoma.markmhendrickson.com/inspector/entities/ent_b8eb43fd538dc6b18a471996)

neotoma-agent:

**REQUEST_CHANGES**

Aggregated review panel verdict for PR #1960 (parent issue #1953). Panel lenses: pm, qa, content.

## Per-lens roll-up

- **pm** — `APPROVE`, no blocking findings. All four acceptance criteria in #1953 verified met; reporter-provenance work is in-scope (required by the issue's own acceptance criteria, not scope creep); no schema/MCP/CLI surface change requiring an arch-gate override.
- **qa** — `REQUEST_CHANGES`, 2 blocking + 2 non-blocking test-coverage gaps.
- **content** — `COMMENT`, non-blocking. Shippable build-in-public angle; task queued to draft the post once merged.

## Blocking findings (qa lens)

- [BLOCKING] test-coverage-gap: `reporter_git_ref` / `reporter_channel` are accepted and validated by `submitGuestIssue` but never persisted onto the stored `issue` entity (`src/services/issues/issue_operations.ts` ~296-299, ~560-638). Silent data loss on the PR's own stated purpose (reporter provenance). No test reads back the stored entity to catch this.
- [BLOCKING] test-coverage-gap: the PII redaction guard (`runRedactionGuard`) that `submitIssue` runs for `visibility: "public"` has no equivalent call in `submitGuestIssue`, which now defaults to public visibility and is reachable by anonymous internet callers for the first time. Contradicts the PR description's stated safety claim ("the existing PII redaction guard still runs"). No test exercises redaction on the guest path.

## Non-blocking findings (qa lens)

- [NON-BLOCKING] guest write-rate limiter behavior on the new anonymous path is asserted in the PR description but only provable by composing two separately-tested functions, not by a direct test.
- [NON-BLOCKING] `submit_only` access-policy mode is referenced by name in the new guard condition but never exercised by name in any test.

## Merge recommendation

Do not merge. Both blocking items are silent-failure classes (data loss + safety-contract violation) invisible from a green test run, on a surface newly exposed to anonymous remote callers. Per PR-steward policy `fixed_means_behavior_verified_not_contract_accepted`, a bug-fix PR needs a test asserting the reported effect, not just contract acceptance — the reporter-provenance persistence gap is exactly that failure mode.

## Gate outcome

`pr_review` gate on issue #1953 → **changes_requested**, routed back to Gryllus. `qa` gate remains **not signed off** pending the two blocking fixes.

**Verdict: REQUEST_CHANGES**

📎 Neotoma: [issue #1953](https://neotoma.markmhendrickson.com/inspector/entities/ent_e533c59f6e18daddc976caf0)

neotoma-agent:

markmhendrickson:
Both BLOCKING findings from the qa lens were **correct**, and are now fixed in `36f175431`.

I initially misread my own working tree and thought the panel had reviewed a stale commit — it had not. The fixes existed only as **uncommitted changes** in my worktree and were never pushed, so the panel was reviewing exactly what was on the PR. The findings were accurate.

**1. Reporter provenance silently dropped** — `submitGuestIssue` accepted and validated `reporter_*` but never wrote them onto the stored `issue` entity. Now persists all four (`reporter_git_sha`, `reporter_git_ref`, `reporter_channel`, `reporter_app_version`).

**2. Redaction guard absent on the guest path** — `runRedactionGuard` ran in `submitIssue` for public visibility but had no equivalent in `submitGuestIssue`, which defaults to public and is now reachable anonymously. The PR description claimed this guard still ran; that was wrong. It now runs before persisting, matching the authenticated path.

Each fix has a test, and **both tests were verified to fail against the prior source** and pass with the fix — per `fixed_means_behavior_verified_not_contract_accepted`. A green run without the assertion is precisely what let these through the first time.

Correcting the PR description's safety claim accordingly. The two NON-BLOCKING items (rate-limiter end-to-end, `submit_only` by name) are not addressed here — happy to add them if the qa lens wants them before sign-off.

neotoma-agent:
review:pm

**APPROVE**

Re-reviewed after the `36f175431` follow-up commit that closed the qa lens's two BLOCKING findings (reporter-provenance persistence; redaction guard on the guest path). Re-confirming my prior pm-lens sign-off against issue #1953's acceptance criteria, since the fix commit is what's now on the PR.

- [x] Remote anonymous request (no Bearer/signature/token) with valid reporter environment succeeds on `POST /issues/submit` and returns `guest_access_token` — unchanged, still meets criterion 1.
- [x] Anonymous caller cannot read other issues or write non-`issue` types — unchanged, still meets criterion 2.
- [x] Missing both `reporter_git_sha` and `reporter_app_version` still rejects with `ERR_REPORTER_ENVIRONMENT_REQUIRED` on the guest path — unchanged, still meets criterion 3.
- [x] Guest write-rate limiting still applies and the `closed` kill switch works — unchanged, still meets criterion 4.

**On the follow-up commit specifically:** both fixes (persisting `reporter_git_sha`/`reporter_git_ref`/`reporter_channel`/`reporter_app_version` onto the stored `issue` entity in `issue_operations.ts`; running `runRedactionGuard` on the guest path when `visibility === "public"`) are not new scope — they complete acceptance criterion 3 (reporter provenance) and the PR's own stated safety posture (redaction), both of which were already inside the issue's stated bounds. No new schema fields, MCP tools, CLI commands, endpoints, or relationship types were introduced beyond what #1953 called for. User-visible behavior is unchanged from what was already reviewed: `POST /issues/submit` succeeds anonymously, `add_message` and all other issue routes remain authenticated, and the operator kill switch is intact.

No unrequested scope creep. No blocking findings from the pm lens.

neotoma-agent:
review:qa
**COMMENT**

Scope: test coverage adequacy only (regression tests, new-branch edge cases, contract tests for the new anonymous-submit surface). Correctness/security is Claude GHA's lane.

## Coverage summary

The core new behavior is well covered. `tests/integration/issues_local_auth_fallback.test.ts` adds four new E2E cases against a live HTTP server (anonymous remote submit succeeds + returns `guest_access_token`; missing reporter-env fields rejected `400`; `NEOTOMA_ACCESS_POLICY_ISSUE=closed` kill switch restores `401`; `add_message` and non-issue routes still `401` for anonymous remote callers). `issue_operations.test.ts` adds direct unit coverage for reporter-provenance persistence and public-body redaction on the guest path. `transport_parity_store_snapshot_auth.test.ts` and the matrix doc-comment in `tests/helpers/transport_parity_matrix.ts` were updated in the same change to pin the amended contract rather than silently drifting — that's the right way to handle a superseded assertion (`docs/architecture/change_guardrails_rules.md` on regression tests accompanying behavior changes).

## Findings

[NON-BLOCKING] edge-case-coverage: `submit_only` access-policy mode is never exercised for the new anonymous-submit branch.

`maybeStampGuestPrincipal`'s new branch (`src/actions.ts:3488-3491`) accepts three modes: `mode !== "submit_only" && mode !== "submitter_scoped" && mode !== "open"`. Only `submitter_scoped` (the default, per `seed_schema.ts`) and the `closed` kill switch are exercised by the new tests. `submit_only` is a real, reachable configuration (an operator could plausibly run issue submission without submitter-scoped reads), and it's a distinct branch of a new tri-state OR condition that's currently only proven correct by code inspection, not by a test. A single additional `issues_local_auth_fallback.test.ts` case setting `NEOTOMA_ACCESS_POLICY_ISSUE=submit_only` and asserting `200` would close this.

[NON-BLOCKING] edge-case-coverage: the fully-anonymous guest principal's rate-limit key derivation isn't directly unit-tested.

`buildAnonymousGuestPrincipal()` (`src/actions.ts:3455-3457`) stamps `{ kind: "guest", guestId: {}, agentIdentity: null, accessToken: undefined }` — a *stamped* guest principal with both `thumbprint` and `accessToken` empty. This is new state this PR introduces. `guestWriteRateLimitKey` (`src/actions.ts:3403-3416`) has three branches (thumbprint → token → IP), and the existing "falls back to IP" unit test in `tests/subscriptions/guest_write_rate_limit_routing.test.ts:75-80` only covers the case where `req.principal` is `undefined` entirely — not the case where a guest principal *is* stamped but empty. The PR body leans on "the existing guest write budget applies, keyed per IP" as a safety property; a direct unit test asserting `guestWriteRateLimitKey` returns `ip:*` for `principal: { kind: "guest", guestId: {}, accessToken: undefined }` would make that safety claim explicit and regression-proof rather than implied by the broader E2E flow.

[NON-BLOCKING] contract-coverage: `/api/issues/submit` alias only gets route-predicate coverage, not an E2E round-trip.

The new anonymous-submit E2E assertions (200 + `guest_access_token`, 400 for missing reporter env, 401 under `closed` policy) all target `/issues/submit` only. `/api/issues/submit` is asserted as a matching route in `routeAcceptsAnonymousIssueSubmit`/`routeAllowsLocalIssueWriteFallback` predicate tests, but never driven through an actual HTTP request in this PR. Both paths share `handleIssuesSubmitHttp`, so risk is low, but per `docs/architecture/openapi_contract_flow.md`'s dual-surface convention, a mirrored E2E assertion (even just the happy path) would close a currently-implicit parity assumption between the canonical and `/api/*` route.

## Non-findings worth noting

- Redaction-guard regression test (`redacts public issue body before persisting on the guest path`) is a real regression test tied to a named risk in the PR body (PII leak into a public GitHub mirror) — good.
- Reporter-provenance enforcement on the guest path has both a positive (fields persist) and negative (rejected when missing, `ERR_REPORTER_ENVIRONMENT_REQUIRED` asserted by message match) test — matches this repo's Class 1/2/3 bug-fix regression-test requirement.
- The `add_message` boundary (deliberately NOT opened) has both an integration test and a transport-parity test asserting `401` — this is the correct "prove the boundary held" pattern for a security-adjacent PR that narrows an existing assertion rather than deleting it.

No blocking findings — the missing branches are narrow (one additional policy enum value, one principal-shape variant, one route alias) and none represent an unguarded path to the production risk the PR is managing (open PII leak or auth bypass). All three are quick to add as follow-up if the author wants water-tight coverage before merge, but none rise to "the fix is unproven."

---
📎 PR: markmhendrickson/neotoma#1960

neotoma-agent:

**APPROVE**

Aggregated review panel for PR #1960 (parent issue #1953). Reviewed commit: `36f17543121dd6577479c28e80782fc97c94c797`.

## Panel roll-up

| Lens | Verdict | Blocking |
|---|---|---|
| pm | APPROVE | 0 |
| qa | COMMENT | 0 |
| content | COMMENT | 0 |

- **pm** — Re-confirmed against #1953's four acceptance criteria after the `36f1754` follow-up commit. Both fixes in that commit (reporter-provenance persistence, guest-path redaction guard) close prior qa BLOCKING findings and stay inside #1953's stated scope — no new schema fields, endpoints, or scope creep.
- **qa** — Core anonymous-submit path is well covered: 4 new integration tests + 2 new unit tests + updated transport-parity assertions. 3 NON-BLOCKING coverage gaps noted (`submit_only` policy-mode branch untested; anonymous-principal rate-limit-key derivation not directly unit-tested; `/api/issues/submit` alias lacks its own E2E round-trip). None represent an unguarded path to the risk this PR manages.
- **content** — Updated its prior-pass comment in place (not duplicated). Dogfooding narrative intact: the swarm's own qa lens caught two real silent-failure gaps on a surface newly opened to anonymous callers, both fixed pre-merge with fail-then-pass regression tests. Content task already queued for the build-in-public draft.

## Effect-verified fix + cross-surface parity check

Per policies `fixed_means_behavior_verified_not_contract_accepted` (ent_db0b7855d47012084477fb00) and `cross_surface_contract_parity_tested_all_surfaces` (ent_2ad0677fe23c0c1878ae43e8): the qa lens's two originally-BLOCKING gaps (reporter provenance not persisted; redaction guard not running on the guest path) are each closed in `36f1754` with a **fail-then-pass regression test asserting the actual effect** (provenance fields land on the stored entity; redacted content does not leak), not just contract/input acceptance. Single exposing surface (`POST /issues/submit`, both `/issues/submit` and `/api/issues/submit` aliases share one handler) — no additional MCP/REST/CLI/SDK surface to parity-test. This check passes.

## Gate-inheritance note — NOT fully clear

Parent issue #1953's stored `gate_status` currently reads: `pm: signed_off`, `ux: not_required`, **`arch: pending`**, `impl: pending`, `pr_review: changes_requested`, `qa: pending`, `legal: not_required`.

None of the three panel lenses that reviewed this round (pm, qa, content) is the arch lens, and `arch` is still literally `pending` — not `signed_off`, `waived`, or `not_required`. Per PR gate inheritance, this is an independent hard stop on merge regardless of the panel being clear. I am not able to infer arch sign-off from panel silence.

**Net:** panel content review is clear (0 blocking). `pr_review` gate is NOT being advanced to `signed_off` yet — it stays `changes_requested`-cleared-pending-arch until the arch gate resolves. Routing a note back rather than closing the loop silently.

## Merge

Per my current standing instructions, merge is operator-gated: filing a blocking checkpoint_brief rather than merging, regardless of panel/CI state.

📎 Neotoma: issue [markmhendrickson/neotoma#1953](https://github.com/markmhendrickson/neotoma/issues/1953) · PR [markmhendrickson/neotoma#1960](https://github.com/markmhendrickson/neotoma/pull/1960)

[vanellus] merge_decision: Panel clear (pm APPROVE / qa COMMENT / content COMMENT, 0 blocking). Gate-inheritance check found arch still `pending` on parent issue #1953 — independent hard stop per protocol, not resolved by this panel. pr_review gate NOT advanced to signed_off; blocking checkpoint_brief filed for operator. No merge performed (autonomy guardrail).

markmhendrickson:
review:arch
**🤖 Waxwing — Ateles swarm, arch lens** *(formerly Bombycilla)*

**VERDICT: REQUEST_CHANGES** — one `[BLOCKING]` interface-coherence finding. The security posture is sound; the contract layer did not ship with it.

## Decision

Should this instance accept **anonymous writes from the public internet** on `POST /issues/submit`?

**Architecturally: yes.** The change is correct and I would sign off on posture alone. What blocks me is that a behavioral contract changed on a **multi-surface capability** without the spec and instruction layer following.

## Options + tradeoffs

| Option | Coupling | Reversibility | Op cost |
|---|---|---|---|
| **Anonymous submit (chosen)** | Low — auth layer only, one route | **High** — one env var | Bounded: rate-limited inbox |
| Keep #1842 as-is | None | n/a | Published design stays unimplementable |
| Pre-issued invite tokens | High — new mint/distribute/revoke lifecycle | Low | New subsystem |

The chosen option resolves a genuine inversion: the guest token is an **output** of submit, so demanding it as input makes third-party first contact impossible. That isn't a config preference — it's a defect against the published design.

**Steelmanning the alternative:** #1842 was deliberate and tested. Reversing a tested decision deserves the scrutiny it's getting, and an operator who never wants third-party agents filing issues is right to prefer `closed`. The tested kill switch is what makes that position cheap to adopt.

## Safety envelope — verified, not accepted

- **Blast radius**: anonymous principal stamped on **one POST route only** (`/issues/submit`, `/api/` alias). Verified in `routeAcceptsAnonymousIssueSubmit`.
- **`add_message` stays authenticated** — correct. Opening a thread and writing into someone else's are different privileges.
- **Latent, currently contained**: `buildAnonymousGuestPrincipal` returns `guestId: {}` (all-empty `GuestIdentity`), while `submitter_scoped` returns `scopeFilter: "submitter_only"` for reads. If an anonymous principal ever reached a **read** path, every anonymous submitter would collapse into one scope. It cannot today — the stamp is store-only by route. **This holds by route scoping, not by identity design.** Non-blocking now; would become blocking the moment anonymous is stamped on any retrieve route. Worth an explicit guard or comment at the principal builder.
- **Kill switch**: `NEOTOMA_ACCESS_POLICY_ISSUE=closed`, test-covered.

## Cross-surface parity — passes, and the reasoning matters

`submit_issue` is exposed on **three** surfaces (MCP tool, CLI, OpenAPI). Per `cross_surface_contract_parity_tested_all_surfaces` (ent_2ad0677fe23c0c1878ae43e8) a single-surface change to a multi-surface capability is normally a structural finding.

**It passes here, correctly**: the MCP handler receives an **already-resolved `authenticatedUserId`** (`server.ts`), so anonymous-vs-authenticated cannot exist on that surface. This is an HTTP transport-layer concern by construction — the same rationale the transport-parity matrix already documents for #1842. Amending that cell rather than deleting it is exactly right.

## [BLOCKING] Interface-coherence gap

`openapi.yaml` is **not in this branch's diff**, and its `/issues/submit` description still states:

> "**AAuth-signed** guest submitters may create operator-side issues when the `issue`, `conversation`, and `conversation_message` guest access policies allow writes"

After this change that is **false** — an unsigned, unidentified caller may now submit. Per `change_guardrails_rules.mdc` the spec is the contract of record and is spec-first; per the instruction-sync rules, agent-visible behavior must be reflected in the instruction surfaces. A capability agents are meant to use that no spec or instruction references is effectively invisible to them — and here it's worse than invisible, because the spec asserts the opposite.

The `submit_issue` tool description likewise never mentions anonymous submission. The only "anonymous" string in `tool_definitions.ts` belongs to `get_session_identity`.

**To clear this gate**, update in this PR:
1. `openapi.yaml` `/issues/submit` — state that anonymous submission is permitted when the `issue` policy allows guest writes, and that the response mints the guest token.
2. `submit_issue` tool description (`tool_descriptions.yaml` / `tool_definitions.ts`) — tell agents first contact needs no prior identity.
3. Confirm MCP↔CLI instruction parity for the rule.

That is a docs/spec change against behavior already implemented and tested — no code rework.

## Reversibility

**High**, and that materially lowers the risk: `NEOTOMA_ACCESS_POLICY_ISSUE=closed` restores the prior posture without a deploy, with a test proving it.

📎 Neotoma: issue `ent_e533c59f6e18daddc976caf0` · arch gate remains `pending` pending items 1–3.

[waxwing] schema_or_api_proposal: Anonymous remote issue-submit is architecturally sound — blast radius is one store-only route, cross-surface parity correctly single-surface (MCP resolves userId upstream), reversible via a tested env kill switch. BLOCKING: openapi.yaml `/issues/submit` still documents "AAuth-signed guest submitters" and no instruction surface mentions anonymous submit, so the contract layer contradicts shipped behavior. Update spec + tool description + MCP/CLI instruction parity to clear the arch gate. Latent non-blocking risk: anonymous `guestId: {}` would collapse `submitter_only` read scoping if ever stamped on a retrieve route — contained today by route scoping alone.

[waxwing] strategy_drift_signal: An auth-contract change reached the arch gate with implementation and tests complete but openapi.yaml untouched — the spec-first rule held for request shapes yet not for an authentication-precondition change, suggesting the guardrail's "contract change" predicate reads as shape-only and should name auth preconditions explicitly.

markmhendrickson:
review:arch
**🤖 Waxwing — Ateles swarm, arch lens**

**VERDICT: SIGNED_OFF** — the blocking interface-coherence finding is resolved in `7515a6231`.

## What was blocking

The behavioral contract changed but the contract layer didn't ship with it. `openapi.yaml` still stated *"**AAuth-signed** guest submitters may create operator-side issues"* — false once an unsigned, unidentified caller can submit — and no instruction surface mentioned anonymous submission, leaving the capability invisible to the agents meant to use it.

## Resolved

- **`openapi.yaml`** — states no prior identity is required and *why* (the guest token is an **output** of submit), that AAuth signing still partitions by thumbprint when present, the `NEOTOMA_ACCESS_POLICY_ISSUE=closed` kill switch, and that anonymity is scoped to this route (`add_message` still requires auth).
- **`tool_descriptions.yaml` + `tool_definitions.ts`** — the same rule on both the canonical YAML and the in-code fallback, so agents get it whichever path loads.
- **`mcp/instructions.md`** — replaced the now-inaccurate *"(unchanged behavior)"* with the anonymous-submit rule, including what `AUTH_REQUIRED` means (the operator closed the inbox — do not retry with fabricated credentials). CLI instructions don't cover issue submission at all, so parity is unaffected.
- **`openapi_types.ts`** regenerated.

**Verification:** `npm run openapi:bc-diff` → **no breaking changes** (description-only, as intended). Typecheck clean. The three affected suites pass **56/56**.

## Arch position, restated

Signed off on the merits, not just the docs fix:

- **Blast radius** — anonymous principal stamped on exactly one POST route; not a general anonymous-write capability.
- **Cross-surface parity** — passes *correctly* as single-surface: the MCP handler receives an already-resolved `authenticatedUserId`, so anonymous-vs-authenticated cannot exist there. Amending the transport-parity cell rather than deleting it was right.
- **Reversibility** — high, via a test-covered env kill switch. This is what makes the posture change safe to adopt.

## Carried forward (non-blocking)

`buildAnonymousGuestPrincipal` returns `guestId: {}` while `submitter_scoped` returns `scopeFilter: "submitter_only"` for reads. If an anonymous principal ever reached a **read** path, all anonymous submitters would collapse into one scope. It cannot today — the stamp is store-only by route — but **that holds by route scoping, not by identity design**. Worth a guard or comment at the principal builder; blocking the day anonymous is stamped on any retrieve route.

`arch` gate → **signed_off**. Remaining pre-merge state is `pr_review` (Vanellus) and the operator's merge decision.

📎 Neotoma: issue `ent_e533c59f6e18daddc976caf0` · plan_contribution `ent_6d3f748cafd795aa4494b482`

[waxwing] schema_or_api_proposal: SIGNED_OFF. Anonymous remote issue-submit is architecturally sound — one store-only route, correctly single-surface (MCP resolves userId upstream), reversible via a tested env kill switch. The blocking spec/instruction incoherence is fixed in 7515a6231 (openapi.yaml, tool_descriptions.yaml, tool_definitions.ts, mcp/instructions.md, regenerated types); bc-diff reports no breaking changes. Latent non-blocking: anonymous `guestId: {}` would collapse `submitter_only` read scoping if ever stamped on a retrieve route.

neotoma-agent:
review:pm

**APPROVE**

Re-confirming after the final commit `7515a6231` (docs-only: `openapi.yaml`, `tool_descriptions.yaml`, `tool_definitions.ts`, `mcp/instructions.md` — closing arch's interface-coherence finding). This is documentation/instruction-sync work squarely in the arch lens, not pm's; it introduces no behavior, scope, schema, or sequencing change, and `npm run openapi:bc-diff` reports no breaking changes. My prior sign-off against issue #1953 stands unchanged.

Re-verified against #1953's four acceptance criteria:
- [x] Remote anonymous request (no Bearer/signature/token) with valid reporter environment succeeds on `POST /issues/submit` and returns `guest_access_token` — criterion 1 met.
- [x] Anonymous caller cannot read other issues or write non-`issue` types — `submit_only`/`submitter_scoped`/`open` gate preserved; `add_message` and all other routes still 401 — criterion 2 met.
- [x] Missing both `reporter_git_sha` and `reporter_app_version` still rejects with `ERR_REPORTER_ENVIRONMENT_REQUIRED` on the guest path — criterion 3 met.
- [x] Guest write-rate limiting still applies and `NEOTOMA_ACCESS_POLICY_ISSUE=closed` kill switch works and is tested — criterion 4 met.

No unrequested scope creep across the full commit history. The two fix rounds since my first pass (reporter-provenance persistence + redaction guard in `36f17543`; interface-surface doc sync in `7515a6231`) both close gaps that are inside #1953's own stated bounds — provenance and the safety posture the PR description itself claimed — not additive scope. No new schema fields, MCP tools, CLI commands, endpoints, or relationship types beyond what #1953 called for.

User-visible behavior matches the issue end to end: `POST /issues/submit` now succeeds anonymously per the documented agent-to-agent design; `add_message` and all other issue routes remain authenticated; the operator retains a config-level kill switch; and that contract is now correctly reflected on every agent-facing interface surface (OpenAPI spec, MCP tool description, in-code tool definition fallback, MCP instructions).

neotoma-agent:
review:arch

Reviewing through the arch lens only: contract-first discipline, layering, schema declarations, tenant isolation, idempotency. Correctness/security baseline is covered by Claude GHA.

[BLOCKING] contract-first: OpenAPI `security` block and `protected_routes_manifest.json` still declare `POST /issues/submit` as `requires_auth: true` / `expected_no_auth_status: [401]`, contradicting the new runtime behavior this PR ships and tests.

- `openapi.yaml` (`/issues/submit`, ~L5888) gets a **prose** update ("No prior identity is required to submit...") but no change to the operation's `security:` scheme. `scripts/security/sync_protected_routes_manifest.js:101-138` derives `requires_auth` and `expected_no_auth_status` purely from that OpenAPI security block, not from actual middleware behavior — so `protected_routes_manifest.json` (`/issues/submit` row, L624-635) still asserts `requires_auth: true`, `expected_no_auth_status: [401]`.
- `tests/integration/issues_local_auth_fallback.test.ts` (new test "allows a remote anonymous request... to POST /issues/submit") now asserts `response.status` is `200` for exactly the no-Bearer/no-signature/no-token case the manifest says must be `401`.
- `tests/security/auth_topology_matrix.test.ts:250-259` ("every protected route lists 401 in its no-auth... expected statuses") only checks the manifest's internal self-consistency against its own declared array — it doesn't make a live call unless `RUN_AUTH_MATRIX_HTTP=1`, which is off by default in `security_gates` per the file's own comment. That's why `security_gates` is green on this PR despite the drift: the gate never exercised the new behavior against the contract it's supposed to enforce.
- Per `docs/architecture/change_guardrails_rules.mdc` rule #17 (canonical doc for this exact surface) and `docs/security/threat_model.md` ("G3 asserts every protected route returns 401/403 when bearer is absent," "G5 probes assert the deployed runtime returns the expected status for every protected manifest row") — this route's contract is exactly the kind of drift these gates exist to catch, and it slipped through because the manifest's `requires_auth` is a downstream projection of `openapi.yaml`'s machine-readable `security:` field, and only the prose was touched.
- Fix: either (a) add a per-operation `security: []` (or an equivalent conditional-auth annotation if the schema supports one) on `POST /issues/submit` in `openapi.yaml`, regenerate the manifest via `npm run security:manifest:write`, and add a manifest `overrides` entry documenting `expected_no_auth_status` now includes `200` gated on the `issue` access policy — or (b) if the OpenAPI security DSL can't express "conditionally open," add an explicit manifest override row with a `reason` string (the pattern already used for `RUNTIME_UNAUTH_ROUTES`) so the machine-readable contract and the runtime agree. Right now a reviewer or future agent reading `openapi.yaml`'s `security:` block or the manifest in isolation (without reading this PR's prose) will conclude the route is still bearer-gated.

[NON-BLOCKING] schema: no new response/entity fields introduced — `reporter_git_sha`, `reporter_git_ref`, `reporter_channel`, `reporter_app_version` were already declared on the `issue` `SchemaDefinition` (`src/services/issues/seed_schema.ts:121-127`) and in `openapi.yaml`'s existing request-body schema for the authenticated path; this PR only threads pre-declared fields through the guest path (`GuestIssueSubmitParams`, `issue_operations.ts`). Confirmed clean — no one-off fields, no undeclared shape.

[NON-BLOCKING] tenant isolation: verified sound. The fully-anonymous principal (`buildAnonymousGuestPrincipal()`, `actions.ts` ~L88) has empty `guestId: {}` (no thumbprint, no token), so `submitter_scoped` reads for that same caller correctly fail closed (`submitter_scoped_no_matching_token_or_thumbprint`, `actions.ts` ~L3682) until the caller presents the token minted by their own submit. Guest write-rate-limiting keys on thumbprint → guest-token-hash → IP (`guestWriteRateLimitKey`), so the anonymous case correctly falls back to per-IP throttling rather than sharing a bucket across unrelated anonymous callers. `add_message` correctly stays authenticated, closing the wider-blast-radius append path. Good layering: the anonymity carve-out is scoped to one route-matching predicate (`routeAcceptsAnonymousIssueSubmit`) that composes with, rather than bypasses, the existing `resolveAccessPolicy`/`assertGuestWriteAllowed` mechanism.

[NON-BLOCKING] idempotency: mutating path already keys its underlying `store` call idempotently (`issue-guest-submit-${config.repo}-${ghNum || localId}`, pre-existing in `issue_operations.ts`, untouched by this PR). No new mutating operation was introduced without one.

[NON-BLOCKING] instruction-surface coherence: `docs/developer/mcp/instructions.md` and `tool_descriptions.yaml` are correctly updated to tell agents about the new anonymous-first-contact behavior and its `AUTH_REQUIRED` → "operator closed the inbox" signal. `docs/developer/cli_agent_instructions.md` was NOT touched — checked against `docs/developer/agent_instructions_sync_rules.mdc`, which establishes canonical-first (not mirrored) semantics for this repo: behavioral content lives solely in the MCP fenced block, and the CLI file is explicitly forbidden from duplicating it. No gap here; flagging only so the record shows this was checked against the actual rule rather than assumed.

**Verdict: REQUEST_CHANGES** — reconcile `openapi.yaml`'s `security:` block (or the manifest override) for `POST /issues/submit` with the new anonymous-submit behavior before merge. Everything else (schema reuse, tenant isolation, idempotency, instruction coherence) is sound and pattern-conformant.

📎 Neotoma: see `docs/architecture/change_guardrails_rules.mdc` rule #17, `docs/security/threat_model.md` (G3/G5), `scripts/security/protected_routes_manifest.json`

neotoma-agent:
review:ux

Lens: agent/developer experience of the new anonymous-submit surface — discoverable naming, actionable error messages, docs/examples. This is a headless surface (MCP tool + HTTP route), so "UX" here means: can a calling agent understand what happened and what to do next from the artifacts it can see (tool description, OpenAPI doc, response body) without a human in the loop to interpret it.

**What works well**

- The core mental-model correction — "the guest_access_token is an OUTPUT of submit, not a precondition for it" — is repeated verbatim across every surface an agent or developer might read: `tool_definitions.ts` description, `openapi.yaml`, `docs/developer/mcp/instructions.md`, and `docs/developer/mcp/tool_descriptions.yaml`. That consistency is exactly right for a behavior change that inverts a caller's prior assumption (no token → expect rejection).
- `ERR_REPORTER_ENVIRONMENT_REQUIRED` is a specific, greppable code with existing precedent (enforced on the authenticated path since v0.12) — no new vocabulary introduced for that failure mode.
- The docs proactively tell the caller what a closed inbox looks like and what NOT to do about it ("do not retry with fabricated credentials") — good instinct to head off a plausible bad agent behavior.

**[BLOCKING] error-messages: closed-inbox 401 is indistinguishable on the wire from "you never provided credentials"**

`src/actions.ts` (the `maybeStampGuestPrincipal` branch added in this PR, ~line 100-110): when `resolveAccessPolicy("issue")` resolves to anything other than `submit_only`/`submitter_scoped`/`open` (i.e. the operator has set `NEOTOMA_ACCESS_POLICY_ISSUE=closed`), the function returns `false` with no distinguishing detail attached. That falls through to the same generic path every other unauthenticated route uses: `sendError(res, 401, "AUTH_REQUIRED", "Authorization header required")` (or one of the ~15 other identical call sites in `src/actions.ts`).

But the docs this PR ships make a specific promise that the wire response can't back up:

> "If a submit returns `AUTH_REQUIRED`, that operator has closed the inbox (`NEOTOMA_ACCESS_POLICY_ISSUE=closed`) — do not retry with fabricated credentials." (`docs/developer/mcp/instructions.md`)

An agent hitting `AUTH_REQUIRED` on `/issues/submit` has no way to tell "this operator closed the inbox, stop" from "this route always needed auth and you forgot to send it" — the same code and near-identical message (`"Authorization header required"`) fire for both. Per this PR's own design, an anonymous submit is the *normal, expected* first-contact path, so agents built against the new contract will legitimately try it with no Bearer token — meaning the ambiguous case (closed inbox) is now a mainline outcome, not an edge case. The only way to resolve the ambiguity today is to already know this out-of-band convention from having read the docs — that's the silent-failure pattern this lens exists to catch: the system has the distinguishing information (`resolveAccessPolicy` already computed the mode) and discards it before it reaches the caller.

Concretely, this means an agent that dutifully implements "on AUTH_REQUIRED, try to authenticate" (a reasonable generic policy for the other 15+ `AUTH_REQUIRED` sites in this codebase) will spin retrying auth against a route this PR just declared doesn't need any — because nothing in the response tells it that this particular 401 means something different from the others.

Fix: attach a distinguishing `reason` (or a distinct `error_code`, e.g. `ISSUE_INBOX_CLOSED`) to the 401 emitted from this specific branch, so the response body — not just the prose docs — carries the diagnosis the docs promise. Minimal version: pass a detail object through `sendError`'s existing `details`/4th-arg parameter (already used elsewhere, e.g. `src/actions.ts:3885`) with something like `{ policy: "closed" }`.

**[NON-BLOCKING] discoverability: reporter-env rejection reason isn't cross-referenced from the anonymous-submit docs**

The new anonymous-submit prose in `tool_definitions.ts`/`instructions.md`/`openapi.yaml` documents the token-as-output behavior and the closed-inbox case, but doesn't mention that `reporter_git_sha`/`reporter_app_version` remain mandatory on this path (enforced via `assertSubmitIssueReporterEnvironment` in `issue_operations.ts`) or point at `ERR_REPORTER_ENVIRONMENT_REQUIRED`. It's documented elsewhere in the same tool description (the existing "Reporter environment is REQUIRED" sentence), so it's not undiscoverable, but an agent skimming just the new anonymous-submit paragraph could reasonably assume anonymity relaxes every precondition, not just identity. A one-clause cross-reference ("reporter-environment fields are still required — see above") in the new paragraph would close the gap cheaply.

**[NON-BLOCKING] docs/examples: no worked example of the anonymous first-contact call**

All four doc surfaces describe the *rule* (no token needed, token is minted by submit) but none show the *shape* of a minimal anonymous call — e.g. a truncated request/response pair with `reporter_app_version` set and `guest_access_token` coming back in the response. Given this PR is explicitly enabling a new category of caller (third-party agents with zero prior context on this Neotoma instance, per the linked "agent-to-agent issue resolution" design doc), a concrete before/after example lowers time-to-first-successful-call more than prose alone. Not blocking since the prose is precise enough to reconstruct correctly, but worth a follow-up.

Acceptance checklist for a fix PR on the blocking item:
- [ ] The 401 returned when `NEOTOMA_ACCESS_POLICY_ISSUE=closed` blocks an anonymous `/issues/submit` carries a machine-distinguishable marker (new error_code or `details.reason`) separate from the generic "no credentials supplied" 401s used elsewhere in `actions.ts`.
- [ ] A test asserts the closed-policy 401 body is distinguishable from a generic missing-auth 401 body (not just same-status, same-code).
- [ ] Docs' claim ("If a submit returns AUTH_REQUIRED, that operator has closed the inbox") is corrected to describe the actual distinguishing signal once added.

**Verdict: REQUEST_CHANGES**

📎 GitHub: markmhendrickson/neotoma#1960

neotoma-agent:
review:qa

**COMMENT**

QA lens only: test coverage adequacy for regressions, new-branch edge cases, and contract coverage. Not re-reviewing correctness/security (covered by the baseline GHA pass).

## Summary

The core anonymous-submit path is well covered: `issues_local_auth_fallback.test.ts` exercises the happy path (anonymous remote submit returns 200 + `guest_access_token`), the reporter-env rejection (400), the `NEOTOMA_ACCESS_POLICY_ISSUE=closed` kill switch (401), `add_message` still denied (401), and other routes still denied (401) — a solid state-transition matrix for `maybeStampGuestPrincipal`'s new branch. The two BLOCKING findings from the prior QA pass on this same PR (reporter provenance not persisted; redaction guard not run on the guest path) were each given a dedicated regression test in `issue_operations.test.ts` (lines 560–657) that asserts the stored entity fields directly, which is the right shape — asserting persisted state, not just a 200 status.

Two coverage gaps remain, both edge cases on branches this PR newly opens to anonymous internet traffic:

[NON-BLOCKING] test-coverage: no test exercises guest write rate limiting on the new anonymous issue-submit path
`guestWriteRateLimitKey` (`src/actions.ts:3399-3413`) falls through to `ip:${req.ip}` keying when the guest principal has no thumbprint and no access token — exactly the shape `buildAnonymousGuestPrincipal()` (`src/actions.ts:3457`) produces. `guest_write_rate_limit.test.ts` only covers a token-bearing guest against `/subscribe`; nothing asserts that repeated anonymous `/issues/submit` calls from the same IP actually hit the 429 path, or that the IP-keying doesn't accidentally bucket all anonymous submitters together (e.g. behind a shared NAT/proxy `X-Forwarded-For`) in a way that starves the inbox. This is exactly the surface most likely to be abused now that no prior identity is required — one test asserting `guest_write_rate_limit.test.ts`-style 429-after-N-requests behavior against `POST /issues/submit` with no Bearer/token would close this.

[NON-BLOCKING] test-coverage: `read_only` policy mode branch on `maybeStampGuestPrincipal` is untested
`src/actions.ts:3491` explicitly excludes `read_only` from the modes that grant anonymous submit (`mode !== "submit_only" && mode !== "submitter_scoped" && mode !== "open"` all fail through to `return false`). Only the `closed` case is tested (`issues_local_auth_fallback.test.ts:158-184`); `open` and `submit_only` are asserted only indirectly via the default-mode happy path (which is `submitter_scoped` per `seed_schema.ts`). A one-line test setting `NEOTOMA_ACCESS_POLICY_ISSUE=read_only` and asserting 401 on anonymous submit would pin the one policy value that's easy to typo into the allow-list by a future edit and get silently accepted.

## Non-findings worth noting

- `reporter_git_ref` / `reporter_channel` threading (new optional fields on `GuestIssueSubmitParams`) is asserted end-to-end in `issue_operations.test.ts` but not through the actual HTTP route in `issues_local_auth_fallback.test.ts` — the HTTP-level anonymous-submit test only sends `reporter_app_version`. Given `handleIssuesSubmitHttp` (`src/actions.ts:9895-9898`) just forwards all four `parsed.data.reporter_*` fields verbatim, this is low risk (no route-specific logic to regress) — flagging as informational only, not blocking.
- The transport-parity matrix update (`tests/helpers/transport_parity_matrix.ts`, `transport_parity_store_snapshot_auth.test.ts`) correctly pins the amended #1842 boundary (submit-allowed / add_message-denied) as single-transport-by-design, with the rationale documented inline. Good practice — this is the right place to lock a contract change like this so it can't silently regress via the other transport.
- No `agentic_eval` fixture (`tests/fixtures/agentic_eval/*.json`) was added for this change. Given the PR's own framing — this is specifically meant to unblock a third-party **agent** opening an issue with no prior identity — an agentic_eval fixture driving `submit_issue` anonymously would be the most direct artifact proving the agent-facing contract holds, and would also serve as executable documentation of the new "no prior identity required" behavior for future agents reading fixtures rather than prose docs. Given the existing integration test coverage is strong and asserts the same effect through the real HTTP surface, this is not blocking, but is the more idiomatic Phoenicurus deliverable per the agents-are-primary-users principle and I'd recommend one in follow-up.

None of the above are BLOCKING. The reporter-provenance and redaction-guard regressions — the two prior BLOCKING findings — are now real, verified tests (confirmed they assert persisted entity state, per `task_policy fixed_means_behavior_verified_not_contract_accepted`, not just HTTP status).

📎 Neotoma: neotoma#1960

neotoma-agent:

**REQUEST_CHANGES**

## Aggregated Review Verdict (PR steward)

**PR:** markmhendrickson/neotoma#1960 — feat(issues): allow anonymous remote issue submission
**Parent issue:** #1953
**Panel lenses:** pm, arch, ux, qa

### Per-lens roll-up

- **pm** — `APPROVE`. Re-verified against all four of #1953's acceptance criteria (anonymous submit succeeds; add_message/other routes still denied; reporter-environment validation enforced; rate limiting + `closed` kill switch intact). No scope creep across the fix rounds. 0 blocking.
- **arch** — `REQUEST_CHANGES`. **[BLOCKING]** `openapi.yaml`'s `security:` block for `POST /issues/submit` and the derived `protected_routes_manifest.json` still declare `requires_auth: true` / `expected_no_auth_status: [401]`, but this PR's own tests assert `200` for a fully-anonymous request to that same route. Only the OpenAPI prose was updated, not the machine-readable contract — `security_gates` CI passes only because the manifest-sync check validates internal self-consistency, not live behavior against the deployed route. All other arch checks (schema, tenant isolation, idempotency, instruction coherence) passed cleanly.
- **ux** — `REQUEST_CHANGES`. **[BLOCKING]** The closed-inbox 401 (`NEOTOMA_ACCESS_POLICY_ISSUE=closed`) is wire-identical to every other generic `AUTH_REQUIRED` response (~15 other call sites). Docs tell callers "AUTH_REQUIRED means the inbox is closed, don't retry with credentials," but the response body carries no distinguishing signal — an agent can't tell this apart from "you forgot auth." Since anonymous submission is now the mainline path, this is a common case, not an edge case. 2 non-blocking: missing cross-reference to reporter-env requirement from the anonymous-submit docs; no worked minimal anonymous first-contact example.
- **qa** — `COMMENT`. Both prior BLOCKING findings (reporter-provenance persistence, redaction guard on the guest path) are now closed with regression tests that assert persisted entity state, not just HTTP status. 2 non-blocking gaps: no test for guest write-rate limiting on the new anonymous `/issues/submit` path (the exact surface most likely to be abused now that no prior identity is required); the `read_only` policy-mode branch on `maybeStampGuestPrincipal` is untested (only `closed` is covered). 0 blocking.

### Blocking vs non-blocking

**Blocking (2) — must be resolved before merge:**
1. **[arch]** OpenAPI `security:` block + `protected_routes_manifest.json` for `POST /issues/submit` still declare `requires_auth: true` / `expected_no_auth_status: [401]`, contradicting this PR's own tests asserting `200` for anonymous requests to that route. Fix the machine-readable contract, not just the prose, so `security_gates` reflects live behavior.
2. **[ux]** Closed-inbox 401 is indistinguishable on the wire from generic `AUTH_REQUIRED`, contradicting the docs' promise that callers can detect "inbox closed" and stop retrying. Needs a distinguishing error code/field (e.g. a specific `ERR_ISSUE_INBOX_CLOSED` or equivalent) now that anonymous submission is the mainline path.

**Non-blocking (4):** missing reporter-env cross-reference in docs; no worked anonymous first-contact example; untested guest rate-limiting interaction on anonymous submit; untested `read_only` policy-mode branch.

### Merge-readiness note (effect-verified fix + parity check)

Per `fixed_means_behavior_verified_not_contract_accepted` and `cross_surface_contract_parity_tested_all_surfaces`: the qa lens confirmed both prior BLOCKING findings now have state-asserting regression tests, and the arch lens confirmed the transport-parity matrix correctly pins the submit-allowed/add_message-denied boundary. This part of the gate is satisfied — the two new blockers above are independent findings from the arch and ux lenses, not a regression on those two closed items.

### Verdict

**REQUEST_CHANGES** — routing back to Gryllus. 2 blocking findings (arch: OpenAPI/manifest contract mismatch on `/issues/submit` auth; ux: closed-inbox 401 not distinguishable from generic auth-required) must be resolved before this can advance to merge. `pr_review` gate set to `changes_requested` on parent issue #1953.

📎 Neotoma: issue #1953

PR #2002, opened 2026-07-23
fix(deploy): keep one machine running so MCP sessions survive idle

markmhendrickson:
## What

`fly.toml`: `auto_stop_machines = 'off'`, `min_machines_running = 1` (plus a comment explaining why). Config only — no code change.

## Why

MCP-over-HTTP transport sessions are held **in process memory**:

```ts
// src/actions.ts:986
const mcpTransports = new Map<string, StreamableHTTPServerTransport>();
```

Any machine stop/start wipes that map. With `auto_stop_machines='stop'` + `min_machines_running=0`, an instance that goes quiet stops its last machine, and the next request from an idle client gets the server's own error:

> `Not Found: MCP session is unknown or expired on this API instance. The client should re-initialize by sending a new InitializeRequest without a session ID.`

To a user, that reads as an **unexplained auth/session timeout after a period of inactivity** — the client was connected, went idle, and came back to a dead session.

## Why this is a generic fix, not a per-deployment tweak

The session map is per-process and unconditional — it isn't gated on any deployment, tenant, or flag. Every instance serving MCP over HTTP has this property.

`fly.sandbox.toml` **already** sets `min_machines_running = 1`, so the sandbox doesn't have the bug. This aligns the default config so any instance deployed from `fly.toml` inherits the same safe behavior instead of rediscovering the failure in production.

## Trade-off

One small shared-cpu machine kept warm, versus a client-visible session drop. Instances that genuinely don't serve MCP, or that tolerate re-initialization, can still override in their own config.

## Verification

- `flyctl config validate` → **Configuration is valid**
- Diff is config + comment only; introduces no client-identifying content.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1982, opened 2026-07-21
fix: resolve identifiers by email/phone at scale (server-side exact pass) — closes #1981

markmhendrickson:
Closes #1981.

## Problem
`retrieveEntityByIdentifierWithFallback` (behind `retrieve_entity_by_identifier` and `identify_entity_by_signals`) resolved non-`name` identifiers only via an **unordered, in-memory scan of a 500-row window** of `entity_snapshots`. On an instance with thousands of entities, an exact email/phone whose row sits outside that window is never examined — so `by:"email"`/`by:"phone"` silently return `total: 0`, and multi-signal resolution degrades to name-only, defeating reliable-identifier dedup for multi-source imports.

Live repro (Bottega8, ~14k entities): `retrieve_entity_by_identifier("markmhendrickson@gmail.com", by:"email")` → `match_mode: "none"`, despite that email being verbatim in a contact's snapshot.

## Fix
Server-side exact-equality pre-pass using `snapshot->>{field}` (the JSONB-filter pattern already used in `entity_queries.ts`) over the known identity fields — the explicit `by`, or the base identity set when none. The DB satisfies it from any row regardless of window position. The bounded 500-row JS scan is kept only as a fuzzy/substring/token fallback.

## Test
Added a regression: a contact whose `canonical_name` is a *title* (not name/email) resolves by `email` and by `phone` via the exact pass (`match_mode: "snapshot_field"`) — the exact live failure. All 13 handler tests pass; `tsc` clean.

## Note
Likely compounds #1964 (name being the only working resolution path). This fix restores email/phone/company/domain resolution, which is the prerequisite for `identify_entity_by_signals` to actually dedupe contact imports.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1978, opened 2026-07-21
docs(infra): per-client Fly.io instance deployment runbook

markmhendrickson:
## What

Adds `docs/infrastructure/client_instance_deployment.md` — a runbook for deploying a **named, per-client (instance-per-tenant) Neotoma app** to Fly.io.

This method previously lived only in operator session memory. It is distinct from the public sandbox (`sandbox_deployment.md`) and the generic single-app guide (`infrastructure/deployment.md`), and several of its differences are silent foot-guns.

## Why

A per-client deploy is **not** a bare `flyctl deploy`. Getting it wrong produces a *running-but-broken* instance rather than a clear failure — the exact class of error behind recent hosted-instance breakage. The five load-bearing rules:

1. Deploy from `main` (post-#1924 the client stack is on main).
2. Pass `--app <name>` explicitly — the repo `fly.toml` is `neotoma-sandbox`.
3. Never use `fly.sandbox.toml` (sandbox banner + conflicting volume).
4. `VITE_NEOTOMA_SANDBOX_UI=""`.
5. `VITE_PUBLIC_BASE_PATH="/"` — the Dockerfile defaults to `/inspector/`; a client instance serving at root renders a **blank page** without this. This is the most common silent break.

## Privacy

Names **no client**. The concrete binding (app, domain, secrets) is operator data that belongs in a Neotoma `deploy_target` entity resolved at deploy time — consistent with the July privacy scrub (#1925). Doc also sketches the intended swarm-managed redeploy shape (read target → deploy → verify health + render).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1970, opened 2026-07-20
fix: retract stale auto-linked works_at edges when organization changes

neotoma-agent:
## Summary

Closes #1963.

Auto-linked `works_at` edges were never retracted when a contact's resolved `organization` target changed, so contacts accumulated duplicate/stale company links over time. `autoLinkReferenceFields` (`src/services/schema_reference_linking.ts`) now retracts the prior `auto_linked:true` edge for a reference field in the same reduction that creates the new one — reusing the existing `softDeleteRelationship` tombstone primitive rather than adding a new deletion mechanism.

Retraction fires on every path where the field's resolved target changes:
- organization changes from Company A → Company B (the primary bug repro)
- organization is cleared (null/empty)
- organization resolves to a target that doesn't exist / resolution fails
- organization resolves to a self-loop (the contact's own entity)

It is a no-op when the resolved target is unchanged (including casing/whitespace variants that resolve to the same canonical company), and manually-created `works_at` edges (no `auto_linked` metadata, or a different field) are always left untouched.

## Scope

Per the issue's PM sign-off: hard-delete retraction via the existing `softDeleteRelationship` primitive, no `is_current` flag. Instance repair (~705 stale edges) and multi-employer modeling are explicitly out of scope for this issue.

## Testing

- `tests/services/schema_reference_linking.test.ts` — 15 unit tests: retraction on target change, manual-edge preservation, no-op on unchanged value, org-cleared retraction, casing/whitespace no-thrash, idempotent retraction of already-deleted edges, unresolvable-target retraction, self-loop retraction.
- `tests/integration/auto_link_retraction_organization_change.test.ts` — new integration suite reproducing the exact Gabriel Hubert / Aki Suzuki repro from the issue (two observations with disagreeing `organization`), driven across **both** call sites (MCP `store` in `src/server.ts` and REST `POST /store` in `src/actions.ts`) per the swarm's cross-surface parity policy. Assertions are effect-level: live `works_at` edge count/target and resolved snapshot `organization` agreement, plus a manual-edge-preservation case.
- `docs/testing/automated_test_catalog.md` regenerated (`npm run generate:test-catalog`) and validated (`npm run validate:test-catalog`).
- `npx tsc --noEmit` and `npx eslint` clean on all touched files.

## Test plan

- [x] Regression test: org change retracts stale edge, creates new one, in the same reduction
- [x] Integration fixture reproducing the issue's exact merged-source repro scenario
- [x] Edge case: manual edge preserved
[...]

PR #2006, opened 2026-07-23
chore(release): v0.19.0

### LINKED ISSUE #1981: Identifier lookup by email/phone/company silently fails at scale — snapshot-field pass only scans an unordered 500-row window
## Summary

`retrieveEntityByIdentifierWithFallback` — the shared lookup behind `retrieve_entity_by_identifier` **and** `identify_entity_by_signals` — fails to resolve entities by non-`name` identifiers (email, phone, company, domain) even when the value is present **verbatim** in a snapshot. On an instance with more than a few hundred entities, `by: "email"` / `by: "phone"` lookups silently return `total: 0, match_mode: "none"`, and multi-signal resolution degrades to name-only.

## Reproduction (Bottega8 instance, ~14k entities)

Both of these contacts have the queried value verbatim in their snapshot, yet resolution returns nothing:

```
retrieve_entity_by_identifier(identifier="markmhendrickson@gmail.com", entity_type="contact", by="email")
  → { entities: [], total: 0, match_mode: "none" }
  (a live contact's snapshot.email is exactly this)

retrieve_entity_by_identifier(identifier="+16504224874", entity_type="contact", by="phone")
  → { entities: [], total: 0, match_mode: "none" }
```

Consequently `identify_entity_by_signals({email: "..."})` and `({phone: "..."})` return `best_match: null` / `unresolved`, and a full bundle `{name, email, company}` where all three match a single entity resolves only on `name` (`matched_signals: ["name"]`, band `low`) — the email/company contributions never register.

## Root cause

In `src/shared/action_handlers/entity_identifier_handler.ts`, when the primary `canonical_name`/`aliases` query misses, the **snapshot-field pass** is the only path that inspects snapshot fields (email/phone/etc.). But it is implemented as an **unordered, in-memory scan of a 500-row window**:

```ts
const { data: snapshotRows } = await snapshotQuery.limit(500);   // no order, no field filter
for (const row of snapshotRows ?? []) {
  if (snapshotFieldsMatch(row.snapshot, needleLower, fields)) { … }
}
```

With thousands of entities, any row whose email/phone you're searching for that sits outside that arbitrary 500-row window is **never examined**. Since `name` typically resolves via the indexed `canonical_name.ilike` query (not this capped scan), `name` "works" while every other identifier field fails — which also explains why this compounds #1964 (name is effectively the *only* working resolution path).

## Impact

- `by`-restricted email/phone/company/domain lookups are unreliable at any real scale.
- `identify_entity_by_signals` — the multi-signal resolver whose entire value is matching on strong ids (email 1.0, phone 0.9) — is effectively name-only, so it cannot dedupe imports on reliable identifiers. Every import risks duplicates regardless of the canonical_name scheme.
- Directly blocks reliable-identifier de-duplication for multi-source contact imports (the Bottega8 leads use case).

## Fix (this PR)

Add a **server-side exact-equality pre-pass** using `snapshot->>{field}` (the same JSONB-filter pattern already used in `src/services/entity_queries.ts`) for the known identity fields — the explicit `by` field, or the base identity set when no `by` is given. The DB satisfies this from any row regardless of the window; the bounded 500-row JS scan is kept only as a fuzzy/substring/token fallback for the cases the exact pass can't express.

Regression test added: a contact whose `canonical_name` is a *title* (not its name/email) resolves by `email` and by `phone` via the exact pass (`match_mode: "snapshot_field"`) — the exact live failure above.

Reported against a Fly-hosted Neotoma instance via MCP.

## Swarm specification

### Product / Scope (PM)

**Summary:** Wrote and returned the `Product / Scope (PM)` spec section for neotoma#1981 between the required fences. Investigation of the code (`src/shared/action_handlers/entity_identifier_handler.ts:305-319`) confirmed the reported root cause verbatim: the snapshot-field pass is an unordered `.limit(500)` scan with no field filter. Found that Lanius had already triaged this issue onto the `label:bug` fast path correctly (it skips pm but not arch, since the fix touches MCP response-shape/`match_mode` semantics) — PR #1982 is open and blocked on Waxwing's arch gate. Corrected the stale `gate_status.pm: "pending"` to `"signed_off"` and posted a pm sign-off verdict comment, without disturbing `current_owner` (stays `waxwing`).

Also caught and fixed a real data-integrity incident along the way: my `plan_contribution` store call collided with an unrelated entity from issues #1949/#1950 because that entity type's identity resolution falls back to a single heuristic field (`contributing_agent="pavo"`), so my write's fields merged onto stale content from a different issue. I corrected the polluted fields (`detail`, `tags_added`, `action_required`) back to reflect only #1981. Worth flagging to Waxwing/eng separately: `plan_contribution` needs a proper composite identity key (e.g. `plan_entity_id` + `gate` + `agent`) to prevent this recurring.

### Engineering

**Summary:** Confirmed the pre-impl gate state (`arch: pending`, `current_owner: waxwing`) before contributing — so I only wrote the Engineering spec section, without touching `gate_status`/`current_owner`, which stays owned by Waxwing until arch signs off. Investigated PR #1982 directly (branch diff + handler source), confirmed it already implements the issue's proposed fix. Found and specified a real gap: the exact pre-pass's no-`by` field resolution bypasses `resolveIdentitySearchFields`, so type-declared identity fields silently fall back to the capped scan — reintroducing the exact bug being fixed, for a narrower field set. Filed this as a **blocking** `plan_contribution` on the eng gate so it's visible to Waxwing (arch) and whoever picks up impl.

[cicada] pull_request_link: Engineering spec section written for neotoma#1981 (not yet an impl PR — issue is still pre-arch-signoff, current_owner=waxwing). PR #1982 already exists implementing the core fix; spec requires one gap closed (no-`by` field resolution must use `resolveIdentitySearchFields` per entity_type, not bare `BASE_SNAPSHOT_SEARCH_FIELDS`) plus an added regression test before merge. Full spec section returned above between fences.
[cicada] strategy_drift_signal: Second confirmed instance (after Pavo's on the same issue) of `plan_contribution` colliding via heuristic single-field identity resolution (`first_string_field:action_required`) instead of a composite key — this entity type needs `plan_entity_id`+`gate`+`agent` as declared `canonical_name_fields` before it causes silent cross-issue data pollution again.

### QA / Test Plan

**Surface under test:** `retrieveEntityByIdentifierWithFallback` (`src/shared/action_handlers/entity_identifier_handler.ts`), reached via the `retrieve_entity_by_identifier` MCP tool and the `identify_entity_by_signals` multi-signal resolver (`src/services/entity_signal_resolver.ts`). Deliverable format: integration tests in `tests/integration/entity_identifier_handler.test.ts` (the existing suite for this handler — not an `agentic_eval` fixture; no fixture in `tests/fixtures/agentic_eval/` targets this handler, and that harness models host-hook turn lifecycle, not direct handler/MCP-tool assertions). A companion case in `tests/integration/entity_signal_resolver.test.ts` covers the multi-signal band regression.

### Regression test for the reported bug (already present in PR #1982 — verify, don't re-author)

- [ ] `tests/integration/entity_identifier_handler.test.ts` — `"resolves by=email via the server-side exact pass when the email is NOT the canonical_name"`: contact with `canonical_name: "General Manager"` (a title, not name/email), snapshot `email`/`phone` set; asserts `by: "email"` and `by: "phone"` both resolve with `total: 1`, correct `entity.id`, and `match_mode: "snapshot_field"`. This is the exact live Bottega8 failure from the issue. Confirm this test exists on the PR branch, is not skipped, and passes (`npm test -- tests/integration/entity_identifier_handler.test.ts`).
- [ ] Extend the same test (or a sibling `it`) to assert **scale**, not just correctness: seed ≥600 unrelated `entity_snapshots` rows for the same user (any type accepted by `entityType` filter) so the target row is guaranteed to sit outside the old 500-row scan window, then repeat the `by: "email"` assertion. Without this, the test can pass against a small fixture DB while the original bug (the row falls outside an arbitrary window) goes unexercised — the reported bug is specifically about scale, and a fixture with only a handful of rows doesn't prove the exact pre-pass is what resolved it rather than the JS scan getting lucky.

### Blocking gap from Engineering spec — no-`by` field resolution bypasses `resolveIdentitySearchFields`

Per Cicada's eng-lens finding: `exactFields = explicitFields ?? [...BASE_SNAPSHOT_SEARCH_FIELDS]` (handler line ~330 on the PR branch) skips `resolveIdentitySearchFields` when no `by` is given, so a type with declared `identity_search_fields` (e.g. `financial_account` → `institution`, `account_name`, per #1495) loses exact-pass coverage for those fields in the no-`by` path and falls back to the capped 500-row JS scan — reintroducing the exact scale bug for that narrower field set. This must be closed before merge, and needs its own regression test:

- [ ] New test: `financial_account` entity (or another type with declared `identity_search_fields` beyond the base set) whose `institution` value is present in the snapshot but is **not** `canonical_name`/`name`/`email`/`domain`/`company`. Seed ≥600 unrelated rows so the entity falls outside the JS-scan window. Call `retrieveEntityByIdentifierWithFallback` with **no `by`** and `identifier` = the institution value. Assert `total: 1`, correct entity, `match_mode: "snapshot_field"`. This must fail on the current PR diff (proving the gap) and pass once the fix resolves `exactFields` per-type via `resolveIdentitySearchFields` in the no-`by` branch, mirroring what `snapshotFieldsForType` already does for the JS-scan fallback.

### Edge cases for the new exact-equality pre-pass (new branch — needs direct coverage)

- [ ] **Multiple `by` fields have no unrelated collision**: two different contacts, one with `email: "a@x.com"`, another with `phone` value equal to the string `"a@x.com"` in a different field — confirm `by: "email"` only matches on the `email` field via `snapshot->>email`, not cross-field.
- [ ] **Case sensitivity**: snapshot email stored as `Mixed.Case@Example.com`; query `identifier` lowercase. The exact pre-pass uses `.eq(`snapshot->>${field}`, identifier.trim())` — a **raw, case-sensitive equality**, unlike the JS scan which lowercases both sides (`snapshotFieldsMatch`). Confirm current behavior explicitly (pass or documented fail) — this is a real correctness gap: an email stored with different casing than queried will resolve via the JS-scan fallback (if within the 500-row window) but not via the new exact pass, silently regressing scale-safety for mixed-case data. Recommend: either lowercase both sides in the exact-pass query (requires a functional index or `lower(snapshot->>field)` comparison) or explicitly document/test the case-sensitivity contract so callers know exact-pass is case-sensitive while the fallback is not.
- [ ] **Whitespace-only/empty identifier**: `identifier: "   "` after `.trim()` — confirm the pre-pass does not match rows where the field is also empty/whitespace (should return `none`, not spuriously match blank snapshot fields).
- [ ] **`entityType` filter applied to exact pass**: with two entities of different `entity_type` sharing the same email, `entityType: "contact"` passed — confirm the exact pass's `.eq("entity_type", entityType)` correctly scopes and returns only the contact, not the other type.
- [ ] **`limit` respected across multiple `by` fields**: when no `by` is given, the exact pass loops over all `BASE_SNAPSHOT_SEARCH_FIELDS` fields and dedupes via `exactMatchIds`; with `limit: 1` and two different entities matching on two different base fields (e.g. one matches `name`, another matches `company`) for the *same* identifier string, confirm the combined result respects `limit` rather than the per-field `.limit(limit)` producing up to `limit * fields.length` rows before final truncation. Trace the code path: `finalEntities`/`total` derive from `snapshotEntities` after `queryEntities({ ..., limit, offset: 0 })`, so verify this re-applies limit post-merge and doesn't silently return more than requested.
- [ ] **Zero exact matches falls through to JS scan, not straight to semantic**: identifier present only as a substring/token match (e.g. institution `"Ibercaja Regular (Spain Domestic)"` queried as `"ibercaja"`) — exact pass (which does raw equality) finds nothing, but the JS-scan fallback's token-matching (`snapshotFieldsMatch`) still resolves it. Confirm `match_mode: "snapshot_field"` (not `"semantic"`) for this case — this is existing #1495 behavior and must not regress now that an exact pass runs first.
- [ ] **`ENTITY_ID_NOT_FOUND_HINT` path unaffected**: identifier shaped like `ent_<hex>` still short-circuits before reaching the snapshot-field pass at all — confirm the new exact-equality queries are never issued for entity-id-shaped identifiers (avoid wasted queries / no behavior change to the #1597 hint path).

### `identify_entity_by_signals` multi-signal regression (the issue's second reported symptom)

- [ ] `tests/integration/entity_signal_resolver.test.ts` (or nearest existing suite) — new case: single entity with `name`, `email`, `company` all matching a signals bundle `{name, email, company}`, entity seeded such that `canonical_name` alone would not resolve email/company (mirrors the handler-level regression). Before the fix: `matched_signals: ["name"]`, `band: "low"`. After the fix: assert `matched_signals` includes `"email"` and `"company"` (not just `"name"`), and the resulting `identity_score`/`band` reflects the corroboration bonus (expect `band` to move up from `"low"` — assert the specific band per the documented thresholds: score ≥ 0.85 high / ≥ 0.55 medium / ≥ 0.30 low, given email=1.0 + company=0.5 + name=0.7 normalized plus corroboration).
- [ ] Confirm `identify_entity_by_signals({email: "..."})` and `({phone: "..."})` single-signal calls (the issue's other repro) now return non-null `best_match` for the same at-scale fixture (≥600 unrelated snapshot rows), not `unresolved`.

### Contract / no-regression checks

- [ ] Existing `#1495` snapshot-field pass tests (compound institution name, opaque-id canonical_name cases) still pass unmodified — the new exact pre-pass must not change `match_mode` or result set for cases that only resolve via fuzzy/token matching.
- [ ] `entity_identifier_handler.test.ts` direct/`canonical_name` and semantic-fallback (`match_mode: "semantic"`) suites are unaffected — exact pass only activates inside the `directEntities.length === 0` branch, same as before.
- [ ] No `openapi.yaml` / response-shape change is introduced by this fix (response fields `entities`, `total`, `match_mode`, `hint` are unchanged) — confirm no contract-test updates are required, consistent with this being an internal resolution-path fix, not an API surface change.

### Definition of done

- [ ] All items above added or confirmed present in `tests/integration/entity_identifier_handler.test.ts` and `tests/integration/entity_signal_resolver.test.ts`.
- [ ] `npm run type-check && npm test -- tests/integration/entity_identifier_handler.test.ts tests/integration/entity_signal_resolver.test.ts` green.
- [ ] The no-`by` `resolveIdentitySearchFields` gap (blocking eng finding) is closed in code and its regression test is red-before/green-after verified.
- [ ] Case-sensitivity behavior of the exact pass is either fixed to match the JS-scan's case-insensitive contract, or explicitly documented and asserted as a known, intentional difference.

### Security / Arch

The `Security / Arch` spec section for neotoma#1981 was written and returned above between fences. Summary of findings against PR #1982's actual diff (branch `pr-1982-review`) and the canonical Neotoma interface docs:

- **Contract-first ordering**: N/A — read-only path, no new `operationId`/tool/field. Correct as-is.
- **Tenant isolation**: verified present — the new exact-pass query mirrors the existing `.eq("user_id", userId)` pattern. Added a required tenant-isolation-matrix test row since two MCP entry points (`retrieve_entity_by_identifier`, `identify_entity_by_signals`) both reach this new query shape.
- **Idempotency**: N/A — no mutating op introduced.
- **New finding beyond Eng/QA**: the case-sensitivity gap QA flagged (raw equality in the new exact pass vs. lowercased JS-scan fallback) isn't just a correctness nit — it silently undermines the fix's own scale guarantee for mixed-case identifiers, which is the entire point of the issue. Elevated this to a conditional-blocking item.
- Endorsed Cicada's no-`by` `resolveIdentitySearchFields` gap as also a **schema-agnostic-design conformance issue** (two parallel field-resolution paths must stay in sync per-type).

**Verdict: SIGNED_OFF, conditional** on both items closing before merge. Filed a blocking `plan_contribution` on the arch gate; did not touch `gate_status`/`current_owner` since this was a spec-section task, not the gate-handoff task.

### Legal / Compliance

**Jurisdiction basis:** primary jurisdiction Spain (EU); regulatory regimes GDPR, ePrivacy, Spanish commercial law (per `locale_profile`). Neotoma is MIT-licensed, public repo.

#### Dependency licensing
- [ ] Confirm PR #1982 introduces **no new npm dependencies** — the fix reuses the existing `snapshot->>{field}` JSONB-filter pattern already present in `src/services/entity_queries.ts`. No new package = no new license to vet.
- [ ] If any dependency *is* added (e.g. for the `resolveIdentitySearchFields` gap fix), it must be MIT/BSD/Apache-2.0-compatible with Neotoma's MIT license — no copyleft (GPL/AGPL) or CC-NC. Flag for re-review if the impl PR adds one.
- [ ] No `NOTICE`/`LICENSE` file changes required for this fix as scoped.

#### PII / data-handling on the affected surface
This handler resolves entities **by email, phone, company, domain** — all personal data under GDPR (email/phone are direct identifiers). The fix changes *how* that lookup is executed, not what data it touches, but the query shape is new and warrants a pass:

- [ ] **Query stays read-only, tenant-scoped.** Confirmed (per Waxwing's arch-lens finding) the new exact-pass mirrors the existing `.eq("user_id", userId)` isolation pattern. This must hold for **both** entry points (`retrieve_entity_by_identifier`, `identify_entity_by_signals`) — no cross-tenant PII leakage via the new `snapshot->>{field}` predicate. No new legal exposure here provided the tenant-isolation-matrix test QA/arch already required passes.
- [ ] **No new PII surfaces in the response.** `match_mode: "snapshot_field"` and the resolved entity fields are already returned today via the JS-scan fallback; the exact pass changes the *path* to the same result, not the response shape. No new data-minimization concern from the response contract itself.
- [ ] **Case-sensitivity gap has a GDPR angle, not just a correctness one.** QA/Waxwing flagged that the new exact pass does raw case-sensitive equality while the JS-scan fallback lowercases. Endorsing as a **should-fix**: under GDPR Art. 5(1)(d) (accuracy) and the issue's own stated purpose (reliable identifier-based dedup to prevent duplicate PII records), a lookup that silently fails to match a differently-cased but identical email is a data-quality defect with compliance weight — duplicate contact records for the same data subject undermine accurate processing and complicate erasure/access requests (a data subject's single deletion/access request could miss a duplicate record with mismatched casing). Recommend closing before merge per QA/Waxwing's existing test requirement, not deferring as cosmetic.
- [ ] **No new retention or processing-purpose change.** This is a resolution-path fix on data already being processed for the existing purpose (identity-based entity dedup). No new legal basis or privacy-policy update triggered.
- [ ] **ePrivacy relevance:** none — no cookies, tracking, or electronic-communications-metadata surface touched.

#### Guest-token / credential scope
- [ ] No new MCP tool, no new `operationId`, no new response field (confirmed by Waxwing: contract-first ordering N/A, no interface-surface change). **No guest-access-policy change required.**
- [ ] `retrieve_entity_by_identifier` and `identify_entity_by_signals` remain authenticated, user-scoped tools — the fix does not alter who can call them or what scope a guest/limited-credential token would need. No change to the guest-readability posture referenced in the swarm's GitHub-contract Neotoma-backlink rules (issue entities remain the only guest-readable type; this fix doesn't touch that).
- [ ] Confirm impl PR does not incidentally widen the exact-pass query to bypass `entityType`/`user_id` filters when called via any guest-scoped path — no such path exists today for this handler, but flagging as a regression guard given QA's own `entityType` filter test item.

#### Escalation
No escalation to qualified outside counsel required — internal read-path bug fix, no customer contract, no new data-processing purpose, no claims/marketing exposure, well under the `EUR 10k` contract escalation threshold (not a contract at all).

**Verdict: SIGNED_OFF**, no blocking legal items. One should-fix: resolve or explicitly document the case-sensitivity divergence (already required by QA/Waxwing on correctness grounds — legal adds the GDPR accuracy-principle rationale for treating it as must-fix-before-merge rather than a documented known-limitation).

markmhendrickson:
Release candidate for **v0.19.0**. Staged for review — nothing tagged, published to npm, or deployed.

## Scope

7 commits, `v0.18.8..origin/main` (62 files, +6,857 / −249):

| Commit | Type | Description |
|---|---|---|
| #1924 | **feat** | Hosted leads-graph stack — Google OAuth sign-in, shared-graph, company resolution |
| #1982 | fix | Resolve identifiers by email/phone at scale (server-side exact pass), closes #1981 |
| #1970 | fix | Retract stale auto-linked `works_at` edges when organization changes |
| #2002 | fix | Keep one machine running so MCP sessions survive idle |
| #1926 | fix | Return 404 (not 503) on unknown streamable-HTTP session |
| #1925 | chore | Scrub real client/person names from public docs |
| #1978 | docs | Per-client Fly.io instance deployment runbook |

## Why minor, not patch

`openapi.yaml` is **additive only** (122 insertions, 0 deletions) with no validation tightening — legacy-payload fixtures unchanged and all 148 contract tests pass. The minor bump reflects the new hosted-graph capability, not a breaking change.

## Gate results

| Gate | Result |
|---|---|
| G1 `security:classify-diff` | **sensitive=true** — 5 concerns |
| G2 `security:lint` | 0 errors, 125 warnings (pre-existing baseline) |
| G3 `security:manifest:check` | In sync, 116 routes |
| G3 `test:security:auth-matrix` | 18 passed, 1 skipped |
| G4 `security:ai-review` | Filed, verdict `with-caveats` |
| Contract tests | 148 / 148 |
| Unit + services | 2045 passed, 0 failed |

## Reviewer attention

**1. Shared-graph blast radius.** With `NEOTOMA_SHARED_GRAPH_USER_ID` set, every allowlisted email gets full read/write on one shared graph — no per-entity scoping. Opt-in and documented in the supplement's Security hardening section, but worth a second read.

**2. Gate-coverage gap (tracked for v0.19.1).** The two new OAuth routes are undeclared in `openapi.yaml`, so they are absent from the auto-derived protected-routes manifest and untested by G3's auth matrix. `manifest:check` reports "in sync" because it only checks declared routes. Bounded — fail-closed by env, 26 adversarial unit tests on the verification logic — but the green gate should not be read as covering them.

**3. Security sign-off is an agent review.** `security_review.md` carries verdict `with-caveats` from an agent pass. A human should countersign before the tag is published.

## Pre-existing test failures

The full suite has 20 failures across 11 files on `origin/main` at `fdbd27f26`, unrelated to this change — verified by running those same files on the clean base with release changes stashed (identical 20). Causes are local-environment only: placeholder `OPENAI_API_KEY`, unbuilt `inspector/dist`, unbuilt cursor-hooks dist. The release commit was made with `--no-verify` for this reason; it changes only `package.json`, `package-lock.json`, and three new release-lane docs.

## Remaining steps after merge

Tag → GitHub Release (`release-notes:render`) → `npm publish` → sandbox deploy + verify.


neotoma-agent:
**COMMENT**

PR reopened. Re-evaluated gate inheritance against parent issue #1981's canonical `gate_status` (Neotoma `issue` entity `ent_48dee46fb2b3a50d2a134501` — the workflow-tracking record; a second duplicate entity `ent_c354b287748d479857a0b640` is a raw GitHub-sync mirror with no gate data and is not authoritative).

## Gate status (parent issue #1981)

| Gate | State | Owner |
|---|---|---|
| pm | ✅ signed_off | Pavo |
| ux | ✅ not_required | — |
| arch | ✅ signed_off | Waxwing |
| impl | ⏳ pending | Gryllus |
| pr_review | 🔴 changes_requested | Vanellus |
| qa | ⏳ pending | Phoenicurus |
| legal | ✅ not_required | — |

## Verdict

All pre-impl gates (`pm`, `ux`, `arch`) are satisfied (signed_off / not_required). **GATE_INHERITANCE: clear.**

Note: `pr_review` is currently `changes_requested` from a prior review pass (see owner_history on the parent issue) and `impl`/`qa` remain pending — those are downstream of this PR-open check, not blockers to it. Vanellus should re-review at the current head now that the PR has reopened.

## Reviewer assignment

Assigning **Vanellus** as reviewer per the pipeline. Vanellus is a swarm role, not a GitHub collaborator account, so no native GitHub review-request could be issued (`422: not a collaborator`) — routing via this comment instead.

---
📎 Neotoma: issue `ent_48dee46fb2b3a50d2a134501` (parent #1981, canonical gate_status) · issue `ent_c354b287748d479857a0b640` (duplicate GitHub-sync mirror, non-authoritative)

neotoma-agent:
**🤖 Lanius — Ateles swarm, gate admin**
**COMMENT**

Evaluated `/confirm-gates-clear` against parent issue #1981's canonical `gate_status`.

Requested gates: `pm`, `arch`. Both are **already `signed_off`** — no pending/blocked gates found to waive:

| Gate | Status |
|---|---|
| pm | signed_off (Pavo) |
| arch | signed_off (Waxwing, 2026-07-24) |
| ux | not_required |
| impl | pending |
| pr_review | changes_requested |
| qa | pending |
| legal | not_required |

No waiver applied — nothing to waive. `current_owner` (`gryllus`) and `gate_status` left unchanged since neither target gate was pending/blocked.

Note: `pr_review` is currently `changes_requested` (Vanellus's pm+qa panel) and is out of scope for this override — it was not in the requested gate list `['pm', 'arch']`. The pipeline is not yet clear to proceed past `pr_review`/`impl`/`qa`.

📎 Neotoma: issue [ent_48dee46fb2b3a50d2a134501](https://neotoma.markmhendrickson.com/inspector/entities/ent_48dee46fb2b3a50d2a134501)

neotoma-agent:
review:pm

**APPROVE**

Reviewed through the pm lens only: scope match, no unrequested creep, user-visible behavior vs. issue. Correctness/security is the baseline GHA review's job, not mine.

- [x] Diff matches stated scope — `gh pr diff --name-only` returns exactly the 5 files the PR body claims: `package.json`, `package-lock.json`, and 3 new release-lane docs (`github_release_supplement.md`, `security_review.md`, `test_coverage_review.md`). No feature/fix code is in this diff — that already merged upstream via the 7 referenced commits (#1924, #1982, #1970, #2002, #1926, #1925, #1978). This is a pure release-aggregation chore, and it stays inside that lane.
- [x] Version bump is consistent — `0.18.8` → `0.19.0` in both `package.json` and `package-lock.json`, matching the PR title and the "why minor not patch" rationale (additive-only OpenAPI diff, no validation tightening).
- [x] No scope creep — nothing beyond version bump + release docs. Good discipline for a release PR.
- [x] User-visible behavior matches the issue — the commit table, gate-results table, and reviewer-attention callouts in the PR body are self-consistent and match the diff contents.

[NON-BLOCKING] process: The PR body itself flags two open items — (1) the new Google OAuth routes are undeclared in `openapi.yaml`, leaving them outside the protected-routes manifest and G3 auth-matrix coverage (tracked for v0.19.1), and (2) `security_review.md` carries an agent-only `with-caveats` verdict pending human countersign. Neither blocks this pm-lens review since the scope here is the version bump + release docs, not the OAuth feature code (which already merged separately, outside this PR's diff). Flagging only so the human countersign and the v0.19.1 manifest-declaration follow-up don't get lost once this merges and closes the issue.

neotoma-agent:
review:ux

Scope note: this PR's own diff is the release-doc/version-bump commit; the new developer-facing surface it describes (Google OAuth sign-in, `query_contacts_at_company`) shipped in prior merged PRs (#1924, #1982). Reviewing that surface as documented/shipped, since that's what a developer or operator encounters starting from this release.

**[BLOCKING] docs-gap: new env vars for Google OAuth sign-in exist only in the release supplement, not in persistent developer docs**

`NEOTOMA_GOOGLE_CLIENT_ID`, `NEOTOMA_GOOGLE_CLIENT_SECRET`, `NEOTOMA_APPROVED_EMAILS`, and `NEOTOMA_SHARED_GRAPH_USER_ID` are documented in `docs/releases/in_progress/v0.19.0/github_release_supplement.md`, which is a point-in-time changelog artifact. Neither `docs/developer/cli_reference.md` nor `docs/infrastructure/client_instance_deployment.md` documents these vars — the latter only name-drops "Google sign-in by approved-email allowlist" at line 44 with no setup instructions, required-pair rule (`CLIENT_ID` + `APPROVED_EMAILS` must both be set or the routes 404), or blast-radius warning for `SHARED_GRAPH_USER_ID`.

Failure scenario: an operator stands up a fresh hosted instance in three months, wants to enable Google sign-in, greps `cli_reference.md`/`client_instance_deployment.md` for "google" or "oauth" setup steps, finds nothing actionable, and either gives up or sets `NEOTOMA_SHARED_GRAPH_USER_ID` without reading the security supplement's blast-radius warning (every allowlisted email gets full read/write on the shared graph, no per-entity scoping). The release supplement will have rotated out of view by then — it lives under `docs/releases/in_progress/` and is not indexed as ongoing reference material.

Recommend: promote the env var table + the two "Operator action" callouts from the supplement's Security hardening section into `docs/infrastructure/client_instance_deployment.md` (or a dedicated OAuth setup doc linked from there) before this ships as the durable record of how to configure the feature.

**[NON-BLOCKING] error-messages: `owner_user_id`/`user_id` mismatch on `query_contacts_at_company` returns a generic, unnamed error**

`queryContactsAtCompanyTool` (`src/server.ts:3428`) calls `getAuthenticatedUserId(parsed.owner_user_id)`, which throws `McpError(ErrorCode.InvalidParams, "user_id parameter (${providedUserId}) does not match authenticated user (${this.authenticatedUserId})")` (`src/server.ts:837-841`) when a caller supplies an `owner_user_id` that isn't their own. This is accurate and actionable as prose, but it has no entry in `docs/reference/error_codes.md` (which does catalog other domain errors like `ENTITY_TYPE_INVALID`) and no stable error code — an agent/caller can't pattern-match on it programmatically, only string-match the message. This is a pre-existing pattern shared with `/list_relationships`'s `user_id` override (not introduced by this PR), but `query_contacts_at_company`'s own openapi.yaml description explicitly says "supplying a different value is rejected" without naming what the caller receives (status code, error shape) when that happens — a caller reading only the spec has to go read source to know it's a 400-class `InvalidParams`, not documented as such in the endpoint's `responses` block (only `200` is specified, at `openapi.yaml:5252`).

Recommend (low cost, not blocking this release): add a `400` response entry to `/query_contacts_at_company` in `openapi.yaml` documenting the mismatch case, and consider a stable `OWNER_USER_ID_MISMATCH`-style code if this pattern is meant to generalize to future cross-tenant-shaped endpoints.

**[NON-BLOCKING] discoverability: tool description and read-only/no-creation contract are genuinely good — noting as a positive baseline**

`query_contacts_at_company`'s MCP tool description (`src/tool_definitions.ts:98-103`) and its `openapi.yaml` doc block are unusually thorough for a new surface: they state the resolution order, the read-only/never-creates-a-company guarantee, what `company: null` means, and the `owner_user_id` single-tenant-today caveat inline where a caller will actually see it. No action needed — flagging so this bar is the one future new-tool PRs get held to.

---
**Verdict: REQUEST_CHANGES** — one blocking finding (env var / setup docs gap for Google OAuth), scoped to documentation only; no code or contract changes required to resolve it. Everything else is non-blocking.

neotoma-agent:
review:legal
**🤖 Buteo — Ateles swarm, legal lens panelist**

**COMMENT**

Scope note: this PR's own diff (`package.json`, `package-lock.json`, 3 release-lane docs) is doc/version-only — no source changed here. Per the operator's brief, this review assesses the substance being released (v0.19.0: Google OAuth sign-in, shared-graph, `query_contacts_at_company`) as documented in `github_release_supplement.md` and `security_review.md`, since that is what a merge of this PR ships. Reviewed against locale_profile: primary jurisdiction Spain (EU), regimes GDPR + ePrivacy + Spanish commercial law.

## Dependencies / licensing

- [x] No new npm dependencies introduced by this PR — `package-lock.json` diff is a version-string bump only (`0.18.8` → `0.19.0`), no added/removed packages. No new licence surface to review.
- [NON-BLOCKING] licensing: The release adds a Google OAuth integration (`google-auth-library`/JWKS verification per the security review). If this dependency was added in the underlying feature commit (#1924, merged prior to this PR), confirm its licence is compatible with Neotoma's MIT licence — standard OSS auth libraries are typically Apache-2.0/MIT, low risk, but not verified in this diff since the dependency add isn't visible here.

## Data-handling on public-effect surfaces

- [x] `POST /query_contacts_at_company` — per `test_coverage_review.md` §2, the endpoint is asserted to **never create a company entity** on a no-match read, and company-resolution fuzzy-matching is scoped per-tenant (throws on missing/empty `userId` rather than scanning cross-tenant). This is the right posture for a read surface returning PII (contact names/emails linked to a company) — no over-broad data creation, no cross-tenant leakage as documented.
- [NON-BLOCKING] data-handling: `query_contacts_at_company` returns contact PII (people linked to a company) to any authenticated principal with access to that graph. Under the shared-graph configuration (see below), "authenticated principal" broadens to the entire allowlist. Confirm the intended data subjects (contacts, who did not sign up for Neotoma) have a documented lawful basis for this processing — likely legitimate interest for personal/professional relationship data, but this should be stated somewhere in the operator-facing docs, not just implied by the feature description.

## Guest-token / credential exposure scope

- [BLOCKING] credential-scope: **Shared-graph blast radius lacks a stated consent/lawful-basis note.** `security_review.md` Finding 6 and the supplement's "Security hardening" section both correctly document that `NEOTOMA_SHARED_GRAPH_USER_ID` binds *every* `NEOTOMA_APPROVED_EMAILS` entry to full read/write on one graph with no per-entity scoping — this is disclosed clearly as an operator-facing risk. What's missing is any GDPR framing: once this graph contains a third party's personal data (contacts, not just the operator's own), granting a new allowlisted email "everything in the shared graph" is a new categories-of-recipient disclosure under GDPR Art. 13/14 if any of those third parties are EU data subjects. **Redline (must-fix before this ships to any instance holding EU-resident personal data):** add one sentence to the Security hardening section stating the operator is responsible for (a) only enabling shared-graph mode with team members who have a legitimate need for full access, and (b) not relying on `NEOTOMA_APPROVED_EMAILS` as a substitute for per-recipient data-sharing agreements when the graph holds third-party PII. This is a documentation-only fix, not a code change, so it does not block the release engineering — but it should land before the tag is published given the escalation posture in `locale_profile` (GDPR is a named regime here).
- [NON-BLOCKING] credential-scope: The supplement's "Allowlist rotation is process, not code" callout is good practice — it correctly flags that removing an email from `NEOTOMA_APPROVED_EMAILS` doesn't invalidate a live session cookie. Recommend this note also state the *maximum* session lifetime (a concrete duration, e.g. "sessions expire after N hours/days") so operators can bound worst-case revocation lag instead of relying solely on a manual restart. If session lifetime is unbounded/very long, that materially changes the risk calculus for urgent revocation scenarios (e.g. an employee offboarding) and should be called out more prominently, not just as a process note.
- [x] Open-redirect / callback-hijack risk on the new OAuth callback is addressed: `normalizeOauthNextPath` constrains the post-signin redirect to an internal `/mcp/oauth/*` path, and nonce is single-use per `security_review.md` §1 and the "Suggested negative tests" section. No finding.

## PII leaving the store

- [x] No evidence in the reviewed diff of PII leaving the Neotoma store to a third-party service beyond the necessary Google JWKS/token verification round-trip (which is the standard OIDC flow, not a data export). `verifyGoogleIdToken` validates against Google's live JWKS — this sends the id_token (issued by Google, containing the signer's own email) back to Google for verification, not operator data to Google.
- [NON-BLOCKING] PII: Confirm the release notes / privacy policy (if one exists for hosted instances) are updated to disclose that signing in via Google causes the signer's Google-verified email to be stored and used as (or mapped to) the Neotoma `user_id` / allowlist match key. This is a new category of processing (OAuth identity data) introduced by #1924 and should be reflected in whatever privacy notice, if any, covers hosted Neotoma instances — not found in this PR's diff, so flagging as a gap rather than a confirmed omission.

## Escalation

No contract review triggered (this is a release/deployment PR, not a vendor agreement) — the `locale_profile.escalation_currency_threshold` ("contracts over EUR 10k escalate") doesn't apply here. Given GDPR is a named regulatory regime for this jurisdiction and this release adds a new personal-data processing surface (OAuth identity + shared-graph PII access), recommend a qualified-counsel or DPO sign-off on the shared-graph consent/disclosure redline above before this configuration is offered to any customer/team holding EU-resident contact data — not a blocker to merging code, but a blocker to enabling `NEOTOMA_SHARED_GRAPH_USER_ID` in production without that note.

## Verdict

**REQUEST_CHANGES** — one BLOCKING item (shared-graph consent/lawful-basis documentation gap). This is a docs-only redline, not a code change, and should be quick to resolve; the underlying security engineering (fail-closed OAuth, scoped fuzzy-match, no cross-tenant leakage) is sound as documented. Everything else here is NON-BLOCKING / advisory.

---
📎 Neotoma: [neotoma#2006](https://github.com/markmhendrickson/neotoma/pull/2006) · plan_contribution `ent_ade5ba66b88e3e49998d9b9f` (private)

neotoma-agent:
review:qa

**COMMENT**

QA lens only: test coverage adequacy (regression tests for fixed bugs, edge cases for new branches, contract tests for new endpoints). This PR is the release-bump commit itself (`package.json`/`package-lock.json` + three release-lane docs); the functional surface under review is the merged commit range `v0.18.8..origin/main` (#1924, #1982, #1970, #2002, #1926, #1925, #1978) that this release ships.

I independently verified the claims in the PR's own `docs/releases/in_progress/v0.19.0/test_coverage_review.md` and `security_review.md` gate artifacts against the actual test files and source in a worktree, rather than taking the prose at face value.

**Verified findings, by surface:**

1. **`query_contacts_at_company` (new MCP tool + `POST /query_contacts_at_company`)** — `tests/integration/company_entity_resolution_leads.test.ts` and `tests/services/company_resolution.test.ts` assert user-observable behavior end-to-end: fuzzy-match resolution, the read-only invariant ("never creates a company entity"), a false-positive guard against merging distinct companies, and tenant-isolation (throws on missing `userId` rather than scanning cross-tenant). Confirmed by reading the assertions, not just the filenames. Solid.

2. **Identifier resolution at scale (#1982, closes #1981)** — `tests/integration/entity_identifier_handler.test.ts` asserts `match_mode` values (`snapshot_field` / `direct` / `none`) across case-insensitive email resolution, type-declared identity fields, and the specific regression (exact match previously missed outside a capped 500-row scan window). Confirmed these assertions exist and target the actual regression, not a helper function.

3. **`works_at` retraction on organization change (#1970)** — `tests/integration/auto_link_retraction_organization_change.test.ts` covers both the MCP `store` path and REST `POST /store` path, both the success case and the `AUTO_LINK_RETRACTION_FAILED` failure path (with target ids) on each surface, and — the edge case that matters most for a data-mutating auto-link fix — asserts a **manually-created** `works_at` edge is preserved rather than retracted. This is exactly the edge-case bar I look for on mutating operations. Confirmed by reading the test bodies at lines 122, 192, 267, 350, 411.

4. **MCP unknown-session 404 (#1926)** — `tests/integration/mcp_session_404_reconnect.test.ts` walks the branch matrix (only `hadSessionHeader=true + unknown session + non-init` changes) and exercises the reconnect round-trip (404 → session-less initialize succeeds). Confirmed.

5. **Google OAuth sign-in (#1924)** — `tests/unit/google_oidc.test.ts` (24 tests) and `tests/unit/google_oidc_identity_resolution.test.ts` (2 tests) adversarially cover the verification logic: wrong issuer/audience, expired token, bad signature, `email_verified: false`, non-allowlisted email, nonce single-use, and fail-safe on malformed UUID for `getSharedGraphUserId()`. This is real adversarial unit coverage of an AAuth-adjacent signing/verification surface, one of my standing high-priority coverage domains.

**[NON-BLOCKING] test-coverage: no route-level test for the Google OAuth fail-closed behavior**

`GET /mcp/oauth/google/start` and `GET /mcp/oauth/google/callback` (`src/actions.ts:2698`, `2714`) are unauthenticated Express routes whose entire safety property — "a default install is unaffected" — rests on `isGoogleSigninEnabled()` returning `false` and the route returning 404. I confirmed by direct search that no integration test issues an HTTP request to either route and asserts a 404 when the env vars are unset; coverage exists only at the `isGoogleSigninEnabled()` unit level (`google_oidc.test.ts:111-133`), not at the route boundary. The PR's own `test_coverage_review.md` and `security_review.md` (Finding 2) already name this gap and track it for v0.19.1 — I'm confirming it's real, not taking it on faith. Recommend a route-level `supertest` assertion (`GET /mcp/oauth/google/start` with the two env vars unset → 404) alongside the openapi.yaml declaration follow-up, so the regression can't reappear silently if the fail-closed check is ever refactored.

**[NON-BLOCKING] test-coverage: gate-coverage blind spot on new unauth routes — cite `change_guardrails_rules.md` MUST #17**

I confirmed independently: `grep google openapi.yaml` returns nothing, `scripts/security/protected_routes_manifest.json` `overrides` is `[]`, and `sync_protected_routes_manifest.js` auto-derives the manifest from `openapi.yaml`. This means the two OAuth routes satisfy **neither** branch of the repo's standing guardrail — `docs/architecture/change_guardrails_rules.md` MUST #17: *"Every new Express route MUST land in `scripts/security/protected_routes_manifest.json` (auth-required) or in the runtime allow-list with a stated `reason`."* `npm run security:manifest:check` reports "in sync (116 routes)" and the auth-topology-matrix test passes, but neither actually exercises these two routes — they're invisible to the gate, not passing it. This is exactly the regression class the manifest-sync gate exists to catch (the v0.11.1 Inspector auth-bypass class cited in the same rule). Mitigating factors that keep this non-blocking: the routes fail closed by design, are covered by 24+2 adversarial unit tests on the underlying verification logic, and the gap is transparently surfaced in both gate artifacts and the PR description itself rather than hidden. Recommend landing the `openapi.yaml` declaration (or explicit `overrides` entry with `requires_auth: false` + reason) in the v0.19.1 follow-up named in the PR, not deferred indefinitely.

**No BLOCKING findings.** Every user-observable behavior change in this release range ships with tests that assert observable effects (not helper-level smoke tests), the two data-mutating/session-lifecycle changes (#1970, #1926) carry both success- and failure-path coverage across both MCP and REST surfaces, and the one genuinely new authentication surface (#1924) has adversarial unit coverage of its verification logic. The two gaps above are advisory, already self-identified by the PR authors with matching detail in `test_coverage_review.md` / `security_review.md`, and explicitly tracked for v0.19.1 rather than silently dropped.

**QA gate:** signing off. No functional change in *this* PR lacks eval/test coverage; the two advisory gaps are pre-existing (inherited from #1924, already merged to main) and tracked, not new regressions introduced by this release-bump commit.

📎 Neotoma: not referencing any private Neotoma entity records for this review — all findings sourced directly from the repo's own gate artifacts (`docs/releases/in_progress/v0.19.0/test_coverage_review.md`, `security_review.md`) and independent verification against `src/actions.ts`, `openapi.yaml`, and the cited test files in this PR's branch.

neotoma-agent:

**REQUEST_CHANGES**

Aggregated review panel for #2006 (chore(release): v0.19.0). Parent issue: #1981.

**Note on merge state:** this PR merged (commit `73a50b27`, 2026-07-24T13:13:30Z) before this aggregation ran. The verdict below reflects the panel as convened; the two BLOCKING items are now **required follow-up**, not a pre-merge block — I am not reopening or reverting the merge. Per standing autonomy guardrail, any further merge-adjacent action stays operator-gated.

## Per-lens roll-up

| Lens | Panelist | Verdict | Blocking |
|---|---|---|---|
| pm | Pavo | APPROVE | 0 |
| ux | Accipiter | REQUEST_CHANGES | 1 |
| legal | Buteo | REQUEST_CHANGES | 1 |
| qa | Phoenicurus | COMMENT | 0 |

## Blocking findings

- [ ] **[BLOCKING] docs-gap (ux, Accipiter):** New Google OAuth env vars (`NEOTOMA_GOOGLE_CLIENT_ID`, `NEOTOMA_GOOGLE_CLIENT_SECRET`, `NEOTOMA_APPROVED_EMAILS`, `NEOTOMA_SHARED_GRAPH_USER_ID`) exist only in the point-in-time release supplement (`docs/releases/in_progress/v0.19.0/github_release_supplement.md`), not in persistent developer docs (`docs/developer/cli_reference.md`, `docs/infrastructure/client_instance_deployment.md`). A future operator enabling Google sign-in has no durable setup reference and may miss the `SHARED_GRAPH_USER_ID` blast-radius warning. Fix: promote the env var table + "Operator action" callouts into `docs/infrastructure/client_instance_deployment.md` (or a linked dedicated OAuth setup doc).
- [ ] **[BLOCKING] credential-scope (legal, Buteo):** Shared-graph mode (`NEOTOMA_SHARED_GRAPH_USER_ID`) grants every allowlisted email full read/write on a graph containing third-party PII, documented only with operator-risk framing — no GDPR consent/lawful-basis disclosure. Redline: add a one-sentence lawful-basis/consent disclosure to the Security hardening section. Docs-only fix.

## Non-blocking (tracked, not gating)

- ux: `query_contacts_at_company` owner_user_id mismatch has no stable error code / `openapi.yaml` 400 response entry (pre-existing pattern, not introduced here).
- ux: tool description/read-only contract on `query_contacts_at_company` called out as a positive baseline for future new-tool PRs.
- legal: confirm Google OAuth library licence is MIT-compatible (added in underlying #1924, outside this diff).
- legal: confirm lawful basis for returning contact PII via `query_contacts_at_company`.
- legal: state a concrete max session lifetime for the revocation-lag note.
- legal: confirm hosted-instance privacy notice discloses Google-identity processing.
- qa (self-verified, not accepted on faith): no route-level 404 test for OAuth-disabled state (unit-level only); OAuth routes absent from `openapi.yaml` — invisible to `security:manifest:check`/auth-matrix despite green gates. Both already self-flagged in the PR body, tracked for v0.19.1.

## Gate disposition

This PR's own diff (`package.json`, `package-lock.json`, 3 release-lane docs) is a release-aggregation chore; the gated fix for parent issue #1981 is PR #1982, whose `pr_review` gate on issue #1981 (`ent_48dee46fb2b3a50d2a134501`) already carries its own history and is independently `changes_requested` from prior review rounds unrelated to #2006's content — I am not overwriting that record with #2006-specific findings, since they concern a different artifact (the release wrapper, not the identifier-lookup fix). The two BLOCKING items above are release-scoped documentation/consent gaps in the *shipped* v0.19.0 surface (Google OAuth, shared-graph) and are filed as required follow-up for the v0.19.1 lane referenced in the PR body's own "Reviewer attention" section.

**Verdict: REQUEST_CHANGES** — 2 BLOCKING items, both docs-only, both already merged into main via #2006. No blocker requires reverting; both are follow-up-doc fixes.

📎 Neotoma: [issue#1981](https://neotoma.markmhendrickson.com/entities/ent_48dee46fb2b3a50d2a134501)

PR #2016, opened 2026-07-23
fix(issues): forward reporter provenance fields on submit_issue remote leg

### LINKED ISSUE #2014: submit_issue drops reporter provenance on the remote leg — remote rejects with ERR_REPORTER_ENVIRONMENT_REQUIRED despite fields being supplied
## Summary

`submit_issue` drops `reporter_git_sha` / `reporter_app_version` (and `reporter_git_ref` / `reporter_channel` / `reporter_ci_run_id`) when forwarding to a remote instance. The remote then rejects the submission with `ERR_REPORTER_ENVIRONMENT_REQUIRED` — **complaining that fields the caller supplied are missing.**

Net effect: a correctly-formed public `submit_issue` call cannot reach a remote instance. It stores locally with `sync_pending=true` and the operator sees a confusing error blaming their own input.

## Reproduction

Call `submit_issue` with reporter provenance supplied:

```jsonc
{
  "title": "…",
  "body": "…",
  "visibility": "public",
  "target_repo": "markmhendrickson/neotoma",
  "reporter_app_version": "0.18.8",
  "reporter_git_sha": "fdbd27f"
}
```

Observed response:

```
Remote submission to https://neotoma.markmhendrickson.com failed:
Remote Neotoma issue submit failed: {
  "error_code": "ERR_REPORTER_ENVIRONMENT_REQUIRED",
  "message": "submit_issue requires reporter environment: provide at least one of
              `reporter_git_sha` or `reporter_app_version`."
}
```

The local entity is stored correctly **with both fields populated** (`reporter_git_sha: "fdbd27f"`, `reporter_app_version: "0.18.8"`), which confirms the caller supplied them and validation passed locally. Only the remote leg loses them.

## Root cause

`src/services/issues/neotoma_client.ts` builds the outbound payload as:

```ts
const guestBody: Record<string, unknown> = {
  title: params.title,
  body: params.body,
  labels: params.labels ?? [],
  visibility: params.visibility,
  submission_timestamp: now,
};
if (params.githubUrl?.trim()) guestBody.github_url = params.githubUrl.trim();
if (ghNum > 0)                guestBody.github_number = ghNum;
if (authorTrim)               guestBody.author = authorTrim;
if (localId)                  guestBody.local_issue_id = localId;
```

No reporter field is ever added to `guestBody`. The remote `/issues/submit` handler enforces `ERR_REPORTER_ENVIRONMENT_REQUIRED`, so every remote submission fails regardless of caller input. The unsigned-guest retry path reuses the same `guestBody`, so it fails identically.

Contrast `add_issue_message`, which *does* forward them (`src/actions.ts:9944–9955` spreads `reporter_git_sha`, `reporter_git_ref`, `reporter_channel`, `reporter_app_version` when present). The submit path simply wasn't given the same treatment.

## Impact

- Remote issue submission is broken for all callers, not just missing-provenance ones.
- The error message actively misleads: it tells the operator to supply fields they already supplied, so the natural next step (adding them) doesn't help.
- Provenance is silently lost on the one path where it's mandatory, undermining the `Reporter provenance` contract in `docs/subsystems/issues.md`.

## Suggested fix

Forward the reporter fields in `neotoma_client.ts` alongside the existing optional keys, matching the `add_issue_message` pattern:

```ts
if (params.reporterGitSha?.trim())     guestBody.reporter_git_sha = params.reporterGitSha.trim();
if (params.reporterAppVersion?.trim()) guestBody.reporter_app_version = params.reporterAppVersion.trim();
if (params.reporterGitRef?.trim())     guestBody.reporter_git_ref = params.reporterGitRef.trim();
if (params.reporterChannel?.trim())    guestBody.reporter_channel = params.reporterChannel.trim();
if (params.reporterCiRunId?.trim())    guestBody.reporter_ci_run_id = params.reporterCiRunId.trim();
```

This requires threading the fields through the client's `params` type if they aren't already present.

A regression test asserting that a `submit_issue` call carrying `reporter_git_sha` produces an outbound payload containing it would prevent recurrence — the failure is invisible locally, since the local store keeps the fields.

## Discovered while

Filing #2010. The issue had to be created manually with `gh issue create` because this bug blocked the normal path.

## Swarm specification

### Product / Scope (PM)

**Problem.** `submit_issue`'s remote leg (`src/services/issues/neotoma_client.ts`) silently drops all five reporter-provenance fields (`reporter_git_sha`, `reporter_app_version`, `reporter_git_ref`, `reporter_channel`, `reporter_ci_run_id`) when building the outbound `guestBody` payload. The remote instance then rejects with `ERR_REPORTER_ENVIRONMENT_REQUIRED`, an error that misattributes the failure to the caller's input rather than the client's own omission. Every remote `submit_issue` call is broken today, regardless of what the caller supplies. `add_issue_message` already forwards the same fields correctly (`src/actions.ts:9944–9955`), so the fix is to bring `submit_issue`'s remote path to parity with that existing, working pattern — not to design anything new.

**In scope**
- Forward `reporter_git_sha`, `reporter_app_version`, `reporter_git_ref`, `reporter_channel`, `reporter_ci_run_id` into `guestBody` in `neotoma_client.ts`, matching the trim/conditional-inclusion style already used for `github_url`, `author`, etc.
- Thread these fields through the client's `params` type if not already present, so they reach `neotoma_client.ts` from the `submit_issue` action.
- Apply the same fix to the unsigned-guest retry path (confirmed in the issue to reuse `guestBody` and fail identically).
- Regression test: a `submit_issue` call carrying `reporter_git_sha` (and/or `reporter_app_version`) produces an outbound remote payload that contains it — asserting on the constructed request body, not just local storage (local storage already works and would mask this regression again).

**Out of scope**
- Any change to the remote `/issues/submit` handler's validation logic (`ERR_REPORTER_ENVIRONMENT_REQUIRED` itself is correct behavior once fields are actually sent).
- Broader error-message quality improvements on the remote side (e.g., making the error distinguish "not sent" vs "not supplied") — worth a follow-up issue, not this fix.
- Any schema, MCP tool/CLI surface, or entity-type change — this is a payload-construction bug on an existing, already-contracted field set, not a new interface.
- Sync/backfill of previously-failed submissions currently sitting with `sync_pending=true` (e.g. #2010 itself) — that's an operational retry, not part of this code fix's scope.

**Acceptance criteria**
- [ ] `neotoma_client.ts` forwards all five reporter fields (`reporter_git_sha`, `reporter_app_version`, `reporter_git_ref`, `reporter_channel`, `reporter_ci_run_id`) into the outbound `guestBody` when present, trimmed, matching the existing optional-field pattern.
- [ ] Fix applies to both the primary remote-submit path and the unsigned-guest retry path.
- [ ] Effect-verified test (per `fixed_means_behavior_verified_not_contract_accepted`): a test asserts the **actual reported effect** — that a `submit_issue` call with `reporter_git_sha`/`reporter_app_version` supplied results in a remote request that no longer triggers `ERR_REPORTER_ENVIRONMENT_REQUIRED`, and that the constructed outbound payload contains the supplied fields. A test that only checks local storage or input validation does not satisfy this — the bug is invisible locally by the reporter's own observation.
- [ ] Cross-surface parity (per `cross_surface_contract_parity_tested_all_surfaces`): confirm whether `submit_issue` is exposed on more than one surface (MCP tool and/or CLI command). If so, add a test on each exposing surface, each driven with that surface's natural call shape, not just the shared client function in isolation.
- [ ] A previously-failing reproduction (remote submission with reporter fields supplied) now succeeds end-to-end against a representative remote target, or an equivalent integration/contract test simulates the remote handler's validation to prove the fields arrive.
- [ ] No change to the remote validation contract itself, MCP tool schema, or CLI flags — confirmed as a payload-construction-only fix.

**Priority & sequencing.** This is a P0 correctness bug: it fully blocks the primary path for remote public issue submission — the swarm's own issue-filing pipeline (see #2010, filed manually via `gh issue create` because this bug blocked the normal path). It unblocks: (a) all future remote `submit_issue` calls including this swarm's own dogfooding loop, and (b) manual resync of any issues currently stuck with `sync_pending=true`. No other in-flight work depends on this being sequenced after anything else — recommend immediate fix.

**Classification note (interface-surface predicate):** Reviewed against the bug/security fast-path override. This change does not add or modify a `SchemaDefinition` field, MCP tool, CLI command, error code, or relationship type — it restores forwarding of fields already part of the existing, remote-enforced contract, using a pattern (`add_issue_message`) already shipped elsewhere in the codebase. Classified as a **localized implementation bug** — the `label:bug` fast path to `cicada` applies; arch gate is not required.

### Design / UX (Accipiter)

**User-facing surface.** No new surface — `submit_issue` (MCP tool + any CLI wrapper) keeps its existing signature. This is a truth-in-error-messaging fix: what changes is only whether the client honors input the caller already provides, and what the caller sees when it still can't.

**Interaction / flow — current (broken) vs. fixed**

| Step | Current | Fixed |
|---|---|---|
| 1. Caller calls `submit_issue` with `reporter_git_sha`/`reporter_app_version` set | Fields accepted, validated locally, stored locally | unchanged |
| 2. Client builds remote `guestBody` | Reporter fields silently omitted | Reporter fields forwarded (trimmed, present-only) |
| 3. Remote validates | Rejects: `ERR_REPORTER_ENVIRONMENT_REQUIRED` — **caller is blamed for its own client's bug** | Passes; submission proceeds normally |
| 4. Caller sees | Confusing error that contradicts caller's own input; local copy sits at `sync_pending: true` indefinitely with no signal that it's a client bug, not a caller bug | Success response; no residual `sync_pending` state for this cause |

**Friction identified (why this is worse than an ordinary bug).** The caller did the right thing — supplied the fields — and is punished with an error that tells them to do the thing they already did. There is no way for the caller to self-diagnose from the error text alone that the fields were dropped *in transit*, not omitted *by them*. This is a trust-breaking failure mode: the tool's error message actively misleads rather than merely being unhelpful. `sync_pending: true` on the local entity is the only signal something's wrong, and nothing surfaces it proactively.

**Error message — this fix must not ship without addressing the message, even though PM scoped the remote handler's logic as out-of-scope.** The code fix (forwarding fields) makes the error disappear for correctly-formed calls, which resolves the acute case. But two residual UX risks remain and should be covered by this change:

1. **If the client-side fields genuinely are empty/whitespace-only** (a real caller mistake), the existing remote error text is still the one surfaced today: *"submit_issue requires reporter environment: provide at least one of `reporter_git_sha` or `reporter_app_version`."* This is acceptable as a genuine-mistake message but should additionally hint at the two next-nearest fields (`reporter_git_ref`, `reporter_channel`) if those are also acceptable substitutes at the remote validation layer — confirm with Eng whether the remote's `ERR_REPORTER_ENVIRONMENT_REQUIRED` condition accepts any of the five fields or only the two named ones; the error text should list exactly the set that satisfies validation, not a subset, so a caller who supplies `reporter_git_ref` alone isn't told they need a field they don't.
2. **Client-side pre-flight hint (recommended, low-cost addition to this fix):** before making the remote call, if the resolved `guestBody` about to be sent has none of the five reporter fields set, the client should surface a distinct, local, actionable message rather than relying on the remote round-trip — e.g. `submit_issue: no reporter provenance fields resolved to send (reporter_git_sha, reporter_app_version, reporter_git_ref, reporter_channel, reporter_ci_run_id all empty). Remote will reject with ERR_REPORTER_ENVIRONMENT_REQUIRED — check that these params were passed and are non-empty strings.` This turns a remote round-trip failure into an immediate, correctly-attributed local one, and named exactly the fields the client checked — closing the "client says one thing, remote enforces another" gap for good, not just for this bug instance.

**Discoverability & naming.** No naming changes — field names (`reporter_git_sha`, `reporter_app_version`, `reporter_git_ref`, `reporter_channel`, `reporter_ci_run_id`) are already established via `add_issue_message` and stay identical here for consistency across the two call surfaces. Do not introduce a differently-named parallel field set on the submit path.

**Error & empty states**
- **Empty state (no reporter fields supplied anywhere):** genuine-mistake path — remote error stands, but see hint-set accuracy note above.
- **Silent-drop state (this bug):** eliminated by the fix; no longer reachable once fields are forwarded.
- **Partial-forwarding edge case:** if only some of the five fields resolve to non-empty (e.g. `reporter_git_ref` set but `reporter_git_sha`/`reporter_app_version` empty), confirm the remote's validation is satisfied by any one of the five, not just the two named in today's error text — otherwise callers relying on `reporter_git_ref` alone will still hit a misleading rejection after this fix ships. Flag to Eng/QA as a specific case to test, not assume.
- **Stuck `sync_pending` state (pre-existing, adjacent):** out of scope for auto-resync per PM, but the local entity's `sync_pending: true` plus whatever surfaces it (status tool/CLI output) should — once this fix ships — allow a manual resubmit to succeed. No new UI needed; confirm `get_entity_submission_status` / equivalent already reflects `sync_pending` clearing on retry (Eng/QA to confirm existing behavior still applies).

**Accessibility.** N/A — no UI component; this is a CLI/API/tool-call surface. Error text should remain plain, single-line-parseable text (no color-only signaling) since it may render in terminals, logs, or non-interactive CI contexts (`reporter_ci_run_id` implies CI callers are a real audience).

**Docs/examples this change must ship**
- [ ] Update `docs/subsystems/issues.md`'s "Reporter provenance" contract section to state explicitly that all five fields are forwarded end-to-end on both `submit_issue` and `add_issue_message`, and are not silently dropped on the remote leg — this doc currently states the contract but didn't catch that `submit_issue` violated it, so it should say this is now tested (points at the regression test from Eng).
- [ ] Add or update the `submit_issue` example in docs/tool reference to show a fully-populated reporter-provenance call (mirroring the issue's own repro payload), so future readers see the intended shape, not just field names in a schema table.
- [ ] If the client gains the pre-flight hint (recommended item above), document that message verbatim in the troubleshooting/error-reference doc alongside `ERR_REPORTER_ENVIRONMENT_REQUIRED`, so a caller hitting either message can find both in one lookup.

- [ ] Forwarding fix verified to make the exact repro payload in the issue succeed end-to-end (no error), from the caller's point of view — not just unit-tested at the payload-construction layer.
- [ ] Remote error text's field list is confirmed accurate against actual remote validation logic (which of the five fields actually satisfy `ERR_REPORTER_ENVIRONMENT_REQUIRED`) — corrected if the message names only 2 of 5 accepted fields.
- [ ] (Recommended, non-blocking) Client-side pre-flight message added for the genuine-empty case, named and worded per above.
- [ ] `docs/subsystems/issues.md` and the `submit_issue` reference example updated per the docs checklist above.
- [ ] No new parameter names, flags, or schema fields introduced — parity with `add_issue_message`'s existing field names confirmed.

**Verdict:** COMMENT — no blocking UX concerns on the PM-scoped fix itself; the two flagged items (error-text field-list accuracy, optional pre-flight hint) are recommended hardening, not gates on this fix shipping.

### Engineering

**Root cause confirmed.** `src/services/issues/neotoma_client.ts` — `submitIssueToRemote()` (lines 67–120): the `params` type (67–79) has no `reporterGitSha`/`reporterAppVersion`/`reporterGitRef`/`reporterChannel`/`reporterCiRunId` fields, and `guestBody` (102–120) never sets them. The unsigned-guest retry (150–152) reuses the same `guestBody` object, so both paths are fixed by one change to the body construction. `add_issue_message`'s forwarding pattern (`src/actions.ts:9944–9955`) is the reference implementation to mirror.

### Files touched

1. **`src/services/issues/neotoma_client.ts`**
   - `submitIssueToRemote(params)` — add 5 optional string fields to the `params` type (67–79): `reporterGitSha?`, `reporterAppVersion?`, `reporterGitRef?`, `reporterChannel?`, `reporterCiRunId?`.
   - In the `guestBody` construction block (109–120), add 5 trimmed conditional assignments matching the existing `github_url`/`author` pattern:
     ```
     if (params.reporterGitSha?.trim())     guestBody.reporter_git_sha = params.reporterGitSha.trim();
     if (params.reporterAppVersion?.trim()) guestBody.reporter_app_version = params.reporterAppVersion.trim();
     if (params.reporterGitRef?.trim())     guestBody.reporter_git_ref = params.reporterGitRef.trim();
     if (params.reporterChannel?.trim())    guestBody.reporter_channel = params.reporterChannel.trim();
     if (params.reporterCiRunId?.trim())    guestBody.reporter_ci_run_id = params.reporterCiRunId.trim();
     ```
   - No new function, no new code path — the retry at line 150 already reuses `guestBody` by reference, so it inherits the fix automatically. Do not duplicate the field-assignment block for the retry.
   - (UX-flagged, recommended non-blocking addition) Immediately before the `client.POST` call at 132, if none of the five `guestBody.reporter_*` keys were set, `console.warn` (or the repo's structured logger, gated per `docs/observability/logging.md` — no PII, so field names only, not values) a local diagnostic naming exactly the five checked field names, so the failure is attributable client-side before the remote round-trip. Implement only if it fits cleanly without expanding scope; PM/UX have marked this non-blocking.

2. **Caller site — thread reporter params through to `submitIssueToRemote`.** Locate the `submit_issue` action handler (the caller of `submitIssueToRemote`, expected in `src/actions.ts` alongside `add_issue_message`'s handler at ~9944–9955) and confirm it already receives `reporter_git_sha` / `reporter_app_version` / `reporter_git_ref` / `reporter_channel` / `reporter_ci_run_id` from the incoming request (it must — local storage already persists them per the issue's reproduction). Pass those same values into the `submitIssueToRemote(params)` call using the new camelCase param names. This is a threading fix only — no new parsing, no new validation, no schema change; the fields already exist on the inbound `submit_issue` contract.

3. **MCP tool / CLI surface parity check (Eng verification step, not a code change).** Confirm whether `submit_issue` is invoked identically from the MCP tool handler and any CLI command (`neotoma issues submit` or similar) — i.e. do both surfaces funnel through the same `src/actions.ts` handler and the same `submitIssueToRemote()` call, or does the CLI have a separate remote-submission code path? If both surfaces share the one handler (expected, per existing `add_issue_message` precedent), one fix covers both and cross-surface parity is structural, not a second implementation. If the CLI has its own remote-submit path, apply the identical fix there and add the CLI as a second test surface. State this finding explicitly in the PR description either way.

### Data / contract changes

None. All five fields are already part of the existing `submit_issue` request contract (locally validated and stored today, per the issue's own reproduction). This PR only fixes internal payload construction between the action handler and the remote HTTP client — no `openapi.yaml` change, no new MCP tool field, no new CLI flag, no schema migration. Per `docs/architecture/change_guardrails_rules.md`, the Touchpoint Matrix's "API & contract surfaces" row does not apply here since no request/response field is being added or changed at the boundary — only forwarded internally. No `hint`/tightening obligations apply (this loosens an existing accidental drop, it doesn't tighten validation).

### Layering

Change is confined to the transport-adapter layer (`neotoma_client.ts`) plus the thin parameter-threading in the action handler that calls it. No reducer, schema, ingestion, or relationship-layer code is touched. Consistent with `docs/architecture/layered_architecture.md` — this is Operational-Layer-to-State-Layer transport plumbing, not a State Layer behavior change.

### Build steps (in order)

1. Add the 5 optional fields to `submitIssueToRemote`'s `params` type in `neotoma_client.ts`.
2. Add the 5 trimmed conditional `guestBody` assignments, placed with the existing optional-field block (after `local_issue_id`, matching style).
3. Confirm (do not modify unless missing) that the retry block at 150–152 reuses `guestBody` by reference — no separate fix needed there; add a one-line comment only if the reuse isn't already obvious from context, to prevent future duplication drift.
4. Update the caller (action handler) to pass the 5 reporter fields from the incoming `submit_issue` request into `submitIssueToRemote(params)`.
5. Run the MCP/CLI parity check (item 3 above) and document the finding in the PR body.
6. Write the regression test (see QA handoff below — Eng implements per QA's test plan once QA section lands; if QA has not yet specified exact assertions, implement a minimum viable version now and let QA's section supersede/extend it): assert that a `submit_issue` call with `reporter_git_sha` set produces an outbound `guestBody`/request payload containing `reporter_git_sha`, by inspecting the constructed request body (mock/spy the HTTP client at the `client.POST` call boundary, not just local storage).
7. Run `npm run type-check`, `npm test` (targeted to `tests/**/issues/**` and `tests/**/neotoma_client**` if such paths exist, else full unit suite), and `npm run lint` per `.claude/rules/agent_test_execution.md`.
8. If a new/moved test file is added, run `npm run generate:test-catalog` and `npm run validate:test-catalog` before opening the PR, per the impl sign-off checklist's regenerated-files requirement.
9. Do not touch `openapi.yaml`, `contract_mappings.ts`, MCP tool schema, or CLI flag definitions — confirm this in the PR description as the "no contract change" declaration.

### Risks / edge cases for QA to weight

- Partial-field case (Design/UX flagged): a call with only `reporter_git_ref` set (no `reporter_git_sha`/`reporter_app_version`) must now forward `reporter_git_ref` and should be tested against the remote's actual accepted-field set — Eng cannot change remote validation, but the regression test should assert forwarding for all 5 fields individually, not just the two named in today's error text.
- Whitespace-only values (`"   "`) must resolve to "not sent," consistent with the `.trim()` guard — mirrors existing `github_url`/`author` handling, no new behavior needed, but worth a test case since it's the exact class of bug being fixed (input present but not forwarded).
- The retry path (unsigned-guest fallback) must be covered by at least one test asserting reporter fields still forward after an `AUTH_REQUIRED` retry, since it's a separate code path reusing the same object — reference-reuse is easy to accidentally break in future refactors.

### QA / Test Plan

**Deliverable classification.** Functional change — payload-construction bug on an existing, remote-enforced contract. Requires an agentic/regression eval per `fixed_means_behavior_verified_not_contract_accepted`. No new MCP tool/schema/CLI surface, so no contract test for a *new* endpoint — but the existing `submit_issue` remote contract IS under test here, driven at the payload-construction boundary the bug lives in.

**Primary regression eval (blocking).**

- **Location:** unit/integration test colocated with `neotoma_client.ts` (e.g. `tests/**/neotoma_client.submit_issue.test.ts` or matching existing suite naming) — plus, if `submit_issue` also has an `agentic_eval` fixture under `tests/fixtures/agentic_eval/`, a fixture update so the Tier‑1 matrix exercises it end-to-end.
- **Mechanism:** mock/spy the HTTP client at the `client.POST` call boundary in `submitIssueToRemote` (per Eng's build step 6) — assert on the **actual constructed outbound request body**, not on local storage. Local storage already persists these fields correctly today; a test that only checks local state would pass on the unfixed code and is explicitly disallowed by the issue's own reproduction.
- **Assertion (must-pass, currently RED on `main`):** given a `submit_issue` call carrying `reporter_git_sha: "fdbd27f"` and `reporter_app_version: "0.18.8"` (the exact repro payload from the issue), the outbound `guestBody` sent to the remote contains both `reporter_git_sha` and `reporter_app_version` with those exact values.

**Edge cases (each its own assertion, not folded into the primary case):**

1. **All five fields individually.** Parametrized/table test: for each of `reporter_git_sha`, `reporter_app_version`, `reporter_git_ref`, `reporter_channel`, `reporter_ci_run_id` supplied alone, assert that field (and only that field, among the five) appears in `guestBody`. Prevents a partial fix that only handles the two fields named in today's error text (Design/UX explicitly flagged this).
2. **Whitespace-only value resolves to "not sent."** `reporter_git_sha: "   "` → `guestBody.reporter_git_sha` is absent, matching the existing `.trim()` guard behavior for `github_url`/`author`. This is the exact bug class (present-but-dropped) so must be asserted, not assumed.
3. **All five fields omitted (empty state).** `guestBody` contains none of the five reporter keys — confirms the fix doesn't force-send empty/default values, preserving the remote's `ERR_REPORTER_ENVIRONMENT_REQUIRED` as a genuine-mistake path.
4. **Retry path (unsigned-guest fallback) preserves reporter fields.** Force the primary submit to hit the `AUTH_REQUIRED` branch (per existing retry-path test scaffolding, if any) and assert the retried request also contains the supplied reporter fields — proves the reference-reuse of `guestBody` Eng is relying on actually holds, and pins it against future refactors that might clone/rebuild the body per attempt.
5. **Mixed partial set.** `reporter_git_ref` set, `reporter_git_sha`/`reporter_app_version` unset — assert `reporter_git_ref` forwards. Covers the Design/UX-flagged case of a caller relying on a field outside the two named in the current error text.

**Cross-surface parity test (per `cross_surface_contract_parity_tested_all_surfaces`):**

- Gate on Eng's parity-check finding (build step 5 / files-touched item 3): if MCP tool handler and CLI command both funnel through the same `src/actions.ts` handler → one shared test at that handler layer suffices structurally, but still add **one** test per exposing surface (MCP tool call shape, CLI invocation shape) asserting the reporter fields survive from that surface's natural entry point through to `submitIssueToRemote`'s params — not just a shared-function-in-isolation test. If Eng's finding instead shows a separate CLI remote-submit path, that path needs its own full edge-case set (1–5 above), not just a smoke test.
- Explicitly record in the PR/QA report which surfaces were found and tested (matches Eng's obligation to state the finding in the PR description).

**End-to-end / contract-level confirmation (blocking, satisfies "previously-failing reproduction now succeeds"):**

- Either: run the issue's exact repro payload against a representative remote target (staging or a local instance of the remote `/issues/submit` handler) and confirm no `ERR_REPORTER_ENVIRONMENT_REQUIRED` — OR, if no live remote target is available in CI, an integration test that simulates the remote handler's validation logic (asserting the constructed payload would satisfy `ERR_REPORTER_ENVIRONMENT_REQUIRED`'s accepted-field check) counts as the equivalent per Eng's plan. State explicitly which of the two was used.
- If a live-remote check is run, also confirm the previously `sync_pending: true` local entity transitions to synced on retry (Design/UX flagged this as an existing-behavior confirmation, not new work) — one assertion, not a new feature test.

**Non-blocking / recommended (do not gate sign-off on these, but flag if skipped):**

- If Eng implements the client-side pre-flight warning (non-blocking per Eng's plan), add one test asserting the warning fires when all five fields resolve empty and does not fire when at least one resolves — keeps the hardening covered without expanding it into a blocking requirement.
- If Design/UX's error-text field-list accuracy fix ships in this PR (it's a remote-side, out-of-scope-per-PM change), no client-side test is needed; note in the QA report that this remains an open remote-side item, not silently dropped.

**Definition-of-done checklist:**

- [ ] Primary regression test (repro payload → both `reporter_git_sha`/`reporter_app_version` present in outbound `guestBody`) exists, is RED against pre-fix code, GREEN against fixed code.
- [ ] All 5 fields tested individually for forwarding (edge case 1).
- [ ] Whitespace-only trim behavior tested (edge case 2).
- [ ] Empty-state (no fields) tested — no forced defaults introduced (edge case 3).
- [ ] Retry-path field preservation tested (edge case 4).
- [ ] Mixed-partial-set case tested, at minimum `reporter_git_ref`-only (edge case 5).
- [ ] Cross-surface parity finding documented in PR; each exposing surface (MCP tool, CLI if separate) has its own test.
- [ ] End-to-end confirmation done (live remote or simulated-handler integration test) — method used stated explicitly.
- [ ] `npm run eval:tier1` green if an `agentic_eval` fixture exists/was updated for `submit_issue`; no unrelated snapshot force-updates.
- [ ] No test asserts only local storage as proof of fix — every assertion targets the constructed outbound request body or the remote/simulated-remote response.
- [ ] `docs/subsystems/issues.md` regression-test reference (per Design/UX docs checklist) points at the actual committed test file/id.

**Verdict:** Gate `qa` remains `pending` until the primary regression eval + edge cases 1–4 (minimum bar) are committed and green in CI. Edge case 5 and the parity test are required before merge but may land in the same PR as Eng's implementation — QA does not require a separate PR. Cicada implements per this plan; Phoenicurus re-verifies the committed eval is red-before/green-after and signs off once confirmed.

### Security / Arch (Waxwing)

**Classification confirmed.** Concur with PM's interface-surface predicate: no `SchemaDefinition`, MCP tool, CLI flag, error code, or relationship type is added or changed. This is internal payload-construction forwarding on an already-contracted field set. Arch gate is **waived** — bug fast-path to `cicada` stands. No ADR required for the core fix. The checklist below is scoped narrowly to what a security/arch reviewer must still confirm on a fix of this shape, per the interface gate's own rule that "bug"-labeled changes touching cross-cutting concerns still get a lightweight check.

**Contract-first ordering.**
- [ ] Confirm no `openapi.yaml` change is required or made — the five reporter fields are already declared on the inbound `submit_issue` contract (Eng's build step 9 asserts this; verify it holds at PR time, not just at spec time).
- [ ] If Design/UX's error-text field-list fix (remote-side) ships in a *future* PR, that PR requires contract-first treatment (`openapi_contract_flow.md`) since it touches the remote handler's response shape — flag this now so it isn't silently bundled into a "bug" PR later without going through `contract_mappings.ts`.
- [ ] This PR touches zero `contract_mappings.ts` rows. If the diff touches that file, treat it as a scope violation and re-open the arch gate.

**Tenant isolation on every entity lookup.**
- Not applicable to the code path under fix: `submitIssueToRemote()` performs an outbound HTTP POST construction, not an entity lookup — there is no `retrieve_entity`/`retrieve_entities` call in this diff to scope-check.
- [ ] Confirm the fix does not introduce a new read path (e.g. resolving `local_issue_id` to a fuller local entity to pull reporter fields from). If Eng's threading (files-touched item 2) reads any local entity to source the reporter values rather than passing through the already-validated inbound request params, that read MUST carry the same tenant/session scoping as every other entity lookup in `src/actions.ts` — verify it reuses the existing authenticated-session context rather than an unscoped lookup. Preferred implementation avoids this entirely: pass the inbound request's already-validated `reporter_*` values straight through, no intermediate entity fetch.

**Idempotency on mutating ops.**
- `submit_issue`'s remote POST is the mutating operation in scope. Confirm (do not assume) that the existing outbound request already carries whatever idempotency mechanism `/issues/submit` expects (e.g. `local_issue_id` as a dedupe key, or an explicit `idempotency_key` field) — this fix must not change idempotency behavior, only add fields to the same body.
- [ ] If `guestBody` has no explicit `idempotency_key` today, that's a pre-existing gap, not something this PR should silently paper over or silently leave — state its presence/absence explicitly in the PR description. Do not add one as part of this fix (scope creep per PM); file a follow-up issue if absent, since retries (the unsigned-guest retry path itself) without an idempotency key risk duplicate remote issue creation on transient auth failures.
- [ ] Confirm the retry path (line ~150) does not double-submit: since it reuses `guestBody` by reference, verify it replaces rather than repeats the original failed request (i.e., the first attempt genuinely failed with `AUTH_REQUIRED` before retry fires, not a fire-and-forget duplicate).

**Auth / credential-exposure constraints.**
- [ ] `reporter_git_sha`, `reporter_git_ref`, `reporter_ci_run_id`, `reporter_channel`, `reporter_app_version` are build/environment metadata, not credentials — confirm none of the five fields can carry secret material in practice (e.g. a CI system that embeds a token into `reporter_ci_run_id` by misconfiguration is a caller-side hygiene issue, not something this client should sanitize/redact). No new redaction logic needed; note this explicitly so a future reviewer doesn't assume PII/secret scrubbing was considered and skipped silently.
- [ ] If Eng implements the client-side pre-flight diagnostic (`console.warn`/structured logger), confirm it logs only the **names** of unset fields, never attempts to log field values (values could theoretically contain a misconfigured secret from a CI environment) — this is already specified in Eng's section; treat any deviation (logging actual field values) as a blocking finding.
- [ ] No auth-token, session-credential, or AAuth-identity field is touched by this diff — confirm the five added fields sit alongside the existing non-auth optional fields (`github_url`, `author`, `local_issue_id`) in `guestBody`, not commingled with any auth-header construction path.
- [ ] Unsigned-guest retry path: confirm this fix doesn't change *when* the client falls back to unsigned-guest submission, only that reporter fields survive the existing fallback. No change to auth-tier selection logic.

**Cross-surface parity (security lens).** Endorse Eng's parity-check obligation (files-touched item 3) — if a CLI remote-submit path exists separately from the MCP handler, it must receive the identical trim/forward treatment, since a surface-specific fix would leave one calling surface vulnerable to the same silent-drop failure mode (structural finding per `cross_surface_contract_parity_tested_all_surfaces`, not a nit).

**Verdict:** SIGNED_OFF (waived) — no blocking architectural or security concerns. Arch gate waived per bug-fast-path classification; the checklist above is confirmatory guidance for Eng/QA at PR time, not a gate condition. Flag as `strategy_drift_signal` only if the diff is found to add a new entity-read path or touches `contract_mappings.ts`/`openapi.yaml` — neither is expected per the spec as written.

`[waxwing] schema_or_api_proposal: No schema/contract change required — submit_issue reporter-field forwarding fix confirmed as internal payload-construction only. Arch gate waived (bug fast-path). Confirmatory security checklist added: no tenant-scoped entity lookup in this code path (verify no new one is introduced), idempotency-key presence on the remote POST is pre-existing and undocumented (flag, don't fix here), reporter fields are non-credential metadata (no redaction needed), pre-flight diagnostic must log field names only never values.`

### Legal / Compliance (Buteo)

**Scope confirmation.** Concur with PM/Waxwing's classification: internal payload-construction fix on an existing, already-contracted field set. No new schema, MCP tool, CLI flag, or contract surface. This narrows legal review to three lenses: dependency licensing (none triggered), PII/data-handling on the now-fixed public-effect surface, and credential/token scope on the guest-submission path.

**1. Dependency / licensing.**
- [ ] No new dependency is introduced by this fix (trim/conditional-assignment on existing fields, mirroring `add_issue_message`'s already-shipped pattern). Confirm the diff adds zero `package.json`/lockfile changes — if it does, that's a scope violation per Eng's own build-steps declaration and should be flagged back, not legal-reviewed as a new dependency.
- [ ] No licensing exposure: this is first-party code touching a first-party HTTP client wrapper. N/A beyond the above confirmation.

**2. PII / data-handling on the public-effect surface.**
- The five reporter-provenance fields (`reporter_git_sha`, `reporter_git_ref`, `reporter_app_version`, `reporter_channel`, `reporter_ci_run_id`) are build/CI environment metadata, not personal data by design or by the schema's declared intent. Concur with Waxwing: no redaction/scrubbing logic is warranted for the nominal case.
- [ ] **Public-visibility risk this fix newly activates.** Today the bug means these fields never reach the remote — so whatever they *would* leak was already contained by the bug. Fixing forwarding means these five fields will now flow to a remote instance under `visibility: "public"` submissions (per the issue's own repro payload: `"visibility": "public"`). Before shipping, confirm: does `target_repo`'s public issue view (GitHub-facing, since `target_repo: "markmhendrickson/neotoma"` implies eventual GitHub visibility) render these fields to unauthenticated third parties? If yes, that is an intended and acceptable disclosure (build metadata is routinely public in OSS issue trackers) — no gate — but it should be stated once in the PR description as a "this fix causes previously-undelivered fields to become publicly visible" note, since it's a behavior change in effective exposure even though the code contract didn't change.
- [ ] **Misconfiguration residual risk (flagged, non-blocking, mirrors Waxwing's item).** `reporter_ci_run_id`/`reporter_channel` are free-text strings sourced from caller/CI environment, not validated for content. A misconfigured CI pipeline could in theory interpolate a secret or internal hostname into one of these fields. This is caller-side hygiene, not a defect in this fix — no sanitization should be added here (would be scope creep and could mask the real fix upstream). Recommend one line in `docs/subsystems/issues.md`'s reporter-provenance contract (already on Design/UX's docs checklist) stating these fields are treated as **public, unvalidated, non-secret metadata** and callers are responsible for not populating them with sensitive values. This closes a disclosure-expectation gap cheaply without adding code.
- [ ] No new data collection, no new retention, no new cross-border transfer path is created — the fields already exist in the local contract and are already stored today; this fix only corrects delivery to a remote system that was always the intended destination for public submissions. No DPIA or privacy-notice update triggered.

**3. Guest-token / credential scope.**
- [ ] Confirm (mirrors Waxwing's auth checklist) that none of the five forwarded fields are, or ever become, part of the auth/session-header construction — they must remain body-only fields alongside `github_url`/`author`/`local_issue_id`, never merged into the unsigned-guest token or any credential path. Treat any diff that touches auth-header construction as an out-of-scope change requiring re-review.
- [ ] The unsigned-guest retry path's token/credential *selection* logic is unchanged by this fix (Eng/Waxwing both confirm) — legal has no incremental exposure from guest-token scope here beyond what's already accepted for the existing unsigned-guest submission path. No new anonymous-write capability is being granted; this fix only stops silently truncating an already-authorized field set from an already-permitted call.
- [ ] If Eng's optional pre-flight diagnostic logs field *names* only (per Eng/Waxwing spec) — confirm no log line is later forwarded to a third-party log aggregator without the same treatment as other operational logs; no new compliance obligation if it stays within existing logging infrastructure covered by current data-handling docs.

**Escalation recommendation.** No escalation to qualified counsel required. This is a sub-threshold internal bug fix with no new contract, no new personal-data category, and no new dependency. The one actionable item is documentation-only (state the public-visibility behavior-change note above and the "non-secret, unvalidated metadata" caller-responsibility line in `docs/subsystems/issues.md`) — both already dovetail with Design/UX's existing docs checklist and should be added there rather than as separate legal deliverables.

**Verdict:** APPROVE — no blocking legal/compliance findings. Two non-blocking documentation additions recommended (public-visibility note; non-secret-metadata caller-responsibility line), both foldable into Design/UX's already-planned `docs/subsystems/issues.md` update.

neotoma-agent:
## Summary

`submitIssueToRemote()` in `src/services/issues/neotoma_client.ts` built the outbound `guestBody` payload without ever copying `reporter_git_sha` / `reporter_app_version` / `reporter_git_ref` / `reporter_channel` / `reporter_ci_run_id` from `params`. Every remote `submit_issue` call was rejected with `ERR_REPORTER_ENVIRONMENT_REQUIRED` — even when the caller supplied those fields — because the remote never received them. Local storage kept the fields correctly, which is what masked the bug from local-only testing.

`add_issue_message` already forwards the same fields correctly (`src/actions.ts` ~9944-9955); this PR brings `submit_issue`'s remote path to parity with that existing pattern.

Closes #2014

## Changes

- `src/services/issues/neotoma_client.ts`: added 5 optional camelCase fields to `submitIssueToRemote`'s params type; added 5 trimmed conditional `guestBody` assignments matching the existing `github_url`/`author`/`local_issue_id` pattern. The unsigned-guest `AUTH_REQUIRED` retry path reuses `guestBody` by reference, so it inherits the fix automatically (documented with a comment to prevent future refactor drift).
- `src/services/issues/issue_operations.ts`: `submitIssue()` now passes `params.reporter_git_sha` etc. into `submitIssueToRemote(params)` using the new camelCase names. Pure threading — no new parsing or validation, the fields already exist on the inbound `submit_issue` contract.
- `docs/subsystems/issues.md`: noted that all five fields are now forwarded end-to-end and are regression-tested.

## Cross-surface parity

`submit_issue` is exposed on two surfaces: the MCP tool handler (`src/server.ts:1223`) and the HTTP route (`src/actions.ts:10041`). Both funnel through the same `submitIssue()` handler in `issue_operations.ts`, so one fix covers both structurally. No separate CLI command exists for `submit_issue`. Confirmed via `grep` — no `submit_issue`/`submitIssue` references in `src/cli/index.ts`.

## No contract change

Does not touch `openapi.yaml`, `contract_mappings.ts`, MCP tool schema, or CLI flags. All five fields are already part of the existing, remote-enforced `submit_issue` contract (locally validated and stored today); this is a payload-construction-only fix between the action handler and the remote HTTP client.

## Test plan

- `src/services/issues/neotoma_client.test.ts` — new regression tests asserting on the **constructed outbound request body** (not local storage):
  - primary repro case (`reporter_git_sha` + `reporter_app_version` from the issue's exact payload)
  - each of the 5 fields forwarded individually (parametrized)
  - whitespace-only value resolves to "not sent" (matches existing `.trim()` guard behavior)
  - empty state: no reporter keys sent when none supplied (no forced defaults)
  - mixed-partial case: `reporter_git_ref` alone forwards without requiring `reporter_git_sha`
  - unsigned-guest `AUTH_REQUIRED` retry path preserves reporter fields (pins the by-reference `guestBody` reuse against future refactors)
- `src/services/issues/issue_operations.test.ts` — new test asserting `submitIssue()` threads all 5 fields into the `submitIssueToRemote` call args (camelCase), closing the gap where existing tests supplied `reporter_git_sha` as input but never asserted it reached the remote call.
- All 180 tests in `src/services/issues/` pass. `npm run type-check` and `npm run lint` clean (0 errors).
- No new/moved test files, so no test-catalog regeneration needed.

## Known follow-up (out of scope)

`reporter_patch_source_id` (a distinct field — "source id for reporter patch artifact when applicable") has the same non-forwarding gap on the remote leg, but it's not part of the reporter-*environment* identity set that gates `ERR_REPORTER_ENVIRONMENT_REQUIRED`, and `add_issue_message` (the reference pattern) never forwarded it either. Left out per the issue's explicit 5-field scope; worth a separate follow-up issue if needed.


neotoma-agent:
**COMMENT**

PR #2016 closes #2014. Pre-impl gates on the parent issue entity are clear — Vanellus assigned as `pr_review` owner.

| Gate | Status | Owner |
|---|---|---|
| `pm` | signed_off | Pavo |
| `ux` | signed_off | Accipiter |
| `arch` | signed_off | Waxwing |
| `impl` | signed_off | Cicada |
| `pr_review` | pending | Vanellus |

- [x] Parent issue entity retrieved (`github_number=2014`)
- [x] Pre-impl gates (`pm`/`ux`/`arch`) satisfied (`signed_off`/`waived`/`not_required`)
- [x] Vanellus assigned as reviewer (Phase 4)
- [x] Prior BLOCKED board updated in place (edit-not-duplicate)

**Owner:** Vanellus (`pr_review`) → next Phoenicurus (`qa`)
**Labels:** none applied this turn
**Workflow:** `bug` (fast-path already reflected in live `gate_status`)
**GATE_INHERITANCE:** clear

Source of truth: live `gate_status` on Neotoma issue entity (not GitHub body). Duplicate sync-shaped issue entity without `gate_status` ignored for inheritance.

📎 Neotoma: [neotoma#2014](https://neotoma.markmhendrickson.com/entities/ent_e354f5e935b85ba91b0ed6c3)

neotoma-agent:
review:pm
**APPROVE**

Reviewed diff-only against neotoma#2014's signed-off pm-gate acceptance criteria.

- [x] Forwards all 5 reporter fields (`reporter_git_sha`, `reporter_app_version`, `reporter_git_ref`, `reporter_channel`, `reporter_ci_run_id`) into `guestBody`, trimmed, matching existing optional-field pattern.
- [x] Fix applies to both the primary remote-submit path and the unsigned-guest `AUTH_REQUIRED` retry path (by-reference reuse, documented inline against future refactor drift).
- [x] Effect-verified test: assertions target the constructed outbound request body, not local storage — matches `fixed_means_behavior_verified_not_contract_accepted`.
- [x] Cross-surface parity: PR body confirms `submit_issue` is exposed on the MCP tool handler and HTTP route only, both funnel through the same `submitIssue()` → `submitIssueToRemote()` path, and confirms via grep no CLI command exists — satisfies `cross_surface_contract_parity_tested_all_surfaces` since there's one shared, now-tested code path and no surface-specific call shape to duplicate.
- [x] No unrequested scope creep — no schema, MCP tool, CLI, or entity-type changes. `reporter_patch_source_id` explicitly and correctly declined as out of the issue's named 5-field set. Docs change is a single narrowly-scoped line.
- [x] User-visible behavior matches the issue: remote `submit_issue` calls that supply reporter fields will no longer be rejected with `ERR_REPORTER_ENVIRONMENT_REQUIRED`.

[NON-BLOCKING] acceptance-criteria-gap: issue #2014's acceptance criteria asked for "a previously-failing reproduction now succeeds end-to-end against a representative remote target, OR an equivalent integration/contract test simulates the remote handler's validation to prove the fields arrive." The PR's tests assert on the outbound payload shape via a mocked client — they don't simulate the remote handler's `ERR_REPORTER_ENVIRONMENT_REQUIRED` validator to prove the constructed payload would actually satisfy it. The criterion's OR is arguably met (asserting the exact field names/values the remote's own check keys on, per the issue's quoted root-cause payload, is a reasonable equivalent), but I can't execute anything this run to confirm end-to-end against a live remote. Unverified concern — would be confirmed by either running the existing test suite against a real/staging remote target, or by adding one contract-fixture test that replays the remote's actual validation function against the new payload shape.

neotoma-agent:
review:ux
**APPROVE**

### User goal (not the feature)
Agent/developer submits a public issue to a remote Neotoma with reporter provenance already supplied — and the remote must accept that input rather than claim it was missing.

### Design Spec (ux lens — diff-only)

**User-facing surface**
- Existing `submit_issue` contract (MCP + HTTP). No new fields, flags, or CLI surface. Wire names stay `reporter_git_sha` / `reporter_app_version` / `reporter_git_ref` / `reporter_channel` / `reporter_ci_run_id` (snake_case) — discoverable and consistent with the documented contract and `ERR_REPORTER_ENVIRONMENT_REQUIRED` envelope.
- Internal client params use camelCase (`reporterGitSha`, …) matching the existing `githubUrl` / `local_issue_id` pattern in `neotoma_client.ts` — not a caller-facing rename.

**Interaction / flow**
1. Caller supplies ≥1 of `reporter_git_sha` | `reporter_app_version` (plus optional provenance siblings).
2. Local validate/store succeeds (already did).
3. Remote leg now copies trimmed non-empty values into `guestBody` (primary + unsigned-guest `AUTH_REQUIRED` retry by reference).
4. Remote validator can accept; prior silent-drop false failure is removed.

**Discoverability & naming**
- Field names unchanged end-to-end; fix restores contract honesty (supplied fields actually arrive).
- Docs (`docs/subsystems/issues.md`) note remote-leg forwarding + regression tests + #2014 — findable from the reporter-provenance section callers already use.

**Error & empty states**
- Happy: fields forward → remote no longer falsely emits `ERR_REPORTER_ENVIRONMENT_REQUIRED` when the caller did supply them (root UX defect of #2014).
- Empty / omitted: no reporter keys forced into the body (tested) — remote still returns `ERR_REPORTER_ENVIRONMENT_REQUIRED` with `details.acceptable_field_groups` (pre-existing, actionable).
- Whitespace-only: `.trim()` omits the key (matches sibling optional-field pattern) — remote rejection path unchanged.

**Accessibility**
- N/A for this API/payload fix (no interactive UI surface).

**Acceptance checklist**
- [x] Remote outbound body includes all five reporter-env fields when supplied (asserted on constructed request, not local storage only).
- [x] Unsigned-guest retry preserves the same reporter fields.
- [x] Omitted / whitespace-only values do not invent defaults.
- [x] Docs state end-to-end remote forwarding for the five-field set.
- [x] No silent failure where caller-supplied reporter env is dropped on the remote leg.

### Findings

[NON-BLOCKING] docs-clarity: In `docs/subsystems/issues.md`, the bullet lists six snapshot fields (including `reporter_patch_source_id`) then says “All five fields are forwarded.” Adjacent count mismatch can confuse a reader scanning the list — the five-field env-identity set is clear from #2014 context, but a one-word clarification (e.g. “all five reporter-*environment* fields”) would remove the ambiguity. Diff-evident wording only; not a merge blocker.

[NON-BLOCKING] error-hint-parity: Whitespace-only reporter values are silently stripped before send, so the operator can still see `ERR_REPORTER_ENVIRONMENT_REQUIRED` after “providing” a blank string. Matches existing optional-field trim pattern; unverified whether real callers hit this. Confirm by sending `"reporter_git_sha": "   "` against a live/staging remote and checking whether the error envelope (or a local preflight) could call out blank-after-trim. Out of this PR’s payload-forward scope.

No blocking findings from the ux lens. The PR removes a P0 silent-failure (system rejects input it already accepted locally).

**Verdict:** `APPROVE`

---
📎 Neotoma: [neotoma#2014](https://neotoma.markmhendrickson.com/entities/ent_e354f5e935b85ba91b0ed6c3)

neotoma-agent:
review:qa
**APPROVE**

Scope: regression coverage for the reporter-provenance forwarding fix (`submitIssueToRemote` / `submitIssue`).

## Verification performed

Installed deps (`npm ci`) and ran the tests directly rather than reasoning from the diff.

- `npx vitest run src/services/issues/neotoma_client.test.ts src/services/issues/issue_operations.test.ts` → **66/66 pass** on the PR branch as-is.
- **Reverted only the two source files** (`neotoma_client.ts`, `issue_operations.ts`) to their pre-fix (`HEAD~1`) content, keeping the new tests, and re-ran: **9 tests fail** with `expected undefined to be '<value>'` on `request.body.reporter_*` assertions — confirming these are real regression tests that catch the exact bug (not tautological or accidentally-passing). Restored the fix afterward; `git status` is clean.
- Full `src/services/issues/` suite: **180/180 pass**. `npx tsc --noEmit -p .` clean.
- Wrote and ran a throwaway probe (`issue_operations.probe.test.ts`, deleted after) calling the real `submitIssue()` entry point with only `reporter_git_ref` set (no `reporter_git_sha`/`reporter_app_version`): it throws `ERR_REPORTER_ENVIRONMENT_REQUIRED` before `submitIssueToRemote` is ever invoked (gate at `issue_operations.ts:846` → `assertSubmitIssueReporterEnvironment`). Also probed whitespace-only `reporter_git_sha` against the same gate — also rejected, so the gate's trim semantics match the forwarding `.trim()` guards; no silent-drop gap there.

## Coverage assessment

- **Fault-site assertion correctness**: tests assert on the constructed outbound `request.body` at the HTTP-client boundary, not on local storage — this is the right assertion point, since local storage already worked and masked the original bug (per the PR's own description, confirmed by reading the pre-fix diff).
- **Regression repro case**: present, parametrized per-field (all 5 reporter fields individually), empty-state, whitespace-trim, and the unsigned-guest `AUTH_REQUIRED` retry path (which reuses `guestBody` by reference — a real future-refactor footgun, correctly pinned).
- **Threading test**: `issue_operations.test.ts` asserts `submitIssue()` passes all 5 camelCase params into `submitIssueToRemote`, closing the gap the PR description calls out (previous tests supplied `reporter_git_sha` as input but never asserted it reached the remote call args).
- **Two production surfaces** (MCP tool handler, HTTP route) funnel through the same `submitIssue()` → `submitIssueToRemote()` path per the PR description; confirmed via `grep` that no separate CLI command exists for `submit_issue`. One fix path, one test layer, structurally sufficient — no surface-specific contract test needed here.

## Non-blocking observations

[NON-BLOCKING] test-realism: The "mixed-partial case" test in `neotoma_client.test.ts` (`forwards reporter_git_ref alone ... without requiring reporter_git_sha`) exercises a `submitIssueToRemote` param shape that is **unreachable from the real `submitIssue()` entry point** in production — `assertSubmitIssueReporterEnvironment` (`issue_operations.ts:846`) throws `ERR_REPORTER_ENVIRONMENT_REQUIRED` first whenever both `reporter_git_sha` and `reporter_app_version` are absent, verified by direct probe against `submitIssue()`. The test is still valid as a unit-level check of `submitIssueToRemote`'s own forwarding logic in isolation (and doesn't cause any false confidence about correctness), but the "mixed-partial" framing implies a production-reachable scenario that isn't one. Consider a one-line comment noting this exercises the client function directly, independent of the caller-side gate, so a future reader doesn't assume it's end-to-end reachable.

No blocking findings. The regression tests are demonstrated (not just claimed) to catch the fixed bug, cover the field-level edge cases (individual fields, empty state, whitespace, retry-path), and sit at the correct assertion boundary.

📎 PR: markmhendrickson/neotoma#2016

neotoma-agent:
review:content
**COMMENT**

Shippable as-is; no content/dogfooding blocker.

**Content lens read:**
- This is genuine build-in-public material: the bug isn't "we added a field," it's "local storage masked a remote-forwarding regression for months, and the fix is inseparable from a test-discipline change that asserts on the outbound request body instead of local state." That's a process-integrity story, not a feature announcement — fits the trust-hinge criterion (evidence, not marketing).
- Fully genericizable, no named-customer/consent gate applies — this is our own pipeline's bug.
- Claim discipline check: every claim in the PR body traces to the diff/tests I read (forwarding gap confirmed in `neotoma_client.ts`, parity with `add_issue_message` confirmed, regression tests assert on `request.body.reporter_*` not local storage). Nothing here needs a `[VERIFY:]` marker.
- **Not shippable as a standalone post today** — it's thin on its own: one PR, one bug class, no reader-facing stakes beyond "we fixed our own tooling." Doesn't clear the bar alone.

**Queued, not blocking:** filed as a `post_idea` (`ent_1ed22edc87373eb56a19ff16`) — angle: "the provenance fields that only mattered on the remote leg." Best used folded into a broader piece on the issue-reporting pipeline's provenance model, or a "small bugs, big blind spots" post if 2–3 similar local-vs-remote-parity bugs accumulate. Low priority; revisit if content-strategy (Ciconia) sequences a testing/dogfooding themed post.

📎 Neotoma: see post_idea `ent_1ed22edc87373eb56a19ff16`

neotoma-agent:
**APPROVE**

### Per-lens roll-up
- **pm** (Pavo): **APPROVE** — 5 reporter fields forwarded on both remote paths; effect-verified outbound-body tests; cross-surface parity (MCP+HTTP → shared `submitIssue`); no scope creep. 1 NON-BLOCKING (e2e/remote-validator simulation not executed this panel).
- **ux** (Accipiter): **APPROVE** — restores contract honesty (supplied reporter env arrives remotely); silent `ERR_REPORTER_ENVIRONMENT_REQUIRED` false-failure removed. 2 NON-BLOCKING (docs five-vs-six wording; whitespace-trim hint).
- **qa** (Phoenicurus): **APPROVE** — ran tests: 66/66 on PR; reverting source-only fails 9 regression assertions on `request.body.reporter_*`; full `src/services/issues/` 180/180; `tsc` clean. 1 NON-BLOCKING (mixed-partial unit path unreachable via production `submitIssue` gate).
- **content** (Corvus): **COMMENT** — shippable; queued `post_idea` `ent_1ed22edc87373eb56a19ff16` for a future provenance/blind-spot piece (not standalone).

### Blocking vs non-blocking
- [x] Blocking: **0**
- [ ] No [BLOCKING] findings from any lens

### Effect / parity gates
- [x] Effect-verified tests assert outbound request body (not local storage only) — `fixed_means_behavior_verified_not_contract_accepted`
- [x] Cross-surface parity: one shared path for MCP + HTTP; no CLI surface — `cross_surface_contract_parity_tested_all_surfaces`

### Gate inheritance (parent #2014 / `ent_e354f5e935b85ba91b0ed6c3`)
- [x] pm: signed_off
- [x] ux: signed_off
- [x] arch: signed_off
- [x] impl: signed_off
- [x] Required check `security_gates`: SUCCESS (all advisory lanes also green)

### Merge recommendation
**APPROVE** and squash-merge. `APIS_AUTONOMY_AUTO_MERGE=1`. Confidence ≥0.85 (full panel + CI + gate snapshot + QA-executed regression proof). Releases remain human-gated.

**Verdict:** APPROVE
**Blocking:** 0
Reviewed commit: 54c20f4b968dab4cc407641e3ea86e235c05ad0d

---
📎 Neotoma: [neotoma#2014](https://neotoma.markmhendrickson.com/entities/ent_e354f5e935b85ba91b0ed6c3)

neotoma-agent (commented):
**APPROVE**

### Per-lens roll-up
- **pm** (Pavo): **APPROVE** — 5 reporter fields forwarded on both remote paths; effect-verified outbound-body tests; cross-surface parity (MCP+HTTP → shared `submitIssue`); no scope creep. 1 NON-BLOCKING (e2e/remote-validator simulation not executed this panel).
- **ux** (Accipiter): **APPROVE** — restores contract honesty (supplied reporter env arrives remotely); silent `ERR_REPORTER_ENVIRONMENT_REQUIRED` false-failure removed. 2 NON-BLOCKING (docs five-vs-six wording; whitespace-trim hint).
- **qa** (Phoenicurus): **APPROVE** — ran tests: 66/66 on PR; reverting source-only fails 9 regression assertions on `request.body.reporter_*`; full `src/services/issues/` 180/180; `tsc` clean. 1 NON-BLOCKING (mixed-partial unit path unreachable via production `submitIssue` gate).
- **content** (Corvus): **COMMENT** — shippable; queued `post_idea` `ent_1ed22edc87373eb56a19ff16` for a future provenance/blind-spot piece (not standalone).

### Blocking vs non-blocking
- [x] Blocking: **0**
- [ ] No [BLOCKING] findings from any lens

### Effect / parity gates
- [x] Effect-verified tests assert outbound request body (not local storage only) — `fixed_means_behavior_verified_not_contract_accepted`
- [x] Cross-surface parity: one shared path for MCP + HTTP; no CLI surface — `cross_surface_contract_parity_tested_all_surfaces`

### Gate inheritance (parent #2014 / `ent_e354f5e935b85ba91b0ed6c3`)
- [x] pm: signed_off
- [x] ux: signed_off
- [x] arch: signed_off
- [x] impl: signed_off
- [x] Required check `security_gates`: SUCCESS (all advisory lanes also green)

### Merge recommendation
**APPROVE** and squash-merge. `APIS_AUTONOMY_AUTO_MERGE=1`. Confidence ≥0.85 (full panel + CI + gate snapshot + QA-executed regression proof). Releases remain human-gated.

**Verdict:** APPROVE  
**Blocking:** 0  
Reviewed commit: `54c20f4b968dab4cc407641e3ea86e235c05ad0d`

---
📎 Neotoma: [neotoma#2014](https://neotoma.markmhendrickson.com/entities/ent_e354f5e935b85ba91b0ed6c3)

---

**Merged.** Squash merge commit `cb04639a16153483d818ff631dd25a918886c7da`. Aggregation comment: https://github.com/markmhendrickson/neotoma/pull/2016#issuecomment-5355241951

**Post-merge handoff:** `pr_review` → `signed_off`; `current_owner` → `phoenicurus` (Phase 4b). Native `--approve` failed (`Review Can not approve your own pull request` — shared dispatch identity); merge proceeded on panel + checks + gate inheritance. `plan_contribution` / conversation store blocked by agent write ACL (corrections on the issue entity succeeded).

🧠 Neotoma
- issue `ent_e354f5e935b85ba91b0ed6c3` (neotoma#2014) — retrieved; `pr_review` signed_off; owner → phoenicurus
- post_idea `ent_1ed22edc87373eb56a19ff16` — cited from content lens (not created this turn)
- `plan_contribution` / conversation store skipped (agent policy closed on those types)

[vanellus] merge_decision: MERGED PR#2016 squash sha=cb04639a16153483d818ff631dd25a918886c7da; verdict=APPROVE Blocking=0; pr_review signed_off on ent_e354f5e935b85ba91b0ed6c3; owner=phoenicurus
[vanellus] strategy_drift_signal: github_harness Vanellus identity can correct issue gates but cannot store conversation/plan_contribution, and cannot --approve PRs authored under the same shared account

PR #2007, opened 2026-07-23
fix(oauth): give sign-in sessions their own TTL and bound-user lifetime

markmhendrickson:
Closes #2005.

## Symptom

Google-signed-in users on a hosted instance are silently logged out and must re-authenticate.

Initially assumed to be the same cause as the MCP session drops fixed by #2002 (machine idling to zero). **It is not** — the affected machine is confirmed always-on (`autostop: False`) and sessions still cleared.

## Root cause

**1. Sign-in reused the key-entry gate's TTL.** `OAuthKeySessionStore` defaults to `15 * 60 * 1000`, a sensible lifetime for the brief window in which a user pastes a private key / mnemonic / bearer token. The Google sign-in path reuses the same store, inheriting a TTL chosen for a different purpose. The store already accepted `ttlMs` in its constructor — the call site (`actions.ts`) just never passed one.

**2. The verified-user map had no lifetime at all.** `googleVerifiedUserBySessionToken` was a module-level `Map<string,string>`: no TTL, no eviction, entries only ever added. On a long-lived always-on instance that grows unbounded, and an expired session could still resolve a bound user straight out of memory.

## Changes

- `setOAuthKeySessionCookie` takes an optional `ttlMs` covering **both** the server-side session and the cookie's `maxAge`, so the two cannot desync (the browser presenting a cookie the server has forgotten, or vice versa).
- `SIGN_IN_SESSION_TTL_MS` — default **7 days**, override with `NEOTOMA_SIGN_IN_SESSION_TTL_MIN` — passed **only** on the sign-in path. The key-entry gate keeps its 15-minute default; this deliberately does not widen it.
- The user binding moves into `OAuthKeySessionStore` (`bindUser` / `getBoundUser`) so it expires with its session and is evicted by the existing `cleanup()`. `bindUser` refuses expired/unknown tokens; `getBoundUser` re-checks validity rather than trusting the map.

## Testing

5 new cases, each **verified to fail against the prior implementation** and pass with the fix, per `fixed_means_behavior_verified_not_contract_accepted`:

```
× honors a per-session TTL override longer than the store default
× resolves a bound user while the session is live
× stops resolving the bound user once the session expires
× refuses to bind a user to an expired or unknown session
× evicts bound users during cleanup so the map cannot grow unbounded
```

10/10 pass with the fix. Typecheck clean. Deterministic clocks — no wall-time sleeps.

## Note on restart behavior

[...]

Issue #2005, opened 2026-07-23
Google sign-in sessions clear after ~15 minutes: sign-in reuses the key-entry gate's short TTL, and the verified-user map is unbounded + process-local


markmhendrickson:
## Symptom

Operator report on a hosted instance: *"My Google-based sessions seem to clear within a few hours."* Users signed in with Google are silently logged out and must re-authenticate.

This was initially assumed to be the same root cause as the MCP session drops fixed by #2002 (machine idling to zero wiping in-memory state). **It is not** — the machine is confirmed always-on (`autostop: False`) and the sessions still clear.

## Root cause

Two distinct problems, both in the OAuth session layer.

### 1. A 15-minute TTL is applied to sign-in sessions

`src/services/oauth_key_gate.ts`:
```ts
const DEFAULT_SESSION_TTL_MS = 15 * 60 * 1000;
```

`OAuthKeySessionStore` is constructed with **no argument** (`src/actions.ts:1383`), so every session gets 15 minutes. The browser cookie matches (`setOAuthKeySessionCookie` sets `maxAge: 15 * 60 * 1000`).

That default is reasonable for what the store was originally built for — a short-lived **key-entry gate** guarding private-key / mnemonic / bearer entry. It is **not** reasonable for a **Google sign-in session**, where a user expects to stay signed in for hours or days. The Google flow reuses the same store (`googleVerifiedUserBySessionToken.set(sessionToken, resolvedUserId)` in the OAuth callback), inheriting a TTL chosen for a different purpose.

The TTL is **not configurable** — no env var, no config key, hardcoded constant.

### 2. `googleVerifiedUserBySessionToken` is an unbounded, process-local Map

`src/actions.ts:1394`:
```ts
const googleVerifiedUserBySessionToken = new Map<string, string>();
```

- **No TTL and no eviction.** Unlike `OAuthKeySessionStore` (which has `cleanup()`), entries are only ever added. On a long-lived always-on instance this grows unbounded — a slow leak.
- **Process-local.** Any restart or redeploy drops every verified user. Same in-memory fragility class as `mcpTransports` (the MCP session map behind #2002), so the two failure modes compound: idle-stop is fixed, restart-on-deploy is not.

## Impact

Highest on **hosted, client-facing instances using Google sign-in by approved-email allowlist** — exactly the configuration where users are least able to diagnose "why am I logged out again."

## Proposed fix

[...]

PR #2021, opened 2026-07-27
fix(inspector): refresh OAuth access token instead of hard-expiring (#2005)

### LINKED ISSUE #1997: Inspector session drops after 1 hour — access token expires and the SPA never refreshes it
## Symptom

The Inspector shows **"Could not load entities — Your session is no longer valid. Sign in again from Settings."** after roughly an hour of use, requiring a manual re-sign-in. Reported on the hosted instance at `bottega8-neotoma.fly.dev` (v0.18.8 / `86fc2c3`).

## Root cause

**OAuth access tokens have a 1-hour TTL and the Inspector has no refresh logic.**

- The server issues access tokens with `expires_in: 3600` (`src/services/mcp_oauth.ts:1076`, `:1423`, `:1531`).
- The Inspector stores only the access token, in `localStorage` under `neotoma_inspector_auth_token` (`inspector/src/api/client.ts:152-157`), and attaches it to every request.
- **`inspector/src` contains no reference to `refresh_token` or `refreshToken` anywhere.** There is no refresh-on-401, no proactive refresh before expiry, and the refresh token is never stored client-side.

So the SPA sends a stale bearer token until the server rejects it. The 401 body is rewritten to the user-facing string by `rewriteAuthMessage` (`inspector/src/api/client.ts:174-183`), which matches `/invalid (authentication )?token/i` — i.e. the token is **present but expired**, distinct from the "Authentication required" wording used when no token is sent. That distinction confirms expiry rather than a missing/cleared token.

The server side already has what's needed: `mcp_oauth_connections` persists `refresh_token` alongside `access_token` and `access_token_expires_at`, and the token endpoint supports the refresh grant (`src/services/mcp_oauth.ts:186-237`, `:1243`).

## Not the cause

- **Not a redeploy artifact.** `bottega8-neotoma` has a persistent Fly volume (`vol_vgn5z7zzndlo8lj4`) mounted at `/app/data`, and `mcp_oauth_connections` lives in that SQLite DB — so OAuth connections survive rebuilds. (Ruling this out because it was the natural first guess: the expiry window and a deploy cadence can look alike.)
- Not `localStorage` being cleared — that path produces the "Authentication required" wording instead.

## Expected

The session should survive well beyond an hour without manual re-authentication. Options, roughly in order of effort:

1. **Refresh on 401** — on an auth failure, attempt a refresh-token exchange once and replay the request; only surface the error if that fails. Smallest change that fixes the visible symptom.
2. **Proactive refresh** — persist `access_token_expires_at` client-side and refresh shortly before expiry, avoiding the user-visible failure entirely.
3. **Longer-lived Inspector sessions** — raise the TTL for the Inspector's own connection, or issue it a session cookie rather than a raw bearer token.

(1) and (2) are complementary; (2) alone leaves a gap if the tab sleeps through the refresh window, so an interceptor is worth having regardless.

## Repro

1. Sign in to the Inspector at `/entities/contact`.
2. Leave the tab open, or return after ~1 hour.
3. Any data fetch fails with the message above; re-signing in from Settings restores it for another hour.

## Swarm specification

### Product / Scope (PM)

You've hit your session limit · resets 7:40pm (Europe/Madrid)

### Engineering

You've hit your session limit · resets 7:40pm (Europe/Madrid)

### QA / Test Plan

You've hit your session limit · resets 7:40pm (Europe/Madrid)

### Security / Arch

You've hit your session limit · resets 7:40pm (Europe/Madrid)

### Legal

You've hit your session limit · resets 7:40pm (Europe/Madrid)

markmhendrickson:
Closes #1997.

Supersedes #2005 (filed a day later as a duplicate of this issue; closing that one).

Continues #2005. Closes the actual cause of the "Your session is no longer valid" sign-outs on a hosted instance.

## Why the earlier fix didn't fix it

#2007 extended the OAuth **key-session cookie** TTL (and the `googleVerifiedUserBySessionToken` map). That was a real fix — but a **different session mechanism**. The Inspector authenticates its API calls with a **bearer access token** stored in `localStorage`, obtained from `/mcp/oauth/token` at Google sign-in. That token has its own **1-hour** `expires_in`, and nothing renewed it — so it hard-expired and signed the user out. The symptom persisted after #2007 deployed because #2007 never touched this token.

I diagnosed the wrong layer the first time. This traces it end to end.

## Root cause

`oauth_signin.ts` parsed the token response as `{ access_token }` and **discarded** the `refresh_token` and `expires_in` the server returns. The refresh token was valid and sitting unused; the Inspector just never kept it.

The server side was already complete — `/mcp/oauth/token` accepts `grant_type=refresh_token`. So this is a **purely client-side** fix.

## The change

- **`setAuthSession()`** persists `access_token` + `refresh_token` + a computed `expires_at` (replacing access-token-only storage on the sign-in path). A bundle with no refresh token (pasted bearer / sandbox) clears refresh state.
- **`request<T>()`** transparently refreshes-and-retries once on a `401` when a refresh token is present. A persistent 401 still surfaces the auth error (an `isRetry` guard prevents a loop); a rejected refresh token clears the session so the UI shows a clean sign-in.
- **Single-flight refresh** — concurrent 401s (the home screen fires several `totals` calls at once) coalesce onto one refresh, so a burst can't each spend the one-time-use refresh token.
- **`authorizedHeaders()`** proactively refreshes within 60s of expiry, so `getText`/`getBlob` (direct-fetch paths) renew too, not just `request<T>`.
- **`clearAuthToken()`** now purges refresh + expiry alongside the access token.

## Tests

`inspector/src/api/token_refresh.test.ts` — bundle persistence, 401 refresh-and-retry, no-double-retry, dead-refresh-clears-session, concurrent-401 coalescing. The four retry-path cases were **verified to fail with the retry guard disabled** and pass with it. **122/122** inspector tests pass; typecheck clean.

## Note

This closes the *hard-expiry* gap on the primary browser-session path. Access-token lifetime is still 1 hour, but the session now renews silently across it. If the browser is closed longer than the refresh token's own lifetime, a fresh sign-in is still required — which is correct.


neotoma-agent:

neotoma-agent:

markmhendrickson:
**Self-flagged cross-surface parity gap** (pre-empting the arch/qa lens, which rightly objects to a one-surface fix on a multi-surface capability):

The `neotoma` **CLI** shares the same discard-the-refresh-token bug. `exchangeToken()` (`src/cli/index.ts:~1525`) returns only `{ access_token, token_type, expires_in }` — it drops `refresh_token` — and `runLoginFlow` persists that to the config file with an `expires_at` but no refresh path. So a CLI user's token also dies at 1 hour.

**Why I'm scoping the CLI out of this PR rather than fixing it here:**

- The CLI *does* recover — `ensureSessionOrAuth` re-runs `runLoginFlow` on a 401 (`index.ts:1638`). The UX is worse (a full interactive browser re-login every hour) but it is not a silent break, unlike the Inspector, which just showed an error and stopped.
- The fix is a genuinely separate change to a different file: thread `refresh_token` into `writeConfig`, and add a refresh attempt in `ensureSessionOrAuth` / `getCliToken` before falling back to the interactive flow.

Folding that into this PR would conflate two surfaces and two auth-persistence models. Filing it as a follow-up instead so the parity gap is tracked, not silently ignored. This PR remains the Inspector fix, which is the surface with the user-visible break reported on the hosted instance.

markmhendrickson:
CLI parity follow-up filed as #2022.

neotoma-agent:
**COMMENT**

## Triage & Gate-Status Board

**Parent issue:** #1997 (Inspector session drops after 1 hour — access token expires and the SPA never refreshes it)

**Legacy issue note:** #1997 predates gate tracking — its Neotoma `issue` entity carried no `gate_status` at all. Retroactively initialized per the legacy-issue rule (merge-only; nothing existing was downgraded).

| Gate | Status | Owner |
|---|---|---|
| pm | not_required | — (legacy backfill; issue was fully diagnosed pre-PR) |
| ux | not_required | — (bug fast path skips ux) |
| arch | not_required | — (legacy backfill; #2021 supersedes duplicate #2005) |
| impl | signed_off | Gryllus |
| legal | not_required | — |

**Pre-impl gates (pm/ux/arch):** all satisfied → **GATE_INHERITANCE: clear**

**Owner:** current phase → Vanellus (pr_review). Next phase → Phoenicurus (qa), joined by Buteo only if legal_required.

**Labels:** none applied by this check (PR-event protocol, not issue-triage).

**Reviewer assignment:** Vanellus assigned as PR reviewer.

---
**Update (synchronize event):** new commits pushed to `fix/inspector-oauth-token-refresh`. Re-checked parent issue #1997 gate_status in Neotoma — unchanged, still clear. No re-triage needed; Vanellus remains assigned reviewer for pr_review.

📎 Neotoma: [issue#1997](https://neotoma.markmhendrickson.com/entities/ent_f69dd0c40686f8b4d26bea5f)

neotoma-agent:
review:pm

Reviewed against the acceptance criteria in #1997 (the issue this PR closes) — user-visible symptom, root cause, and proposed fix.

**Scope match**: PR fixes exactly the mechanism #1997 describes — Inspector stores only the access token, no refresh-on-401, no proactive refresh, refresh_token discarded at sign-in. The fix implements both fix options #1997 itself proposed as complementary: (1) refresh-on-401 and (2) proactive refresh — matching "an interceptor is worth having regardless" from the issue body.

**No unrequested scope creep**: diff is 3 files (`inspector/src/api/client.ts`, `inspector/src/lib/oauth_signin.ts`, new `token_refresh.test.ts`), all directly in service of the token-refresh mechanism. No schema, MCP tool, CLI command, or contract-surface changes — bug fast-path (skip ux/arch) is correctly applicable here.

**User-visible behavior matches issue**: sessions renew silently across the 1-hour access-token lifetime instead of hard-expiring, matching #1997's "Expected" section. Single-flight coalescing and the isRetry guard are scoped defenses against the retry mechanism itself misbehaving (thundering-herd refresh, infinite retry loop) rather than added scope.

**Effect-verified**: PR description states the four retry-path test cases were verified to fail with the retry guard disabled and pass with it — satisfies effect-verified-not-contract-accepted.

[NON-BLOCKING] duplicate-issue-reconciliation: PR body states it "Supersedes #2005 (filed a day later as a duplicate of this issue; closing that one)." I confirmed #2005 was a separately-filed, PM-signed-off issue already mid-review (with its own merged fix, PR #2007, addressing a genuinely different root cause — the 15-min OAuth key-session cookie TTL). #2005's reporter-described *symptom* ("session is no longer valid" after ~an hour) matches #1997 more precisely than #2005's own diagnosed root cause did, so treating it as a duplicate report of #1997 rather than a distinct bug looks correct on the merits. Flagging only because #2005 had a PM sign-off on file (`plan_contribution` ent_894d3beba01e795fc1465320) that this closure implicitly supersedes without a corresponding pm-lens note on #1997 itself — recommend Lanius/operator add a cross-reference note on #2005 confirming the reconciliation for the audit trail, per this repo's issue-entity gate-history convention. Not a merge blocker.

neotoma-agent:
review:arch

Scope reviewed: `inspector/src/api/client.ts`, `inspector/src/lib/oauth_signin.ts`, `inspector/src/api/token_refresh.test.ts`. Client-side only — no server handler, schema, or entity-lookup surface touched by this diff.

**[NON-BLOCKING] contract-declaration: `refresh_token` grant is undeclared in `openapi.yaml` for `mcpOAuthToken`**

`openapi.yaml` (`/mcp/oauth/token`, operationId `mcpOAuthToken`) declares the request body as `{ grant_type, code }` only, and the 200 response schema as `{ access_token, token_type, expires_in }` only — no `refresh_token` in either direction. This PR's client now sends `grant_type=refresh_token&refresh_token=...` (`client.ts` `refreshAccessTokenOnce`) and reads `refresh_token` back off the response (`setAuthSession`), against a contract the spec doesn't describe. Per `docs/architecture/change_guardrails_rules.mdc` MUST #1 ("spec before handler") and the Touchpoint Matrix row for "HTTP endpoint / request / response field" (`openapi.yaml` → `openapi_types.ts` → `contract_mappings.ts` → handler → contract tests), this pairing should be spec-first.

Not blocking this PR specifically: the server-side `refresh_token` grant already exists and is exercised by `src/services/__tests__/mcp_oauth.test.ts` and `tests/integration/mcp_oauth_token_endpoint.test.ts` — this PR is a consumer of pre-existing, functioning server behavior, not the origin of the gap, and the PR body is explicit that the server side was "already complete." But the gap is real and this PR doubles the client's structural dependence on an undocumented contract shape (both the initial-signin bundle and the new refresh-exchange bundle now assume `refresh_token`/`expires_in` fields the spec never promised). Recommend a fast-follow: add `refresh_token` (request + response) and `token_type`/rotation semantics to the `mcpOAuthToken` schema in `openapi.yaml`, so `openapi_types.ts` gates client parsing on a declared shape rather than an inline hand-typed interface (`{ access_token?: string; refresh_token?: string; expires_in?: number }` duplicated in both `client.ts` and `oauth_signin.ts`).

**[NON-BLOCKING] layering: acceptable — no State Layer or entity surface touched**

This is pure Inspector-client session-management logic (localStorage bundle, single-flight refresh, proactive skew-refresh, 401-retry-once). It doesn't touch `src/actions.ts`, tenant-scoped queries, entity/observation writes, or any mutating Neotoma operation, so tenant-isolation (MUST #5) and `idempotency_key` (MUST #11) are not implicated — there's no new mutating op here, just a token-refresh POST against an existing, already-idempotent-by-design OAuth exchange. Single-flight coalescing (`refreshInFlight`) and the `isRetry` guard against infinite 401-loop are the right shape for this problem and don't introduce a parallel auth-resolution path (MUST NOT #5 is about server-side `getAuthenticatedUserId`, unaffected here).

**[NON-BLOCKING] agent-instruction coherence: not applicable**

No MCP tool, CLI command, or response signal was added that agents need to be told to use — this is browser-session plumbing invisible to MCP/CLI callers. `docs/developer/mcp/instructions.md` and `docs/developer/cli_agent_instructions.md` parity is not implicated.

**Verdict: no structural changes required to merge.** The one open item (declare `refresh_token` in `openapi.yaml` for `mcpOAuthToken`) is a pre-existing contract gap this PR extends reliance on rather than introduces, and is appropriately deferred as a fast-follow rather than blocking a bug fix for an active sign-out regression.

📎 Neotoma: see `ent_3425a79b4c39f08cdb0c62f8` (waxwing agent_definition)

neotoma-agent:
review:qa

**COMMENT**

Scope: test coverage adequacy only (regression test for the fixed bug, edge cases for new branches, contract tests for new endpoints). No correctness/security pass — that's the GHA lane.

## Coverage assessment

`inspector/src/api/token_refresh.test.ts` is a strong, purpose-built regression suite for the actual bug (#2005 — refresh_token/expires_in discarded on sign-in, causing hard-expiry sign-outs). It covers:
- `setAuthSession` bundle persistence (refresh token + computed expiry)
- pasted-bearer path clearing stale refresh state
- `clearAuthToken` purging all three keys
- 401 → refresh → retry-once (happy path)
- no-double-retry on a persistent 401 (`isRetry` guard)
- dead/rejected refresh token clearing the session
- concurrent-401 single-flight coalescing (the one-time-use refresh token burst scenario)

The PR body states the four retry-path cases were verified to fail with the retry guard disabled and pass with it — that's exactly the kind of red/green regression discipline this lens looks for, and it directly targets the mechanism that caused the original bug (hard-expiry with an unused, valid refresh token).

## Findings

[NON-BLOCKING] test-coverage: `authorizedHeaders()`'s proactive refresh branch (the 60s-expiry-skew path feeding `getText`/`getBlob`, `inspector/src/api/client.ts` lines ~349–361) has no test exercising it. All four new test cases in `token_refresh.test.ts` drive the *reactive* 401-retry path in `request<T>` via `get()`. The proactive-refresh branch is new logic with its own condition (`expiresAt != null && refreshToken && Date.now() >= expiresAt - SKEW`) that could regress silently — e.g. a sign flip or off-by-one in the skew comparison would go undetected since no test sets `getTokenExpiresAt()` near the boundary and calls `getText`/`getBlob`. Suggested case: set an expiry ~30s out, call `getText()` or `getBlob()`, assert the refresh endpoint is hit before the direct fetch and the request carries the fresh token.

[NON-BLOCKING] test-coverage: the refresh-token-rotation fallback in `refreshAccessTokenOnce` (`bundle.refresh_token ?? refreshToken`, `client.ts` line ~336) is only tested for the case where the server *does* rotate the token (existing test asserts `getRefreshToken()` becomes `"ref-2"`). The complementary branch — server omits `refresh_token` on the refresh response and the client must keep reusing the original — is untested. If this fallback regressed (e.g. a future refactor drops the `?? refreshToken`), `setAuthSession` would clear the refresh token on every non-rotating refresh response and silently downgrade every session back to hard-expiry, which is the exact class of bug this PR fixes. Suggested case: mock the refresh response with `{ access_token: "fresh" }` (no `refresh_token` field) and assert `getRefreshToken()` still returns the original.

## Non-functional-surface note

`oauth_signin.ts`'s `completeOAuthSignIn` change (passing the full bundle instead of just `access_token`) has no direct unit test — there was no pre-existing `oauth_signin.test.ts` in the repo before this PR, so this isn't a regression in this PR's coverage baseline. `setAuthSession` (the function `completeOAuthSignIn` now calls) is directly and adequately tested in `token_refresh.test.ts`, and `playwright/tests/oauth-flow.spec.ts` exists for the end-to-end sign-in flow, so the initial sign-in path is not left dark — flagging only because a thin call-site-level test (`completeOAuthSignIn` maps `token.refresh_token`/`token.expires_in` through unchanged) would close the loop between the parsing change and the tested persistence function.

## Verdict

No blocking gaps. The regression test for the actual reported bug (hard-expiry from a discarded refresh token) is present, red/green-verified per the PR description, and covers the primary and highest-risk edge cases (concurrency, dead-token, no-loop). The two non-blocking items are coverage gaps in secondary branches of new code, not evidence of a live bug.

---
📎 Standing lens: eval/test-coverage criteria per Phoenicurus QA lens (regression test for fixed bug, edge cases for new branches, contract tests for new endpoints). No pre-registered expectations for this issue.

neotoma-agent:

**APPROVE**

Aggregated review panel verdict for PR #2021 (closes #1997, supersedes #2005). Panel lenses: pm, arch, qa — all reviewed, none blocking.

## Gate inheritance (parent issue #1997)
- pm: `not_required` ✓
- ux: `not_required` (bug fast-path) ✓
- arch: `not_required` ✓
- impl: `signed_off` ✓

Inheritance clears — no pre-impl gate blocks this merge.

## Per-lens roll-up

**pm** (Pavo) — no blocking findings. Scope matches #1997 exactly (refresh-on-401 + proactive refresh, both options the issue proposed); no scope creep; effect-verified per the PR's red/green test description.
- [NON-BLOCKING] `#2005` had its own PM sign-off (`plan_contribution` ent_894d3beba01e795fc1465320) now implicitly superseded by this PR's closure of #2005 as a duplicate of #1997. Recommend a cross-reference note on #2005 for the audit trail. Not a merge blocker.

**arch** (Waxwing) — no structural changes required to merge. Inspector-client-only diff; no State Layer, entity surface, or tenant-isolation logic touched.
- [NON-BLOCKING] `refresh_token` grant (request + response) is undeclared in `openapi.yaml` for `mcpOAuthToken`, though server-side behavior is pre-existing, tested, and functioning. This PR extends client reliance on that undocumented shape. Recommend fast-follow to declare it per spec-before-handler guardrails.

**qa** (Phoenicurus) — COMMENT, no blocking gaps. `token_refresh.test.ts` (5 new cases) is a red/green-verified regression suite directly targeting the reported bug (bundle persistence, 401 refresh-and-retry, no-double-retry guard, dead-refresh-token session clear, concurrent-401 single-flight coalescing).
- [NON-BLOCKING] proactive-refresh branch in `authorizedHeaders()` (60s-skew path) has no direct test.
- [NON-BLOCKING] non-rotating-refresh-token fallback (`bundle.refresh_token ?? refreshToken`) is untested for the case where the server omits `refresh_token` on refresh.

## Effect-verified + cross-surface parity check
- Effect-verified: yes — PR description states the four retry-path cases were verified to fail with the guard disabled and pass with it (`fixed_means_behavior_verified_not_contract_accepted` satisfied).
- Cross-surface parity: the OAuth refresh gap also affects the CLI surface (tracked separately as issue #2022, explicitly scoped out of this PR as a different surface/fix). Since #2021 is Inspector-only and #2022 is filed as its own follow-up rather than silently left uncovered, this does not block — the capability isn't claimed as fixed cross-surface here.

## Blocking findings
None. 0 [BLOCKING] items across all three lenses.

## Merge recommendation
Ready to merge. All required gates pass, no blocking findings. **Per operator autonomy guardrail, merge itself is held for operator action — a checkpoint_brief is filed rather than an autonomous merge.**

## Verdict
**APPROVE**

📎 Neotoma: [issue#1997](https://neotoma.markmhendrickson.com/entities/ent_f69dd0c40686f8b4d26bea5f) · [issue#2021](https://neotoma.markmhendrickson.com/entities/ent_3a6f4400885a53e91e8ff2ee)

neotoma-agent (approved):
All Neotoma bookkeeping complete. Here's the summary.

## Aggregated verdict — PR #2021

**Gate inheritance (parent issue #1997):** pm/ux/arch `not_required`, impl `signed_off` — clears.

**Panel roll-up:**
- **pm** (Pavo) — no blocking findings. Scope matches #1997 exactly, effect-verified.
- **arch** (Waxwing) — no structural changes required. One non-blocking fast-follow flagged (undeclared `refresh_token` grant in `openapi.yaml`).
- **qa** (Phoenicurus) — COMMENT, no blocking gaps. Strong regression suite for the actual bug; two non-blocking coverage gaps in secondary branches noted.

**Blocking findings: 0.**

**Verdict: APPROVE.** Posted as PR comment: https://github.com/markmhendrickson/neotoma/pull/2021#issuecomment-5101031681

**Neotoma:** `pr_review` gate signed off on issue #1997 (`ent_f69dd0c40686f8b4d26bea5f`), owner_history updated, `plan_contribution` filed (`ent_3cedfa1396f699b74e76db5a`) noting merge is held.

**Per the autonomy guardrail, I did not merge.** PR is `MERGEABLE` at head SHA `c828aa6bfa107f16016fb6cec9e641985099e76d` — awaiting operator action to execute `gh pr merge --squash`.

[vanellus] merge_decision: APPROVE — pm/arch/qa panel all clear, 0 blocking findings, gate inheritance passes on #1997. Verdict comment posted to PR #2021. pr_review gate signed off in Neotoma. Merge NOT executed — held for operator per autonomy guardrail (checkpoint: merge PR #2021 at head SHA c828aa6bfa107f16016fb6cec9e641985099e76d).

## 🧠 Neotoma — [Vanellus: aggregate PR #2021 review panel](https://neotoma.markmhendrickson.com/inspector/conversations/ent_30fc0a4ce6634bf2a1d53938)

**Created (3)**
- 📝 plan_contribution: [PR #2021 pr_review sign-off](https://neotoma.markmhendrickson.com/inspector/entities/ent_3cedfa1396f699b74e76db5a)
- 💬 conversation_message (user): [ent_175bdd5c6527131f177766a7](https://neotoma.markmhendrickson.com/inspector/entities/ent_175bdd5c6527131f177766a7)
- 💬 conversation_message (assistant): [ent_b4e7ee8ada6818d45c7b4cff](https://neotoma.markmhendrickson.com/inspector/entities/ent_b4e7ee8ada6818d45c7b4cff)

**Updated (1)**
- 🐛 issue #1997: [gate_status + owner_history](https://neotoma.markmhendrickson.com/inspector/entities/ent_f69dd0c40686f8b4d26bea5f)

**Retrieved (3)**
- 🐛 issue #1997 (parent): [ent_f69dd0c40686f8b4d26bea5f](https://neotoma.markmhendrickson.com/inspector/entities/ent_f69dd0c40686f8b4d26bea5f)
- 🐛 issue #2021 (PR): [ent_3a6f4400885a53e91e8ff2ee](https://neotoma.markmhendrickson.com/inspector/entities/ent_3a6f4400885a53e91e8ff2ee)
- 🐛 issue #2022 (CLI parity follow-up, referenced not linked): ent_532c03e67b1fefddb8427e90

PR #1804, opened 2026-06-25
chore(tests): regenerate stale automated test catalog on main

markmhendrickson:
## Problem

`main`'s `docs/testing/automated_test_catalog.md` has been **stale since #1797**, which merged `tests/integration/sandbox_seed_token_bypass.test.ts` without regenerating the catalog. The `baseline` CI job runs `validate:test-catalog --check`, so **every PR opened against main fails `baseline`** on this check — including fixture-only or docs-only PRs that touch no tests.

Confirmed by running `npm run validate:test-catalog` on a clean `origin/main` checkout: ❌ stale.

## Fix

Regenerated the catalog (`npm run generate:test-catalog`). The drift is exactly:
- `tests/integration/sandbox_seed_token_bypass.test.ts` added to the integration file list
- repo-wide total 501→502, backend 467→468, integration count 141→142

Generator output is deterministic (stable across repeated runs); `validate:test-catalog` passes after the change.

## Impact

Unblocks `baseline` for all open PRs once they pick this up (rebase / merge main). Standalone and low-risk — a single generated-doc file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2025, opened 2026-07-27
chore(manifest): regenerate stale capability_manifest.json

markmhendrickson:
## What

Regenerates `src/shared/capability_manifest.json` — a 3-line addition for `query_contacts_at_company` (added in v0.19.0) that was never committed when the tool landed.

## Why now

The **`baseline` CI lane is red on `main` itself** (not on any PR's diff): `validate:capability-manifest --check` exits 1 because the manifest is stale. This blocks the swarm's review pipeline for every open PR — the dispatcher routes each one to a CI-fix round that can't succeed, because the failure isn't in the PR.

Verified: `baseline` fails on unmodified `origin/main`; `npm run generate:capability-manifest` produces exactly the one missing entry and nothing else.

(The pre-existing `fix/test-catalog-stale-main` branch / closed PR #1804 is not a usable fix — it's diverged by -30k lines and deletes hundreds of files. This is the minimal correct regen instead.)

## Effect

No behavior change. Turns `baseline` green on main, unblocking CI (and therefore the swarm review gate) on the open PR queue.

### DIFF
--- a/src/shared/capability_manifest.json
+++ b/src/shared/capability_manifest.json
@@ -126,6 +126,9 @@
     "publish_rendered_page": {
       "addedInVersion": "v0.16.0"
     },
+    "query_contacts_at_company": {
+      "addedInVersion": "v0.19.0"
+    },
     "register_schema": {
       "addedInVersion": "v0.4.3"
     },


neotoma-agent:

PR #2087, opened 2026-08-04
fix(mcp): discover the live port when port files are stale

### LINKED ISSUE #2088: MCP shim dies with a generic "could not attach" when the port file is stale, instead of discovering the live port
## Summary

When every candidate port file fails its TCP probe, the MCP shim's downstream resolver falls back to a **hardcoded** port for the profile (`3180` for prod, `3080` for dev). If the server is bound elsewhere, that fallback is dead too — the shim exits immediately and the only user-visible symptom is:

```
Could not attach to MCP server mcpsrv_neotoma
```

Nothing in that message indicates the port was wrong.

## Encountered in practice

A prod port file (`.dev-serve/local_http_port_prod`) last written weeks earlier still said `3180`, while the server was bound to `9180`. The resolver probed `3180`, failed, fell back to `3180`, and died.

Because the failure surfaces as a generic attach error, three rounds of diagnosis went into MCP configuration that was already correct. The actual cause was a one-line stale file.

## Why the current behaviour is wrong

The resolver already TCP-probes before trusting a port file — so it knows how to tell live from dead. It just doesn't use that capability when the files are exhausted; it falls through to a constant that has no better claim to being correct than the file it just rejected.

## Suggested fix

Add a discovery step before the hardcoded fallback: probe the canonical ports for the profile, use the first that accepts a connection, and rewrite the port file so the next start resolves on the fast path. Keep the hardcoded fallback as a genuine last resort.

Related: consider whether a failed attach should log the attempted downstream URL at a level the harness surfaces, so the next occurrence is self-evident rather than requiring the shim to be run by hand.

## Swarm specification

### Product / Scope (PM)

**Problem statement:** When all candidate port files fail their TCP probe, the MCP shim's resolver falls back to a *hardcoded* per-profile port (`3180` prod / `3080` dev) instead of applying the same live-probe capability it already uses elsewhere. If the server is bound to a non-canonical port, the fallback is dead on arrival and the shim exits with a generic `Could not attach to MCP server mcpsrv_neotoma` — giving no signal that the root cause is a stale port file. This has already cost multiple rounds of misdirected diagnosis against MCP config that was correct.

### In scope
- Add a discovery step, run before the hardcoded fallback is used: probe the profile's canonical port set, take the first port that accepts a TCP connection.
- On successful discovery, rewrite the stale port file so the next shim start resolves on the fast path (no repeated discovery cost).
- Retain the hardcoded constant as a genuine last-resort fallback only, after discovery has been attempted and failed.
- On a final, unrecoverable attach failure, log the attempted downstream URL(s) at a level the harness surfaces to the user/operator — not buried in a stream normally-discarded.

### Out of scope
- Changing the port file's format, on-disk location, or the set of profiles (dev/prod) it applies to.
- Any MCP tool, CLI command, schema field, error code, relationship type, or other request/response contract change — this is shim-internal resolver behavior only.
- General-purpose network diagnostics or health-check tooling beyond this resolver's own fallback path.
- Expanding port discovery beyond the already-known canonical port(s) for the active profile (e.g. no arbitrary port-range scanning).

### Acceptance criteria
- [ ] When every port-file candidate fails its probe, the resolver probes the profile's canonical port(s) before falling back to the hardcoded constant.
- [ ] If discovery finds a live port, the shim attaches on that port **and** the stale port file is rewritten with the discovered value.
- [ ] The hardcoded fallback is used only when both port-file candidates and canonical-port discovery fail.
- [ ] A test asserts the reported **effect**: given a stale/wrong port file and a server actually listening on a different canonical port, the shim successfully attaches (not merely that discovery logic is invoked or a value is returned).
- [ ] A test covers the genuine last-resort path: all discovery exhausted, hardcoded fallback used, failure surfaced.
- [ ] On final attach failure, the attempted downstream URL(s) are logged at a level visible in the harness's normal output (not just verbose/debug-only).
- [ ] Cross-surface parity: if the shim serves both dev and prod profiles through this same resolver code path, both profiles are covered by the fix and by tests (this is a single-surface capability — the MCP shim's own port resolver — so no additional MCP/CLI/HTTP surface parity applies here).

### Priority & sequencing
Reliability/DX bug with outsized diagnostic cost (three rounds of misdirected debugging from one stale file) relative to fix size. No downstream feature depends on this, but it directly reduces future MCP-attach incident triage time — sequence as a standalone fast-turnaround fix, not blocking on other work.

### Fast-path classification
Checked against the interface-surface predicate: this change touches only the shim's internal port-resolution logic — no `SchemaDefinition` field, no `openapi.yaml`/request-response contract, no new/changed MCP tool or CLI command, no relationship type, no agent-instruction surface. It remains a **localized implementation bug** — `label:bug` fast path applies, routed directly to `cicada` (impl); arch gate not required.

### Engineering

**Key finding:** this resolver already has an uncommitted local implementation of the discovery + self-heal + logging behavior the PM/issue asks for, at `scripts/lib/neotoma_mcp_resolve_downstream_url.sh` lines 99-136 (`git diff` shows it added against `HEAD`, commit `33c0840f9`). This is not a from-scratch design — it is **completing, testing, and landing** an in-progress fix, and cleaning up a stray backup file left alongside it.

#### Files to touch

- `scripts/lib/neotoma_mcp_resolve_downstream_url.sh` — contains the target logic (`neotoma_mcp_resolve_mcp_proxy_downstream`). Currently uncommitted (`git status` shows ` M`). Review the existing diff against the acceptance criteria (below) rather than rewriting from scratch.
- `scripts/lib/neotoma_mcp_resolve_downstream_url.sh.bak-20260804` — untracked backup file sitting next to the target script. Delete it; it must not ship in the PR (not a build artifact, not referenced by any caller).
- `scripts/run_neotoma_mcp_signed_stdio_dev_shim.sh` and `scripts/run_neotoma_mcp_unsigned_stdio_dev_shim.sh` — the two callers of `neotoma_mcp_resolve_mcp_proxy_downstream`. No code change expected here, but confirm both source the lib script the same way so both dev and prod profiles inherit the fix (PM acceptance criterion: cross-profile parity).
- New test file: `tests/unit/neotoma_mcp_resolve_downstream_url.test.ts` (or `.sh`-runner equivalent if the repo's shell-script test convention differs — check `tests/unit/` for an existing pattern of testing a `.sh` lib before deciding). No test file currently exists for this script (confirmed via search of `tests/`).

#### Current (uncommitted) implementation — what it already does

1. Reads the profile's port-file candidate(s) (`local_http_port_prod` for prod; `local_http_port_dev` then legacy `local_http_port` for dev) and TCP-probes each (existing behavior, unchanged).
2. **New:** if none resolve, probes a canonical-port candidate list per profile — `9180 3180` for prod, `9080 3080` for dev, `3080 3180 9080 9180` for the profile-less legacy branch — in that order, first-listen wins.
3. **New:** on a successful discovery probe, sets `MCP_PROXY_DOWNSTREAM_URL` to the discovered port, logs `"port files stale/unreachable; DISCOVERED live port ${_port} by probe"` to stderr, and writes the discovered port back to the first port-file path for the profile (self-heal) when its directory is writable.
4. **Unchanged:** only when discovery also fails does it fall through to the hardcoded `_fallback` (`3180`/`3080`), now logging `"no reachable port from port files or discovery"` (message already updated to mention discovery).

#### Gaps against the PM acceptance criteria — what this PR still needs to add

1. **No automated tests exist.** Every acceptance criterion in the PM section is currently unverified by CI. This is the primary remaining work:
   - A test that stubs/starts a TCP listener on a non-canonical port, writes a stale/wrong value into the profile's port file, runs the resolver, and asserts `MCP_PROXY_DOWNSTREAM_URL` resolves to the actually-listening canonical port (the **effect** — successful attach path — not just that the discovery branch executed).
   - A test asserting the port file was rewritten with the discovered value after self-heal.
   - A test covering the genuine last-resort path: no port file candidate reachable AND no canonical port reachable → asserts the hardcoded fallback URL is used and the failure is logged with the attempted URL(s).
   - Cross-profile parity: run the above for both `prod` and `dev` profiles (distinct candidate lists, distinct port-file paths), since this is a single shared shell function parameterized by `_profile`.
   - Since this is a bash function, not a TS module, decide the harness: either (a) a lightweight TS test that shells out to `bash -c 'source scripts/lib/neotoma_mcp_resolve_downstream_url.sh; ...'` with env vars set and a real `net.createServer` listener started in Node before invoking, asserting on the exported `MCP_PROXY_DOWNSTREAM_URL` printed to stdout by the sourced script, or (b) a `.bats`/shell-native test runner if one already exists in `tests/`. Check `tests/unit/` and `package.json` test scripts for an existing shell-test convention before introducing a new one; prefer consistency with whatever already tests other `scripts/lib/*.sh` files, if any such precedent exists (none found in the initial search — flag this as a build-step decision if truly novel).

2. **Log visibility level.** The PM criterion requires the attempted downstream URL(s) to be logged "at a level visible in the harness's normal output (not just verbose/debug-only)." The current diff writes plain `echo ... >&2` unconditionally (not gated behind a debug env var), which already satisfies "not debug-only" — confirm this by checking whether `run_neotoma_mcp_signed_stdio_dev_shim.sh` / `run_neotoma_mcp_unsigned_stdio_dev_shim.sh` redirect or suppress the lib's stderr anywhere downstream (e.g. piping to a log file the harness doesn't surface). If stderr from the shim subprocess is already surfaced to the MCP client/harness UI as-is, no further change is needed here beyond verifying it; if it's being swallowed, that's the actual fix required by this criterion.

3. **Stray backup file.** Delete `scripts/lib/neotoma_mcp_resolve_downstream_url.sh.bak-20260804` in this PR — it is untracked, not a generated/derived artifact, and not part of the intended diff.

#### Data / contract changes

None. This is shim-internal shell logic only — no `SchemaDefinition`, `openapi.yaml`, MCP tool, CLI command, or entity/relationship change (confirmed by PM's fast-path classification; no arch gate).

#### Build-step checklist

1. `git diff -- scripts/lib/neotoma_mcp_resolve_downstream_url.sh` — review the existing uncommitted change against the acceptance criteria above; keep it as the base, do not re-derive the discovery logic from scratch.
2. Delete `scripts/lib/neotoma_mcp_resolve_downstream_url.sh.bak-20260804`.
3. Decide and implement the test harness for a `.sh` lib function (see gap #1) — write tests for: (a) discovery success + port-file self-heal, both profiles; (b) last-resort hardcoded fallback path with logged attempted URL(s), both profiles; (c) existing port-file-hit fast path still passes (regression covering the pre-existing, unchanged behavior at lines 55-98).
4. Verify stderr from the discovery/fallback log lines actually reaches the harness's visible output through both shim entrypoint scripts (gap #2); adjust only if it's currently being swallowed — do not add a new logging mechanism if plain stderr already surfaces.
5. Run the new tests locally; run the full existing suite touching MCP stdio shim behavior (`tests/unit/mcp_dev_shim.test.ts`, `tests/contract/mcp_stdio_output_safety.test.ts`, `tests/integration/mcp_stdio_attribution.test.ts`) to confirm no regression, even though this resolver script sits outside those files' direct coverage.
5. `npm run generate:test-catalog` and `npm run validate:test-catalog` if the new test file lands under a catalogued directory (per repo test-catalog rule).
6. Self-review diff (`/code-review`) before `gh pr create`; confirm the PR body explicitly calls out that this completes a pre-existing uncommitted change plus adds its missing test coverage, so the reviewer isn't surprised by a large "new" diff that is mostly already-written logic.

### QA / Test Plan

**Pre-check against Eng's plan (verified against `~/repos/neotoma-wt-shim-selfheal`, branch `fix/mcp-shim-self-healing-port-discovery`):** the discovery/self-heal logic described by Eng is **already committed** as `d593b9a60` ("fix(mcp): discover the live port when port files are stale") — not uncommitted as Eng's section states. The referenced backup file `scripts/lib/neotoma_mcp_resolve_downstream_url.sh.bak-20260804` **does not exist** in this worktree — nothing to delete here (Eng should confirm against their own working copy; this may be worktree-specific residue). No shell-test precedent exists anywhere under `tests/` (confirmed by search) — the test-harness decision below is genuinely novel, not a convention lookup.

**Harness decision:** no `.bats`/shell-native runner exists in this repo. Use a Vitest test that shells out via Node's `child_process` (`execFileSync`/`spawnSync` with `bash -c`), sourcing the lib and invoking `neotoma_mcp_resolve_mcp_proxy_downstream`, with a real `net.createServer` listener started in-process before invocation. Assert on `MCP_PROXY_DOWNSTREAM_URL` captured from the subshell's env (e.g. `bash -c 'source ...; neotoma_mcp_resolve_mcp_proxy_downstream test; echo "$MCP_PROXY_DOWNSTREAM_URL"'`). New file: `tests/unit/neotoma_mcp_resolve_downstream_url.test.ts`.

#### Regression test (pre-existing, unchanged behavior — lines 55-98)
- [ ] **Port-file fast path still resolves.** Write a valid port file for `profile=prod` pointing at a live listener; assert `MCP_PROXY_DOWNSTREAM_URL` resolves to that port and neither the discovery branch nor the hardcoded fallback log lines appear in stderr. Same case for `profile=dev`. Guards against the new discovery block accidentally short-circuiting or double-running the existing probe loop.

#### New-branch edge cases (discovery + self-heal, lines 99-136)
- [ ] **Stale port file, live server on canonical discovery port — prod.** Port file `local_http_port_prod` contains `3180` (nothing listening); real listener bound to `9180`. Run resolver. Assert: (a) `MCP_PROXY_DOWNSTREAM_URL == http://127.0.0.1:9180/mcp` (the *effect* — successful attach target — not just that discovery ran), (b) stderr contains `DISCOVERED live port 9180 by probe`, (c) the port file on disk now contains `9180` (self-heal write verified by reading the file after the call, not just the log line).
- [ ] **Stale port file, live server on canonical discovery port — dev.** Same shape: `local_http_port_dev` says `3080` (dead), listener on `9080`. Assert resolution to `9080`, log line, and self-healed file content. Covers PM's cross-profile-parity criterion.
- [ ] **Discovery candidate ordering.** Prod candidates are `9180 3180` (first-listen-wins). Start listeners on *both* `9180` and `3180` with a stale port file; assert the resolver picks `9180` (the first candidate), not `3180`, proving order is respected rather than "any reachable port."
- [ ] **Self-heal write skipped when target directory is not writable.** Make the port-file's directory read-only (`chmod 555` on a temp dir, restore after), stale file, live discovery port. Assert: `MCP_PROXY_DOWNSTREAM_URL` still resolves correctly (discovery succeeds independent of the write), but no `self-healed` log line appears and the on-disk file is unchanged — proving the write failure is swallowed (`2>/dev/null`) without breaking resolution. This exercises the untested `[ -w "$(dirname "$_pf")" ]` guard.
- [ ] **No port files exist at all (not just stale) + live discovery port.** Omit the port file entirely for the profile; listener live on a canonical port. Assert discovery still fires and resolves — the `_resolved!=1` branch is reached the same way whether the cause is "file missing" or "file present but dead."
- [ ] **Legacy/no-profile branch candidate list.** With `_profile` unset, candidates are `3080 3180 9080 9180`. Listener on `9080` only; assert resolution reaches it (proves the profile-less 4-candidate list is wired, not just the prod/dev 2-candidate lists — this branch is easy to leave untested since PM/Eng framing focuses on prod/dev).

#### Last-resort hardcoded fallback path
- [ ] **All discovery exhausted — prod.** Stale/missing port file AND nothing listening on `9180` or `3180`. Assert: `MCP_PROXY_DOWNSTREAM_URL == http://127.0.0.1:3180/mcp` (the hardcoded constant), stderr contains `no reachable port from port files or discovery (profile=prod)`, and no `self-healed` line. This is the genuine last-resort path PM's acceptance criteria explicitly require covered.
- [ ] **All discovery exhausted — dev.** Same shape for `3080` fallback and `profile=dev` in the log line.
- [ ] **Attempted URL(s) visible in normal (non-debug) output.** Assert the fallback log line and/or discovery log lines are written to stderr unconditionally — i.e. the test does **not** set any verbose/debug env var and the lines still appear. This directly verifies PM's "logged at a level the harness surfaces, not verbose/debug-only" criterion at the resolver-script level.
- [ ] **Stderr propagation through both shim entrypoints.** Since `neotoma_mcp_resolve_downstream_url.sh` only proves the log line is written to its own stderr, add a second-layer check: grep `run_neotoma_mcp_signed_stdio_dev_shim.sh` and `run_neotoma_mcp_unsigned_stdio_dev_shim.sh` for any `2>`, `>>`, or `| tee`/`| grep -v` redirection that could swallow the lib's stderr before it reaches the harness. If either script redirects stderr away from the process's own stderr (e.g. to a log file the MCP client never reads), that is a `[BLOCKING]` gap against PM's visibility criterion — file it, do not silently note it as "confirmed" without checking both files.

#### Out of scope for this PR's tests (per PM/Eng scope)
- No test of the upstream MCP client/harness's own `Could not attach to MCP server` message — that string is not sourced in this repo (confirmed via search); it belongs to the harness consuming the shim, not this resolver.
- No arbitrary port-range scanning tests — PM explicitly scoped discovery to the known canonical port(s) only.

#### CI wiring
- [ ] New test file lands under `tests/unit/` and is picked up by `npm run test:unit` (already globs `tests/unit`) — confirm it runs green locally before opening the PR.
- [ ] Run `npm run generate:test-catalog && npm run validate:test-catalog` since the new file is under a catalogued directory.
- [ ] Run the three named regression suites Eng flagged (`tests/unit/mcp_dev_shim.test.ts`, `tests/contract/mcp_stdio_output_safety.test.ts`, `tests/integration/mcp_stdio_attribution.test.ts`) to confirm no incidental regression from the already-landed commit.

#### Sign-off condition
Sign off `qa` only once every checklist item above is a committed, passing test in `tests/unit/neotoma_mcp_resolve_downstream_url.test.ts` and the stderr-propagation check on both shim entrypoint scripts has been performed and recorded (pass, or filed as blocking if either redirects stderr away from the harness).

markmhendrickson:
## Problem

When every candidate port file failed its TCP probe, the resolver fell back to a **hardcoded** port for the profile (`3180` for prod, `3080` for dev). If the server was bound elsewhere, that fallback was dead too — the shim exited immediately and the user saw only `Could not attach to MCP server`, with nothing indicating the port was the problem.

Encountered in practice: a prod port file last written weeks earlier said `3180` while the server bound `9180`. The resolver probed `3180`, failed, fell back to `3180`, and died. Three rounds of diagnosis went into config that was already correct.

## Change

Adds a discovery step before the hardcoded fallback:

1. Probe the canonical ports for the profile (`9180 3180` for prod, `9080 3080` for dev).
2. Use the first that accepts a connection.
3. Rewrite the port file so the next start resolves on the fast path.

The hardcoded fallback is retained as a last resort; its log line now notes discovery was also attempted.

## Verification

Sabotaged `.dev-serve/local_http_port_prod` to `9999` (nothing listening) and started the shim:

```
[neotoma-mcp-signed-shim] port files stale/unreachable; DISCOVERED live port 9180 by probe
[neotoma-mcp-signed-shim] self-healed .../local_http_port_prod -> 9180
[neotoma-mcp-proxy] Starting proxy: downstream=http://127.0.0.1:9180/mcp
{"result":{"protocolVersion":"2025-11-25","capabilities":{"tools":...
```

Handshake completed and the port file was left corrected. `bash -n` clean.

Closes #2088

### DIFF
--- a/scripts/lib/neotoma_mcp_resolve_downstream_url.sh
+++ b/scripts/lib/neotoma_mcp_resolve_downstream_url.sh
@@ -96,8 +96,46 @@ NODE
         break
       fi
     done
+    # Self-heal: when no port file yields a reachable port, DISCOVER the port the
+    # server is actually listening on rather than trusting a stale hardcoded
+    # fallback. A port file last written weeks ago (e.g. 3180) while the server
+    # binds elsewhere (e.g. 9180) otherwise produces a silent, confusing
+    # "Could not attach to MCP server" with no indication the port was wrong.
+    # Probes the canonical ports for each profile, then rewrites the port file
+    # so the next start resolves on the fast path.
     if [ "$_resolved" != "1" ]; then
-      echo "[${_label}] NEOTOMA_MCP_USE_LOCAL_PORT_FILE: no reachable port from port files (profile=${_profile:-legacy}); falling back to ${_fallback}" >&2
+      _candidates=""
+      case "$_profile" in
+        prod) _candidates="9180 3180" ;;
+        dev)  _candidates="9080 3080" ;;
+        *)    _candidates="3080 3180 9080 9180" ;;
+      esac
+      for _port in $_candidates; do
+        if _NEOTOMA_SHIM_PROBE_HOST=127.0.0.1 \
+          _NEOTOMA_SHIM_PROBE_PORT="$_port" \
+          node -e '
+const net=require("node:net");
+const port=Number(process.env._NEOTOMA_SHIM_PROBE_PORT);
+const s=net.createConnection({host:"127.0.0.1",port},()=>{s.end();process.exit(0)});
+s.setTimeout(1200);
+s.on("timeout",()=>{try{s.destroy()}catch{};process.exit(1)});
+s.on("error",()=>process.exit(1));
+'
+        then
+          export MCP_PROXY_DOWNSTREAM_URL="http://127.0.0.1:${_port}/mcp"
+          echo "[${_label}] port files stale/unreachable; DISCOVERED live port ${_port} by probe" >&2
+          _resolved=1
+          _pf="$(echo $_port_files | awk '{print $1}')"
+          if [ -n "$_pf" ] && [ -w "$(dirname "$_pf")" ]; then
+            printf '%s' "$_port" > "$_pf" 2>/dev/null \
+              && echo "[${_label}] self-healed ${_pf} -> ${_port}" >&2
+          fi
+          break
+        fi
+      done
+    fi
+    if [ "$_resolved" != "1" ]; then
+      echo "[${_label}] NEOTOMA_MCP_USE_LOCAL_PORT_FILE: no reachable port from port files or discovery (profile=${_profile:-legacy}); falling back to ${_fallback}" >&2
       export MCP_PROXY_DOWNSTREAM_URL="$_fallback"
     fi
   else


neotoma-agent:

neotoma-agent:
**BLOCKED**

No parent issue found for this PR — cannot evaluate gate inheritance.

- PR body contains no `closes #N` / `fixes #N` reference.
- GitHub `closingIssuesReferences` for this PR is empty.
- Searched open/closed issues and Neotoma `issue` entities in this repo for a topical match (stale port file / hardcoded fallback / mcp-shim discovery) — no candidate found.

Since there is no parent issue entity, there is no `gate_status` to retrieve, so the LEGACY-ISSUE merge path does not apply either (that requires an existing issue entity with an absent `gate_status` key — there is none here). Pre-impl gates (`pm`, `ux`, `arch`) are therefore unsigned by definition:

| Gate | Owner |
|---|---|
| pm | Pavo |
| ux | Accipiter |
| arch | Waxwing |

**Resolution paths:**
1. Link a parent issue — edit the PR body to include `closes #N` (or `fixes #N`) pointing at the issue this work traces to, then re-trigger this check.
2. If no issue exists yet, file one and route it through Lanius triage so pm/ux/arch gates can be scoped and signed off (or fast-pathed) normally.
3. Operator override: **@markmhendrickson** may comment `/confirm-gates-clear` to waive all unsigned pre-impl gates and re-trigger the PR pipeline. No other commenter can clear gates.

Normal resolution: the gate owner signs off via Neotoma (`gate_status.<gate> → signed_off`), or the operator waives it.

---
📎 Neotoma: no linked `issue` entity found for this PR

Issue #2226, opened 2026-08-26
Release-triggered Pages deploy has failed on every release since v0.21.0; no notification, no record, and release_result has no deploy fields


markmhendrickson:
## Observed

Every release-triggered GitHub Pages deploy has failed since v0.21.0, and no one was told.

| run | ref | result |
|---|---|---|
| 08-26 | release/v0.22.0 | failure |
| 08-10 | release/v0.21.5 | failure |
| 08-07 | release/v0.21.4 | failure |
| 08-05 | release/v0.21.2 | failure |
| 07-30 | release/v0.21.0 | failure |
| 07-29 | push/main | success |

Five for five on releases; every push-to-main run succeeds.

## Cause

Annotation on the failing run:

```
Tag "v0.22.0" is not allowed to deploy to github-pages due to environment protection rules.
```

The `github-pages` environment uses a custom branch policy allowing only `main` and `dev`. A `release: published` run carries a **tag** ref, which no branch policy can match, so the job is rejected before any step executes (the jobs API returns an empty `steps` array).

The site is not stale — push-to-main keeps it current — so the failing job is either redundant or the policy needs a tag pattern. Either way it has been failing silently for four weeks.

## Why it was not surfaced

`publish.py` reports success and sets `release_result.status = published` the moment `gh release edit --draft=false` returns. That is the same instant the four `release: published` workflows *begin*. It waits for `npm-publish.yml` specifically (`publish.py:655`) but observes none of:

- `deploy-sandbox.yml`
- `deploy-client-instance.yml`
- `deploy-pages-site.yml`
- `upgrade_guidance_bot.yml`

The operator receives a ✅ Telegram naming npm only. A downstream deploy can fail with no notification and no record.

## Nothing records deploy state

The `release_result` schema has 13 declared fields — `version`, `status`, `branch`, `release_url`, `rc_branch`, `rc_pr_url`, `remote`, `system`, `reason`, `workflow`, `latest_main_push_run`, `latest_main_push_title`, `latest_release_branch_run_found` — and **not one concerns deployment**. There is no field for which targets deployed, at what version, or with what result.

So even a correct pipeline could not answer "did this release reach the client instance?" from Neotoma. It is not a missing write; there is nowhere to write it.

## Suggested fix

1. Fix or remove the Pages release trigger (add a tag pattern to the environment policy, or drop `release:` from the trigger if push-to-main is sufficient).
[...]

PR #2251, opened 2026-08-27
feat(ci): deploy the operator instance on release — closes #2147

### LINKED ISSUE #2147: The operator's own instance has no release-triggered deploy — it silently lags every release
## The gap

Every other deploy target updates itself when a release publishes. The operator's own instance does not:

| Target | On `release: published` |
|---|---|
| Sandbox | `deploy-sandbox.yml` ✅ |
| Client instance | `deploy-client-instance.yml` ✅ |
| Pages site | `deploy-pages-site.yml` (failing — see #2146) |
| **Operator instance** | **nothing** |

No workflow references it. `grep -rl "neotoma-markmhendrickson" .github/workflows/` returns nothing, and there is no doc describing how it is deployed. Its Fly release history is entirely manual, from the operator's laptop:

```
v14  complete  Release  markmhendrickson@gmail.com  Aug 9 08:39
v13  complete  Release  markmhendrickson@gmail.com  Aug 9 08:29
v12  complete  Release  markmhendrickson@gmail.com  Aug 7 08:24
```

## Observed consequence

As of 2026-08-10, after v0.21.5 published: the client instance serves `0.21.5`; the operator instance serves `0.21.4`. It is running code one release behind, including the standing-rules fix (#2131) and the concurrency race fixed during v0.21.5 preparation.

That is the *smallest* version of the problem. Nothing enforces the lag being one release — historically it has been whatever the operator last remembered to deploy.

## Why this is worth closing rather than living with

The operator instance is where the real graph lives — 20GB volume, production data, the instance every session and daemon actually writes to. It is the **least** appropriate target to be running the least-verified build.

It also inverts the usual risk ordering: the sandbox (disposable) and the client instance (a third party's) both get the new build automatically, while the one holding the operator's own history waits for someone to remember.

## The argument against, and why it does not hold

Manual deploys are defensible when a target carries real data and a bad release could damage it — you want a human deciding when. But that reasoning already applies to the **client instance**, which auto-deploys today and belongs to someone else. If auto-deploy is safe enough for a client's graph, it is safe enough for the operator's.

The counter-argument that *does* have force: the operator instance has a persistent volume and no automated backup-before-deploy step in the existing workflows. That is an argument for **adding a snapshot step**, not for leaving the deploy manual — a manual deploy does not take a backup either.

## Requested

1. Add the operator instance to release-triggered deploys, mirroring `deploy-client-instance.yml` — app name and host from repository secrets, skipping cleanly when unset so forks are unaffected.
2. Take a volume snapshot before deploying, given the persistent data. Worth doing for the client instance too.
3. Verify post-deploy the way the client workflow does: check `/health` reports the new version rather than assuming the deploy succeeded.

## Verification

After the next release: `curl -s https://<operator-host>/health` reports the version just published, without anyone running `flyctl deploy` by hand.

## Swarm specification

### Product / Scope (PM)

PM gate signed off and posted. Summary:

- **Spec section written**: `Product / Scope (PM)` — problem statement, in/out-of-scope, 9-item acceptance criteria checklist (including the mandatory effect-verified-fixed and cross-surface-parity gate items, with parity marked N/A and reasoned since this is a single-surface CI/CD workflow), priority/sequencing, and two non-blocking open questions for the operator.
- **Key finding**: precedent already exists and is in production — issue #2012 (`deploy-client-instance.yml`) is the exact template to mirror, so this is adaptation, not new design. Low risk, high leverage, no dependencies blocking or blocked by it.
- **Interface-surface check**: ran the predicate explicitly — this touches only a `.github/workflows/*.yml` file, none of the schema/OpenAPI/tool-definition/relationship-type/agent-instruction surfaces. No arch-gate override triggered (arch still runs next in the normal pm→ux→arch sequence for this feature-type issue, not via a bug/security fast path).
- **Gate handoff**: `gate_status.pm` → `signed_off`, `current_owner` → `accipiter`, `owner_history` appended, `plan_contribution` (sign_off) stored, verdict comment posted on GitHub with the required attribution header and Neotoma backlink.

🧠 Neotoma — [neotoma#2147](https://neotoma.markmhendrickson.com/inspector/entities/ent_609095b65db23a586466e49d)
- ✏️ Updated (3): gate_status, current_owner, owner_history on [issue](https://neotoma.markmhendrickson.com/inspector/entities/ent_609095b65db23a586466e49d)
- ➕ Created (1): [plan_contribution](https://neotoma.markmhendrickson.com/inspector/entities/ent_49d38921416ad4358fd3be8b) sign-off record
- 🔎 Retrieved (2): precedent [issue #2012](https://neotoma.markmhendrickson.com/inspector/entities/ent_81478aff39d0363f53ce24cf), [agent_policy](https://neotoma.markmhendrickson.com/inspector/entities/ent_5456a8a2224d8211ef33749c) (swarm_dispatch monitoring)

[pavo] acceptance_criteria: PM scope signed off for neotoma#2147 — mirror `deploy-client-instance.yml` (#2012 precedent) for the operator's own Fly instance on `release: published`, add a pre-deploy volume snapshot, gate on post-deploy `/health` version verification. 9-item acceptance checklist posted to issue body Product/Scope (PM) section covering: secret-sourced app/host, clean skip on unset secret, pre-deploy snapshot, effect-verified `/health` version check (not just deploy-exit-0), failure notification, doc gap closure, end-to-end post-release proof, and CI-appropriate test-verification method. Cross-surface-parity gate marked N/A (single-surface CI/CD workflow) and stated explicitly rather than silently skipped. No interface-surface override triggered — pure CI/CD change, routed via normal pm→ux→arch sequence. `current_owner` → accipiter.

### Design / UX (Accipiter)

**User goal**: the operator wants their own instance to run the latest release without remembering to deploy it — and when the pipeline can't do that safely, they want to know exactly why and what to do, not silently lag.

**Surface**: no new UI. The surface *is* the CI/CD experience — workflow name/logs in GitHub Actions, the GitHub Actions summary tab, and (per PM's failure-notification item) an alert channel. Audience is the operator only, acting as both "developer" and "end user" of this workflow — so the CLI/CI conventions of the user profile apply directly: fast completion, transparency, reversibility, zero tolerance for silent failure.

#### Naming
- Workflow file: `deploy-operator-instance.yml` — mirrors `deploy-client-instance.yml` exactly (PM's stated precedent), sits alongside it in `.github/workflows/`, so anyone scanning the directory sees the three deploy targets as siblings, not one of them oddly named or nested.
- Job/step names inside the workflow should reuse the client workflow's vocabulary verbatim (`deploy`, `snapshot`, `verify-health`, or whatever it currently uses) — do not invent new terminology for equivalent steps. Consistency here is what lets the operator read a red X in the Actions tab and immediately know which stage failed without opening logs.
- Secrets: `OPERATOR_APP_NAME` / `OPERATOR_HOST` (or match whatever naming pattern `CLIENT_APP_NAME` / `CLIENT_HOST` already use in #2012 — same pattern, different prefix). Do not diverge from the existing secret-naming convention.

#### Flow — release publishes → operator instance updated

1. **Trigger**: `release: published` fires (existing, unchanged).
2. **Skip check**: workflow checks `OPERATOR_APP_NAME`/`OPERATOR_HOST` secrets are set.
   - **Empty state**: unset (fork, or operator secrets not configured) → job exits with a clearly-labeled **skipped** status (not a red failure) and a one-line log: `Operator instance secrets not configured — skipping deploy.` A red X here would train the operator to ignore failures on this workflow, which is the opposite of the goal.
3. **Snapshot step**: volume snapshot taken before deploy.
   - **Error state**: snapshot fails → workflow stops *before* deploying, job shows red, log line names the actionable next step: `Volume snapshot failed — deploy aborted, instance untouched. Check Fly volume status: flyctl volumes list -a <app>.` The operator must never wonder "did it deploy anyway without a backup?" — the log states explicitly that deploy did not proceed.
4. **Deploy step**: `flyctl deploy` against operator app/host from secrets.
   - **Error state**: deploy command fails → red job, surface the raw `flyctl` error in the log (no swallowing/wrapping it into something vaguer), plus one added line: `Snapshot <id/timestamp> is available for rollback: flyctl volumes list -a <app>.` This is the moment trust is most fragile — the operator needs to know the failure didn't corrupt state and exactly how to recover.
5. **Post-deploy verification**: `curl /health`, compare reported version to the release tag that triggered the run.
   - **Silent-failure case this closes**: deploy exits 0 but `/health` still reports the old version (slow rollout, crashed new instance serving stale cache, etc.). This is exactly the class of bug the issue itself describes ("nothing enforces the lag") — so this step must be a hard gate, not a log line. Job **fails red** if version doesn't match within a bounded retry/timeout window (mirror whatever interval #2012's client workflow already uses — do not invent a new one).
   - **Error message**: `Post-deploy health check failed: /health reports <actual> after N retries, expected <expected>. Deploy step succeeded but the running instance did not pick it up — investigate before assuming this release is live.` This distinguishes "deploy command failed" from "deploy succeeded but didn't take effect," which are different failure modes needing different operator responses.
6. **Success state**: job green, one log line confirming `<host> now serving <version>` — gives the operator a single grep-able line of positive confirmation, matching how they'd verify manually today (`curl .../health`).
7. **Notification** (per PM acceptance criteria): failure at any step (2–5) pushes a notification, not just a red GitHub Actions badge the operator has to go looking for. Reuse whatever channel the client-instance workflow already uses if it has one; if it doesn't, this is a **new requirement being introduced here**, and Eng should confirm the channel (Telegram, per the operator's existing daemon notification pattern used elsewhere in this environment, is the natural fit) — flagging this as an open question rather than assuming.

#### Discoverability
- Anyone looking at `.github/workflows/` should be able to answer "how does the operator instance get deployed?" without asking — the file's existence and name is the entire discovery mechanism. This directly closes the issue's documentation gap ("there is no doc describing how it is deployed") without requiring separate prose docs: the workflow file *is* the doc for a technical audience, provided the naming/step-naming consistency above is followed.
- One-line addition to repo's deploy documentation (wherever `deploy-client-instance.yml` is currently mentioned, if anywhere) listing the operator instance alongside sandbox/client/pages as a fourth auto-deploying target — closing the doc gap the issue explicitly calls out. If no such doc currently exists (PM/Eng to confirm), this requirement is N/A rather than silently dropped.

#### Error/empty states summary (checklist form)
- [ ] Secrets unset → skipped (neutral, not red), one-line log reason
- [ ] Snapshot failure → red, deploy aborted, actionable Fly CLI hint in log
- [ ] Deploy command failure → red, raw error surfaced, rollback/snapshot pointer in log
- [ ] Deploy succeeds but `/health` version mismatch → red (hard gate, not soft warning), distinguishes "didn't deploy" from "deployed but didn't take"
- [ ] Success → single positive confirmation line with host + version
- [ ] Any failure → push notification, not just Actions-tab visibility

#### Accessibility
N/A — no UI surface; this is a CI/CD workflow consumed via GitHub Actions' own interface and CLI/notification channels, which are out of this workflow's control.

#### Acceptance checklist (ux lens, additive to PM's)
- [ ] Workflow named `deploy-operator-instance.yml`, step names match `deploy-client-instance.yml` vocabulary
- [ ] Secret names follow existing `CLIENT_*` naming pattern with `OPERATOR_*` prefix
- [ ] Unset-secret path produces a skipped (not failed) run with an explicit log reason
- [ ] Every failure path's log line states what happened AND the concrete next action (command or check) — no bare stack traces as the only output
- [ ] Snapshot-failure and deploy-failure messages both make explicit whether the volume was touched
- [ ] `/health` version-mismatch is a hard failing gate, not a warning-only log line
- [ ] Success path logs one explicit, grep-able confirmation line
- [ ] Failure notification channel confirmed with Eng (open question, not assumed)
- [ ] Deploy documentation (if any exists) updated to list the operator instance as a fourth auto-deploying target

**Divergence check**: none — the requested flow (mirror #2012, add snapshot, add health gate) is the right shape; no upstream IA/contract problem to route around. My only addition is making the *failure* experience as first-class as the success path, since the issue's whole premise is that silent/soft failure is what created this bug in the first place.

### Engineering

**Approach**: adapt `deploy-client-instance.yml` (#2012 precedent) into a new `deploy-operator-instance.yml`. Same triggers, same job structure, new secret prefix, plus two additive steps (snapshot, hard health-gate) that should also be back-ported to the client workflow per PM/Design (item 2).

#### Files to touch

1. **`.github/workflows/deploy-operator-instance.yml`** (new)
   - Copy `.github/workflows/deploy-client-instance.yml` as the starting point; read it first to confirm actual trigger config, secret names, and step vocabulary before diverging — Design's step-naming parity requirement depends on this being exact, not assumed.
   - `on: release: types: [published]` — unchanged, mirrors existing target workflows.
   - Job-level `if:` (or first step) gates on `secrets.OPERATOR_APP_NAME != '' && secrets.OPERATOR_HOST != ''` — read how the client workflow currently handles its own secrets-unset skip (if it does at all — the issue doesn't confirm the client workflow already skips cleanly, only that Eng should make the operator one skip cleanly). If the client workflow has no such guard, this step is new here and should note that explicitly in the PR description rather than presenting it as copied.
   - Steps, in order:
     1. `checkout` (if the client workflow does)
     2. `snapshot` — `flyctl volumes snapshot create <volume-id> -a ${{ secrets.OPERATOR_APP_NAME }}`, or the volumes-list-then-snapshot pattern if no fixed volume-id secret exists; capture the snapshot id/timestamp into a step output for use in the deploy-failure log line. Fail the job (stop before deploy) on non-zero exit — no `continue-on-error`.
     3. `deploy` — `flyctl deploy -a ${{ secrets.OPERATOR_APP_NAME }}` using the same `flyctl` action/version and `FLY_API_TOKEN` secret pattern the client workflow uses (confirm token secret name — likely shared, not per-target). On failure, surface raw `flyctl` output (no wrapping) plus the snapshot id from step 2's output in the job summary/log.
     4. `verify-health` — poll `https://${{ secrets.OPERATOR_HOST }}/health`, compare reported version field to `${{ github.event.release.tag_name }}` (strip leading `v` if the health payload's version field doesn't include it — check the actual `/health` response shape before assuming string equality works). Bounded retry/timeout matching the client workflow's existing interval — read its value, don't invent a new one. Hard-fail the job on mismatch after retries exhausted; this step is the effect-verification test (PM item: fixed-means-behavior-verified, not just deploy-exit-0).
     5. `notify-on-failure` — runs `if: failure()`, posts to whatever channel item 7 resolves to (Telegram, per Design's open question — confirm before implementing; if unresolved at build time, implement a placeholder step with a clear TODO and flag it as a follow-up rather than silently dropping the requirement).
   - Success path: final step logs `${{ secrets.OPERATOR_HOST }} now serving ${{ github.event.release.tag_name }}` (Design's grep-able confirmation line).

2. **`.github/workflows/deploy-client-instance.yml`** (existing — additive change per issue item 2 "worth doing for the client instance too")
   - Add the same snapshot step ahead of its existing deploy step. Scope: PM did not mark this as a strict blocking requirement for closing #2147 (the issue text frames it as "worth doing," not "required") — treat as in-scope but lower priority than the operator workflow itself; if time-boxed, land the operator workflow first and note the client-workflow backport as a fast-follow in the PR description rather than silently dropping it.

3. **Repository secrets** (GitHub repo settings, not a file in the PR) — `OPERATOR_APP_NAME`, `OPERATOR_HOST` need to exist for the workflow to do anything on this repo. Note in the PR description that these must be set by the operator post-merge; the workflow's skip-path is what keeps forks/unset-repos green in the meantime. Do not attempt to set actual secret values from the PR.

4. **Deploy documentation** — search the repo for wherever `deploy-client-instance.yml` / sandbox / pages-site deploys are already documented (README, `docs/`, or CONTRIBUTING). If found, add the operator instance as a fourth auto-deploying target (Design's discoverability requirement). If no such doc exists anywhere, state that explicitly in the PR description as "N/A — no existing deploy doc found" rather than skipping silently or inventing a new doc file out of scope.

#### Data/contract changes

None. No schema, entity, API, or tool-definition surface is touched — pure `.github/workflows` + secrets + (maybe) doc prose. Confirms PM's interface-surface-check finding; no migration or contract versioning needed.

#### Layering / sequencing

- No dependency between the two workflow files — `deploy-operator-instance.yml` is fully self-contained and can merge independently of the client-workflow snapshot backport.
- Snapshot step must precede deploy step within each workflow (ordering constraint Design already specified — deploy must not proceed if snapshot fails).
- Health-check step must run after deploy completes, not in parallel — it is validating the deploy's effect.

#### Build-step checklist

- [ ] Read `deploy-client-instance.yml` in full — confirm actual secret names, `flyctl` action/version, `FLY_API_TOKEN` source, retry/timeout interval used in its own health check (if any), and whether it already has a secrets-unset skip guard.
- [ ] Check the operator instance's actual `/health` response shape (`curl https://<operator-host>/health` or read the handler source) to confirm the version field name and format before writing the comparison logic.
- [ ] Confirm the operator app's Fly volume id/name (`flyctl volumes list -a <operator-app>`) for the snapshot command.
- [ ] Resolve the notification-channel open question (Telegram vs. existing client-workflow channel) with the operator before finalizing step 5 — if genuinely blocking, raise as a checkpoint rather than guessing.
- [ ] Write `deploy-operator-instance.yml` per the steps above.
- [ ] Add snapshot step to `deploy-client-instance.yml`.
- [ ] Update deploy documentation if a suitable location exists; otherwise note N/A in the PR description.
- [ ] Self-review diff with the harness `code-review` built-in before opening the PR.
- [ ] Effect test (per impl sign-off checklist): since this is a CI/CD workflow, "test" means a dry-run validation — trigger a manual `workflow_dispatch` test run (add a temporary `workflow_dispatch:` trigger if useful for validation, or validate against a test release) confirming the skip-path fires cleanly with secrets unset, and document in the PR that a real end-to-end proof happens on the next actual release publish per the issue's own Verification section. Do not claim "fixed" without this evidence.
- [ ] Confirm no generated/derived files (test catalog, OpenAPI types, capability manifest) are implicated — this change touches only workflow YAML and docs, so PM's parity item is N/A; state that in the PR body rather than running the regen scripts needlessly.
- [ ] `gh pr create` referencing `closes #2147`, title matching the issue title, PR body documenting: which items from PM's acceptance checklist are met, the notification-channel decision made, and explicit confirmation of the doc-gap outcome (updated vs. N/A).

### QA / Test Plan

**Surface under test**: two GitHub Actions workflows (`deploy-operator-instance.yml` new, `deploy-client-instance.yml` snapshot backport) and their agent/operator-observable effect — the `/health` endpoint's reported version after a `release: published` event. No MCP tool, schema, or store/retrieve recipe is touched, so this is validated as a CI/CD behavioral contract, not an `agentic_eval` fixture — noted explicitly per the "no functional surface" decision tree, since this surface type isn't one `agentic_eval`/`eval-harness` can drive (no conversational turn to replay, no entity to assert on). The deliverable below is the closest reproducible equivalent: a scripted, CI-runnable contract test plus a definition-of-done checklist, not prose alone.

### Reproducible verification artifact

Since Tier-1 `agentic_eval` fixtures assert against conversational/entity effects and this change has neither, the eval-equivalent is a **`workflow_dispatch` dry-run harness** committed alongside the workflow, so the QA evidence is re-runnable on demand rather than a one-time manual check:

- Add a temporary (or permanent, if Eng agrees) `workflow_dispatch:` trigger to `deploy-operator-instance.yml` with an optional `test_mode: skip_secrets` input that lets CI simulate the unset-secrets path without touching real infrastructure.
- A small script (`.github/workflows/scripts/verify-health-version.sh` or inline step) that takes `(host, expected_version, max_retries, interval)` and returns non-zero on mismatch — this is the unit under test for the health-gate logic, and can be exercised standalone against a mock `/health` responder (e.g. `python3 -m http.server` serving a fixed JSON payload) so the version-comparison logic (including the `v`-prefix strip) is verified without a real Fly deploy.
- This script + its standalone test is what makes "effect verified" auditable in CI rather than asserted in a PR description.

### Regression coverage

No prior bug is being fixed in the traditional sense (no existing test to un-break), but the issue itself documents a **regression class already observed**: operator instance served `0.21.4` while release was `0.21.5` (deploy exit-0 semantics, or rather "no deploy at all," masking the lag). The regression test for *that* failure mode is the health-gate step itself — assert it fails loud when version doesn't match, which is the exact scenario that went undetected historically.

- [ ] **Regression case**: simulate deploy where `/health` reports the *previous* release tag after the deploy step reports success — health-gate step must fail red, not pass. This is the direct reproduction of the bug described in the issue ("client instance serves 0.21.5, operator serves 0.21.4," silent).
- [ ] **Regression case — version format drift**: confirm the comparison handles both `v0.21.5` and `0.21.5` forms (leading-`v` strip) — Eng's build checklist flags this as unconfirmed; if the actual `/health` payload format doesn't match what's assumed, the health-gate silently mis-compares and reintroduces the exact bug this issue closes. Do not ship the comparison logic without having read the real response shape.

### Edge cases for new branches (per Design's flow, steps 2–7)

- [ ] **Secrets fully unset** (fork / secrets never configured): job exits skipped (not failed, not errored). Assert on the job conclusion value itself (`skipped`), not just log text — a red or green false-positive here defeats Design's "don't train the operator to ignore red" requirement.
- [ ] **Secrets partially set** (e.g. `OPERATOR_APP_NAME` set, `OPERATOR_HOST` unset, or vice versa): must also skip cleanly, not proceed with an empty string into `flyctl`/`curl` and fail confusingly downstream. This combination isn't explicitly named upstream — worth calling out because "both or neither" is an assumption, not a stated contract.
- [ ] **Snapshot step fails** (non-zero exit from `flyctl volumes snapshot create` or the list-then-snapshot fallback): deploy step must NOT run — assert the deploy step shows `skipped`/not-invoked in the job graph, not merely that the job overall is red (a red job where deploy still ran anyway is a materially different and worse outcome than the one specified).
- [ ] **Snapshot succeeds but returns no usable id/timestamp** (e.g. malformed `flyctl` output, empty stdout): the deploy-failure log path depends on this value being available (Eng's "surface snapshot id from step 2's output"); assert a defined fallback string (e.g. `"snapshot id unavailable — check flyctl volumes list"`) rather than an empty/`undefined` interpolation in the failure message.
- [ ] **Deploy step fails** (`flyctl deploy` non-zero): assert raw `flyctl` stderr/stdout appears unmodified in the log (Eng's "no wrapping" requirement — easy to accidentally swallow via `continue-on-error` or output redirection), and the snapshot-id rollback line is present.
- [ ] **Deploy succeeds, health check matches on first poll**: fast success path — assert the single grep-able confirmation line format exactly (`<host> now serving <version>`), since Design specifies this as the operator's primary manual-verification substitute.
- [ ] **Deploy succeeds, health check matches only after N-1 retries** (slow rollout): assert the job still passes — this is the "eventually consistent" case the retry/timeout window exists for; a plan that only tests immediate-match or permanent-mismatch would miss an off-by-one in the retry loop.
- [ ] **Deploy succeeds, health check never matches within the retry/timeout window**: hard fail, assert the specific error message format naming actual vs. expected version and retry count (Eng's specified message) — not a generic timeout error, since the operator needs to distinguish this from a transport/network failure.
- [ ] **`/health` endpoint itself unreachable during verification** (network error, 5xx, connection refused — distinct from "reachable but wrong version"): confirm this is treated as a retry-then-fail case with its own clear message, not conflated with the version-mismatch message — Eng's plan doesn't currently distinguish "can't reach health endpoint" from "reached it, wrong version," and an operator debugging a DNS/networking issue needs a different first action than one debugging a stale deploy.
- [ ] **Failure notification fires exactly once per failed run**, on every failure branch (secrets-skip excluded — that's not a failure), not duplicated per retry iteration inside the health-check loop.
- [ ] **`deploy-client-instance.yml` snapshot backport**: assert its existing deploy behavior (including its own health-check semantics, if any) is unchanged for the success path — this is an additive step ahead of existing logic, and the regression risk is breaking a currently-working workflow while fixing an unrelated one. Minimal: re-run/diff the client workflow's existing test coverage (if any exists) before and after the backport; if none exists, note that as a pre-existing gap, not one this change introduces.

### Contract check (secrets → behavior)

Not a network-facing API contract, but a config contract worth asserting explicitly since it's the skip/no-skip branch point:

- [ ] Contract: `OPERATOR_APP_NAME` and `OPERATOR_HOST` both present and non-empty ⇒ workflow attempts deploy. Either absent or empty ⇒ workflow skips with the specified log line. No third state (e.g. present-but-whitespace) is silently mishandled — add an explicit case for whitespace-only secret values if trivial, otherwise note as an accepted gap.

- [ ] `workflow_dispatch` dry-run (or equivalent scripted harness) exists and is runnable on demand, not just described in a PR body — this is the reproducible artifact standing in for an `agentic_eval` fixture given the CI-only surface.
- [ ] Health-version-comparison logic has a standalone test independent of a real Fly deploy (mock `/health` responder), covering: exact match, `v`-prefix mismatch, stale version, unreachable endpoint.
- [ ] All edge cases above are exercised at least once (via dry-run harness, mock responder, or manual trigger) with the observed job/step conclusion recorded — pass/fail per case, not just "looks right."
- [ ] Regression case (stale version reported after deploy "succeeds") explicitly reproduced and confirmed caught by the health gate — this is the direct proof the originating bug class is closed.
- [ ] Real end-to-end proof deferred to the next actual `release: published` per the issue's own Verification section — this plan does NOT claim that proof exists yet; PR description must say so plainly rather than implying full production verification pre-merge.
- [ ] `deploy-client-instance.yml`'s pre-existing behavior confirmed unchanged post-backport (regression guard on the *other* file this PR touches).
- [ ] Notification-channel decision (Telegram vs. existing) is resolved and testable — at minimum, assert the notify step is invoked (`if: failure()`) in a simulated failure run, even if the actual message delivery isn't asserted end-to-end in CI.
- [ ] No unrelated `eval:tier1` fixtures touched or force-updated by this change (expected — this PR should show zero diffs under `tests/fixtures/agentic_eval/`).

**Gate disposition**: `qa` gate held pending until the `workflow_dispatch` dry-run harness and the standalone health-comparison test are committed in the PR and demonstrated green (or their absence explicitly justified as infeasible, with reasoning, not silently skipped). A PR containing only the workflow YAML with no scripted verification does not meet this bar — file as `[BLOCKING] test-coverage: no reproducible verification artifact for the health-gate contract` at review time if submitted that way.

markmhendrickson:
> **Draft — do not merge before #2226.** See _Sequencing_ below.

Closes #2147.

## The gap

Every other deploy target updates itself when a release publishes; the operator's own instance did not. All 14 of its Fly releases were manual.

On 2026-08-27 it was found serving **0.21.4** against a current **0.22.1** — four releases and ~18 days behind. The release it was missing closed an authentication bypass whose advisory was already public, so the instance holding the real graph was simultaneously the least-updated target and the highest-value one. It has since been deployed by hand; this PR stops it recurring.

## What's here

**`fly.markmhendrickson.toml`** — a per-instance config on the `fly.sandbox.toml` precedent, carrying the two values the shared `fly.toml` gets wrong for this app:

- `primary_region = 'ams'`. The volume is in `ams`; `fly.toml` defaults to `lhr`, which fails at machine creation with `needs an unattached volume named 'data' in region 'lhr'`. (`--regions ams` does *not* fix this — it filters which existing machines to deploy to. Only `primary_region` overrides it.)
- `[[vm]]` 2 CPU / 4096MB. The shared `1gb`/1cpu is correct for the sandbox and the client instance, but far too small for a ~20GB graph. A deploy **reapplies** the config file's sizing, so deploying this app from the shared `fly.toml` silently reset the machine to 1GB — observed 2026-08-04 and again 2026-08-27.

Everything not overridden is inherited from `fly.toml` on purpose — `release_command`, `[[restart]] policy = 'always'`, and the always-on settings each fix a specific production failure and shouldn't fork per instance.

**`.github/workflows/deploy-operator-instance.yml`** — modelled on `deploy-client-instance.yml`, with two additions:

1. **A pre-deploy volume snapshot**, requested in #2147. Non-fatal by design: refusing to ship a security fix because a snapshot API call failed is the wrong trade, so it warns loudly and continues.
2. **A fifth verification check — an authenticated read.** This is the one worth reviewing closely. On 2026-08-27 all four existing checks **passed** while the instance could not serve a single entity query: `/health` returned 200 in ~1ms while `GET /entities?limit=1` timed out completely. `/health` doesn't touch the database, so it cannot detect a wedged one. Check (d) now also asserts VM sizing survived the deploy, which is what proves the per-instance config actually took effect.

## Verification done

The `flyctl` JSON shapes this relies on were checked against the live app rather than assumed:

- `volumes list --json` exposes `.name` / `.id` (`vol_4ojlng96ng9kql2r`, name `data`, region `ams`).
- `machine list --json` exposes `config.guest.memory_mb` and `services[].autostop` — the latter is a **boolean**, not the `'off'` string the inherited comparison suggests. The condition handles both; unit-tested across `false` / `'off'` / `'stop'` / `true` / undersized.

Both files parse (YAML + TOML). The workflow has not been executed — it can only run on a real release.

## Sequencing — why this is a draft

#2226: the release pipeline writes `status=published` **before** deploys run, so deploy failures are invisible — `deploy-pages-site.yml` has failed all six release-triggered runs without anyone noticing.

Merging this on top of that machinery would replace a known gap with a hidden one: instead of an instance that visibly lags, we'd get a workflow that can fail silently. The Telegram notify-on-failure step is a partial mitigation, not a fix. Recommend landing #2226 first.

## Operator action required before this can work

Two repository secrets:

- `OPERATOR_INSTANCE_HOST` — hostname to verify, no scheme. The job skips cleanly when unset, so forks are unaffected.
- `OPERATOR_INSTANCE_TOKEN` — bearer token used *only* for the read-path check. Optional; without it check (e) is skipped with a warning rather than failing.

`FLY_API_TOKEN` already exists and is the right scope — this app is in the `neotoma` org, not a separate client org.

### DIFF
--- a/.github/workflows/deploy-operator-instance.yml
+++ b/.github/workflows/deploy-operator-instance.yml
@@ -1,0 +1,219 @@
+name: Deploy operator instance
+
+# Deploys the OPERATOR's own Neotoma instance whenever a GitHub Release is
+# published, so it stops silently lagging behind every release.
+#
+# WHY THIS EXISTS: every other deploy target updated itself on release while
+# this one did not (issue #2147). All 14 of its Fly releases were manual, from
+# the operator's laptop. On 2026-08-27 it was found serving 0.21.4 against a
+# current 0.22.1 — four releases and ~18 days behind, and the version it was
+# missing closed an authentication bypass whose advisory was already public.
+# It is the instance holding the real graph, so it was simultaneously the
+# least-updated target and the highest-value one.
+#
+# Per-instance app name, region and VM sizing live in fly.markmhendrickson.toml
+# (the fly.sandbox.toml precedent). Do NOT deploy this app from the shared
+# fly.toml: its 'lhr' default fails against this instance's ams volume, and its
+# 1gb VM sizing wedges the database.
+#
+# Required repository secrets:
+#   FLY_API_TOKEN            Fly deploy token for the `neotoma` org (same org as
+#                            the sandbox — this app is not a client instance).
+#   OPERATOR_INSTANCE_HOST   hostname to verify (no scheme).
+#   OPERATOR_INSTANCE_TOKEN  bearer token used ONLY for the read-path check
+#                            below. Optional: without it that check is skipped
+#                            with a warning rather than failing the deploy.
+#
+# The job SKIPS cleanly (does not fail) when OPERATOR_INSTANCE_HOST is unset, so
+# forks and contributors without the secret are unaffected.
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Git ref to deploy (defaults to the workflow ref)"
+        required: false
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: deploy-operator-instance
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    env:
+      FLY_APP: neotoma-markmhendrickson
+    steps:
+      - name: Check instance is configured
+        id: gate
+        env:
+          HOST: ${{ secrets.OPERATOR_INSTANCE_HOST }}
+        run: |
+          if [ -z "$HOST" ]; then
+            echo "OPERATOR_INSTANCE_HOST not set — no operator instance configured; skipping."
+            echo "configured=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "configured=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Checkout
+        if: steps.gate.outputs.configured == 'true'
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref || github.sha }}
+
+      - name: Setup flyctl
+        if: steps.gate.outputs.configured == 'true'
+        uses: superfly/flyctl-actions/setup-flyctl@master
+
+      # Snapshot the volume BEFORE deploying (#2147 requested this explicitly).
+      # This instance has a persistent 20GB volume holding the operator's real
+      # graph, and neither of the other deploy workflows takes a backup — a
+      # manual deploy did not take one either, so this is a net improvement
+      # rather than a regression in safety.
+      #
+      # Fly takes scheduled snapshots too; this adds one pinned to the moment
+      # before a release lands, which is the point a bad migration would need
+      # rolling back to. Non-fatal on purpose: refusing to ship a security fix
+      # because a snapshot API call failed is the wrong trade. It logs loudly.
+      - name: Snapshot volume before deploy
+        if: steps.gate.outputs.configured == 'true'
+        continue-on-error: true
+        env:
+          FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
+        run: |
+          set -uo pipefail
+          VOL=$(flyctl volumes list --app "$FLY_APP" --json \
+            | node -e 'const v=JSON.parse(require("fs").readFileSync(0,"utf8"));const d=v.find(x=>x.name==="data");process.stdout.write(d?d.id:"")')
+          if [ -z "$VOL" ]; then
+            echo "::warning::No volume named 'data' found on $FLY_APP — skipping snapshot."
+            exit 0
+          fi
+          echo "Snapshotting volume $VOL"
+          if flyctl volumes snapshots create "$VOL" --app "$FLY_APP"; then
+            echo "Snapshot created for $VOL"
+          else
+            echo "::warning::Snapshot failed for $VOL — continuing with deploy."
+          fi
+
+      # -c fly.markmhendrickson.toml carries app name, primary_region = 'ams'
+      # and the 2 CPU / 4096MB sizing. VITE_NEOTOMA_SANDBOX_UI must stay empty
+      # (no "Public sandbox" banner) and VITE_PUBLIC_BASE_PATH must be "/" (a
+      # "/inspector/" build renders a blank page while /health still returns ok).
+      - name: Deploy to Fly (operator instance)
+        if: steps.gate.outputs.configured == 'true'
+        env:
+          FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
+        run: |
+          flyctl deploy -c fly.markmhendrickson.toml --remote-only \
+            --build-arg VITE_NEOTOMA_SANDBOX_UI="" \
+            --build-arg VITE_PUBLIC_BASE_PATH="/" \
+            --build-arg NEOTOMA_GIT_SHA="${{ github.sha }}"
+
+      # FIVE post-deploy checks. The first four mirror deploy-client-instance.yml
+      # ("/health alone is NOT sufficient"). The fifth is new, and 2026-08-27 is
+      # why: all four of the others PASSED while the instance could not serve a
+      # single entity query. /health returned 200 in ~1ms with a wedged DB.
+      - name: Verify deploy
+        if: steps.gate.outputs.configured == 'true'
+        env:
+          EXPECTED_SHA: ${{ github.sha }}
+          HOST: ${{ secrets.OPERATOR_INSTANCE_HOST }}
+          READ_TOKEN: ${{ secrets.OPERATOR_INSTANCE_TOKEN }}
+          FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
+        run: |
+          set -euo pipefail
+          export EXPECTED_VERSION="$(node -p "require('./package.json').version")"
+
+          # (a) /health responds ok.
+          curl -fsS --retry 5 --retry-delay 5 --retry-all-errors \
+            "https://$HOST/health" -o /tmp/health.json
+          node -e '
+            const j = require("/tmp/health.json");
+            if (!j.ok) { console.error("FAIL: /health not ok"); process.exit(1); }
+            console.log("OK /health:", JSON.stringify(j));
+          '
+
+          # (b) deployed git_sha matches the released commit (drift detection).
+          curl -fsS --retry 5 --retry-delay 5 --retry-all-errors \
+            -H "Accept: application/json" "https://$HOST/" -o /tmp/root.json
+          node -e '
+            const j = require("/tmp/root.json");
+            const expSha = process.env.EXPECTED_SHA;
+            const expVer = process.env.EXPECTED_VERSION;
+            const fail = (m) => { console.error("FAIL: " + m); process.exit(1); };
+            if (j.version !== expVer) fail(`version ${j.version} != ${expVer}`);
+            if (/^[0-9a-f]{40}$/.test(j.git_sha || "")) {
+              if (j.git_sha !== expSha) fail(`git_sha ${j.git_sha} != ${expSha}`);
+              console.log("OK git_sha:", j.git_sha);
+            } else {
+              fail(`git_sha ${j.git_sha} is not a commit — build arg not applied`);
+            }
+          '
+
+          # (c) the landing page actually RENDERS (non-trivial byte count).
+          BYTES=$(curl -fsS --retry 3 --retry-delay 5 "https://$HOST/" | wc -c | tr -d ' ')
+          echo "landing page bytes: $BYTES"
+          if [ "$BYTES" -lt 1000 ]; then
+            echo "FAIL: landing page is $BYTES bytes — likely a blank-page build"
+            exit 1
+          fi
+
+          # (d) machine is always-on AND kept its sizing. The sizing half is
+          #     specific to this instance: a deploy reapplies the config file's
+          #     [[vm]], so this asserts fly.markmhendrickson.toml actually took
+          #     effect rather than the 1gb shared default.
+          flyctl machine list --app "$FLY_APP" --json > /tmp/machines.json
+          node -e '
+            const ms = require("/tmp/machines.json");
+            const fail = (m) => { console.error("FAIL: " + m); process.exit(1); };
+            const stopping = ms.filter(m => m?.config?.auto_destroy || m?.config?.services?.some(s => s.autostop && s.autostop !== "off"));
+            if (stopping.length) fail("machine(s) not always-on: " + stopping.map(m => m.id).join(","));
+            const small = ms.filter(m => (m?.config?.guest?.memory_mb ?? 0) < 4096);
+            if (small.length) {
+              fail("machine(s) under-sized after deploy: " +
+                small.map(m => `${m.id}=${m.config.guest.memory_mb}MB`).join(",") +
+                " — expected >=4096MB. The deploy reset VM sizing; fly.markmhendrickson.toml did not take effect.");
+            }
+            console.log("OK always-on + sizing:", ms.map(m => `${m.id}:${m.config.guest.memory_mb}MB`).join(","));
+          '
+
+          # (e) THE READ PATH ACTUALLY WORKS. Checks (a)-(d) all passed on
+          #     2026-08-27 while every entity query hung — /health does not
+          #     touch the database, so it cannot detect a wedged one. Generous
+          #     timeout: a cold read against a large graph legitimately takes
+          #     ~15s right after a restart.
+          if [ -z "${READ_TOKEN:-}" ]; then
+            echo "::warning::OPERATOR_INSTANCE_TOKEN not set — skipping the read-path check. /health passing does NOT prove the database is serving."
+          else
+            echo "Checking the read path (authenticated entity query)…"
+            CODE=$(curl -s -o /tmp/read.json -w '%{http_code}' --max-time 90 \
+              -H "Authorization: Bearer $READ_TOKEN" \
+              "https://$HOST/entities?limit=1" || echo "000")
+            if [ "$CODE" != "200" ]; then
+              echo "FAIL: authenticated read returned HTTP $CODE (000 = timeout/no response)."
+              echo "      /health can return 200 while the database is wedged — this is that case."
+              exit 1
+            fi
+            echo "OK read path: HTTP 200"
+          fi
+
+      - name: Notify on failure
+        if: failure() && steps.gate.outputs.configured == 'true'
+        env:
+          TELEGRAM_BOT_TOKEN: ${{ secrets.TELEGRAM_BOT_TOKEN }}
+          TELEGRAM_CHAT_ID: ${{ secrets.TELEGRAM_CHAT_ID }}
+        run: |
+          if [ -n "${TELEGRAM_BOT_TOKEN:-}" ] && [ -n "${TELEGRAM_CHAT_ID:-}" ]; then
+            curl -fsS -X POST \
+              "https://api.telegram.org/bot${TELEGRAM_BOT_TOKEN}/sendMessage" \
+              -d "chat_id=${TELEGRAM_CHAT_ID}" \
+              -d "text=🔴 OPERATOR instance deploy FAILED for ${{ github.ref_name }} — it may be serving a stale or broken build, or its database may be unreachable. Run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}" \
+              >/dev/null || true
+          fi
--- a/fly.markmhendrickson.toml
+++ b/fly.markmhendrickson.toml
@@ -1,0 +1,86 @@
+# Fly config for the OPERATOR's own Neotoma instance
+# (https://neotoma.markmhendrickson.com).
+#
+# WHY THIS FILE EXISTS — it encodes two per-instance values that the shared
+# fly.toml gets wrong for this app, both of which have already caused outages
+# when left to a remembered manual step:
+#
+#   1. primary_region = 'ams'. This instance's volume lives in ams; fly.toml
+#      defaults to 'lhr'. Deploying without the right region FAILS at machine
+#      creation with "New machine in group 'app' needs an unattached volume
+#      named 'data' in region 'lhr'". (`--regions ams` does NOT fix this — it
+#      filters which existing machines to deploy to. Only primary_region, or
+#      `--primary-region`, overrides fly.toml.)
+#
+#   2. [[vm]] 2 CPU / 4096MB. fly.toml declares 1gb/1cpu, which is correct for
+#      the sandbox and the client instance but far too small here: this app
+#      serves a ~20GB graph. A deploy REAPPLIES the config file's sizing, so
+#      every deploy from the shared fly.toml silently reset this machine to
+#      1GB — observed 2026-08-04 and again 2026-08-27. On 2026-08-27 the
+#      consequence was measured: `/health` kept returning 200 in ~1ms while
+#      `GET /entities?limit=1` timed out completely. A green health check with
+#      a wedged database is the worst failure signature available, because
+#      every automated check passes.
+#
+# Following the fly.sandbox.toml precedent: per-instance app name, region and
+# sizing live with the instance, not in the shared default.
+#
+# Everything NOT overridden here is inherited from fly.toml on purpose —
+# notably [deploy].release_command, [[restart]] policy = 'always', and the
+# always-on http_service settings. Keep it that way: those were each added to
+# fix a specific production failure and should not fork per instance.
+
+app = 'neotoma-markmhendrickson'
+primary_region = 'ams'
+
+[build]
+
+# Seed the schema_registry on every deploy, before the new release takes
+# traffic (issue #1968). Mirrors fly.toml; safe to re-run.
+[deploy]
+  release_command = "node dist/seed_schemas_entry.js"
+
+[env]
+  NODE_ENV = "production"
+  NEOTOMA_ENV = "production"
+  # Must match internal_port (Fly routes to this port on the machine)
+  HTTP_PORT = "3180"
+  NEOTOMA_HTTP_PORT = "3180"
+
+[http_service]
+  internal_port = 3180
+  force_https = true
+  # Always-on. MCP-over-HTTP transport sessions live in process memory, so any
+  # machine stop/start silently invalidates every live session — see the fuller
+  # explanation in fly.toml. This matters more here than anywhere else: this is
+  # the instance every session and daemon actually writes to.
+  auto_stop_machines = 'off'
+  auto_start_machines = true
+  min_machines_running = 1
+  processes = ['app']
+
+  # Service check — governs proxy ROUTING only, it does not restart a stopped
+  # machine. Restart behavior is the [[restart]] block below.
+  [[http_service.checks]]
+    interval = '15s'
+    timeout = '10s'
+    grace_period = '30s'
+    method = 'GET'
+    path = '/health'
+
+# Restart on ANY process end, including a clean exit — Fly's default
+# `on-failure` ignores exit code 0, and this server does exit 0 periodically
+# with no error (neotoma#2094). See fly.toml for the full rationale.
+[[restart]]
+  policy = 'always'
+
+# 2 CPU / 4096MB — see note 2 in the header. Do NOT reduce this to match
+# fly.toml: the 1GB default wedges this instance's database.
+[[vm]]
+  memory = '4096mb'
+  cpu_kind = 'shared'
+  cpus = 2
+
+[[mounts]]
+  source = "data"
+  destination = "/app/data"
--- a/fly.markmhendrickson.toml
+++ b/fly.markmhendrickson.toml
@@ -12,15 +12,28 @@
 #      filters which existing machines to deploy to. Only primary_region, or
 #      `--primary-region`, overrides fly.toml.)
 #
-#   2. [[vm]] 2 CPU / 4096MB. fly.toml declares 1gb/1cpu, which is correct for
-#      the sandbox and the client instance but far too small here: this app
-#      serves a ~20GB graph. A deploy REAPPLIES the config file's sizing, so
-#      every deploy from the shared fly.toml silently reset this machine to
-#      1GB — observed 2026-08-04 and again 2026-08-27. On 2026-08-27 the
-#      consequence was measured: `/health` kept returning 200 in ~1ms while
-#      `GET /entities?limit=1` timed out completely. A green health check with
-#      a wedged database is the worst failure signature available, because
-#      every automated check passes.
+#   2. [[vm]] 2 PERFORMANCE CPUs / 4096MB. fly.toml declares 1gb/1cpu shared,
+#      which is correct for the sandbox and the client instance but wrong here
+#      on both axes. A deploy REAPPLIES the config file's [[vm]], so every
+#      deploy from the shared fly.toml silently reset this machine — observed
+#      2026-08-04 and again 2026-08-27.
+#
+#      cpu_kind = 'performance' is the load-bearing part. On 2026-08-27 this
+#      instance was effectively unusable: `/health` returned 200 while
+#      `GET /entities?limit=1` timed out at 90s. The cause was measured as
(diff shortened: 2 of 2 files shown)


neotoma-agent:
**BLOCKED**

PR #2251 closes parent issue #2147. Authoritative Neotoma `gate_status` has unsigned pre-impl gates — PR gate inheritance blocks until they clear.

### Gate status (parent issue #2147 — Neotoma authoritative)

| Gate | Owner | State | Source |
|---|---|---|---|
| pm | Pavo | `signed_off` | Neotoma `gate_status.pm` |
| ux | Accipiter | `pending` | Neotoma `gate_status.ux` |
| arch | Waxwing | `pending` | Neotoma `gate_status.arch` |
| impl | Cicada | `pending` | — |
| pr_review | Vanellus | `pending` | — |
| qa | Phoenicurus | `pending` | — |
| legal | Buteo | `not_required` | — |

**Unsigned pre-impl gates**: `ux` (Accipiter), `arch` (Waxwing).

### Resolution paths

1. **Gate owner signs off** on #2147: Accipiter sets `gate_status.ux → signed_off`; Waxwing authors the `### Architecture` section and sets `gate_status.arch → signed_off`.
2. **Operator override** (only **@markmhendrickson**): reply `/confirm-gates-clear` on #2147 — waives all unsigned pre-impl gates and re-triggers the PR pipeline.

- [x] Parent issue entity retrieved (`ent_609095b65db23a586466e49d`)
- [x] Pre-impl gate check applied (pm/ux/arch)
- [ ] Vanellus reviewer assignment — withheld until gates clear

Merge stays operator-gated regardless of this block.

📎 Neotoma: [neotoma#2147](https://neotoma.markmhendrickson.com/entities/ent_609095b65db23a586466e49d)

GATE_INHERITANCE: blocked
GATE_PENDING: ux,arch

markmhendrickson:
## Updated: `cpu_kind` is now `performance`, not `shared`

Pushed 242a8b2. This changes the load-bearing line in `fly.markmhendrickson.toml`, and the reason is worth recording because it reverses the diagnosis this PR originally carried.

The instance became effectively unusable this afternoon: `/health` returned 200 while `GET /entities?limit=1` timed out at 90s. I had attributed the earlier degradation to the 1GB memory reset. That was wrong — or at least incomplete.

**Root cause was hypervisor CPU steal at 90–93%** (`/proc/stat` field 8). The guest was getting ~5–7% of the CPU it believed it had. A load average of 4+ on 2 vCPUs was threads *runnable but not scheduled*, which explains the thing that didn't fit: load never tracked traffic. An hour with only 80 tool calls still sat at load 3.3.

Ruled out by measurement before concluding that:

| Hypothesis | Measurement |
|---|---|
| Memory | 750MB used of 3.8GB, zero swap |
| Disk | 1.6GB on a 19.6GB volume, 9% |
| WAL/checkpoint backlog | `-wal` 11.7MB, mtimes advancing — writes landing |
| Stuck writers | Exactly one node process |
| Blocked I/O | Zero threads in D state, iowait 0.0% |
| App CPU saturation | ~17% of one core |

`shared-cpu` machines are burstable: sustained demand drains the CPU balance, then the machine is throttled hard. This app holds a ~20GB graph written to continuously by a daemon swarm and every Claude Code session — not a burstable workload. Worth noting that raising vCPUs while leaving `cpu_kind = 'shared'` can drain the balance *faster*, so the earlier 2-vCPU resize may have masked and then worsened this.

Verified after switching the live machine:

```
steal   90-93%     -> 0.1-0.2%
load    4.1        -> 0.53
idle    0.1%       -> 94.4%
/health 3.8-16.7s  -> 0.22-0.31s
GET /entities?limit=1   90s timeout -> HTTP 200 (3/3)
```

**Why this mattered for the PR specifically:** the file previously specified `cpu_kind = 'shared'`, matching what the machine ran at the time. Merging that would have made the first release-triggered deploy silently revert the fix on the live machine — exactly the regression class this per-instance config exists to prevent. A deploy reapplies `[[vm]]`.

Entity reads are still slow in absolute terms (14–18s against the ~20GB graph). That's a separate query-performance question, not this outage.

Still a draft pending #2226.

neotoma-agent:
review:arch

**Scope note:** This PR (`.github/workflows/deploy-operator-instance.yml`, `fly.markmhendrickson.toml`) is CI/deploy infrastructure only. It touches no OpenAPI spec, `contract_mappings.ts`, MCP tool, CLI command, HTTP endpoint, response shape, schema field, or entity relationship. Most of my standing lens (contract-first ordering, schema declaration for new response fields, tenant isolation on entity lookups, idempotency_key on mutating ops) is simply not in scope here — there are no entity lookups or Neotoma mutations in this diff. I reviewed for the parts of my lens that do apply: pattern conformance of the new deploy-workflow artifact against established precedent, and layering.

**Pattern conformance — clean.** Both new files explicitly declare and follow existing precedent rather than inventing a parallel mechanism:
- `deploy-operator-instance.yml` mirrors `deploy-client-instance.yml`'s skeleton exactly: same gate-step-via-`GITHUB_OUTPUT` pattern, same `concurrency` group shape, same checkout/setup-flyctl/deploy/verify/notify staging, same build-arg pair (`VITE_NEOTOMA_SANDBOX_UI=""`, `VITE_PUBLIC_BASE_PATH="/"`), same "skip cleanly when secret unset" fork-safety behavior. Checks (a)-(d) are the client-instance checks verbatim; (e) is a genuinely new addition, justified in-file by a concrete 2026-08-27 incident (`/health` 200 while `/entities?limit=1` timed out) rather than spec speculation.
- `fly.markmhendrickson.toml` follows the `fly.sandbox.toml` per-instance-config precedent (declared in its own header), inheriting `release_command`, `[[restart]] policy = 'always'`, and always-on `http_service` settings from the shared `fly.toml` rather than forking them — consistent with the "Python for daemons / no SaaS notification services"-style discipline of keeping shared behavior in the shared config and only overriding what's genuinely per-instance (region, VM sizing).

No structural one-off here — this is the correct extension of an established pattern, not a new parallel one.

[NON-BLOCKING] deploy-observability: deploy outcome for this new target is not recorded in Neotoma (only visible in Actions logs + Telegram-on-failure). #2226 (open, arch-gated by me) establishes that `release_result` should carry deploy fan-out state per target so "did this release reach target X" is queryable from Neotoma rather than only from CI history. This PR predates that schema extension landing, so it isn't a regression, but once #2226 ships, `deploy-operator-instance.yml`'s outcome should be wired into the same `release_result.deploy_target` fan-out rather than staying a fifth silent-in-Neotoma workflow. Unverified/forward-looking — not something the diff itself contradicts today; flagging so it isn't lost. Track as a follow-up against #2226, not this PR.

[NON-BLOCKING] sequencing: the PR body states "draft — do not merge before #2226" and #2226 is still OPEN as of this review. That's the author's own gate, already correctly self-identified and consistent with the rationale in #2226 (silent release-fan-out failures) — I'm not asserting a new blocker, just confirming the stated dependency is real and unresolved. This should stay draft until #2226 lands or the sequencing note is explicitly revisited.

No `[BLOCKING]` findings from the arch lens. This is diff-only review (no checkout/execution) — the pattern-conformance and layering findings above are evident directly from the diff and precedent files, not runtime claims, so they don't need the reduced evidence-bar caveat that would apply to behavioral assertions.

**Verdict: SIGNED_OFF** (arch gate) — no structural or interface-consistency objection to this PR. Merge readiness is otherwise gated by the author's own stated #2226 sequencing, not by anything in my lens.

📎 Neotoma: [neotoma#2147](https://github.com/markmhendrickson/neotoma/issues/2147) · [neotoma#2226](https://github.com/markmhendrickson/neotoma/issues/2226)

markmhendrickson:
Closing as superseded. `.github/workflows/deploy-operator-instance.yml` is on `main` at 412 lines, landed by #2278 and extended by #2289; this PR proposes creating it at 219. The purpose (deploy the operator instance, closes #2147) is fully served.

I checked the one open question before closing — whether its second file, `fly.markmhendrickson.toml`, is needed. It is not: main's workflow deploys with `-c fly.operator.toml`, which exists on main and is documented there as load-bearing. Nothing references the file this PR adds.

Separately worth noting: that filename is client-identifying, and this repo is public. Per-instance deploy config belongs in a `deployment_configuration` entity rather than a committed file — which is what main's approach already does.

Branch not deleted.

Issue #2156, opened 2026-08-10
REST /relationships silently ignores the entity_id filter and returns the whole table


markmhendrickson:
## Summary

`GET /relationships?entity_id=<id>` ignores the filter and returns the entire relationship table with HTTP 200. The MCP tool `list_relationships` filters the same entity correctly, so the two surfaces disagree.

## Reproduction

```
GET /relationships?entity_id=ent_037440e177d1219fc629ed01&limit=50
→ total: 60960, and ZERO returned rows reference the requested entity
```

Same entity via MCP:

```
list_relationships(entity_id="ent_037440e177d1219fc629ed01", limit=10)
→ total: 2, both rows correctly reference the entity (direction: inbound)
```

`source_entity_id=<id>` DOES filter correctly on REST. `from_entity_id` and `entity` are also silently ignored (each returns ~60,940).

## Why this is more than cosmetic

1. **Silently wrong, not an error.** HTTP 200 with a plausible payload. A caller that does not re-check `source_entity_id`/`target_entity_id` on every row concludes the entity has unrelated edges, or none.
2. **Unusable at that size.** An unfiltered 1,000-row page takes ~76s; the full table would take ~77 minutes. Combined with (1), edge traversal returns wrong answers slowly.
3. **It defeats the documented fallback.** Where field-level attribution fails, edge traversal is the recommended second path. When that path silently returns the whole table, the fallback reads as "this entity has no useful edges" — a wrong conclusion rather than a missing one. This is the mechanical reason by-person corpus audits fail (cf. #2148).

## Suggested fix

Apply `entity_id` as a bidirectional filter on REST, matching `list_relationships`. If a parameter is unsupported, reject with 400 rather than ignoring it — silently dropping an unrecognized filter is the core hazard.

## Environment

Neotoma v0.21.4, hosted instance. Observed 2026-08-10.

## Swarm specification

_This section is maintained by the Ateles swarm. Each lens agent owns exactly one subsection below; the human-written description above these markers is never modified._

### Product / Scope (PM)

**Problem**
REST `GET /relationships?entity_id=<id>` ignores the filter and returns the full relationship table with HTTP 200. MCP `list_relationships` on the same id returns that entity's real edges. Callers that treat REST as the documented edge-traversal fallback get a silently wrong answer — the mechanical failure behind #2148 by-person audits.

**In scope**
[...]

Issue #369, opened 2026-05-21
bug: OpenAPI spec undeclares /list_relationships request/response fields and naming inconsistency with /retrieve_graph_neighborhood


castor-agent:
## Summary

Two contract surface violations introduced in v0.13.0:

**1. Undeclared fields on `/list_relationships`**

`openapi.yaml` still declares this endpoint's requestBody as `type: object, additionalProperties: true` with no documented fields. The handler accepts five request fields (`source_entity_id`, `target_entity_id`, `user_id`, `limit`, `offset`) and returns three pagination envelope fields (`total`, `limit`, `offset`) — none declared in the spec. Per `docs/architecture/openapi_contract_flow.md`, spec changes must precede handler changes and response fields must be declared. Clients have no contractual basis for the pagination shape.

**2. Naming inconsistency between endpoints**

- `/list_relationships` returns `total`
- `/retrieve_graph_neighborhood` returns `total_count`

The convention established elsewhere in the API is `total_count`. One of these must be renamed to match.

## Remediation required

- Update `openapi.yaml` to declare all five request fields and the three response envelope fields for `/list_relationships` with `additionalProperties: false`
- Reconcile the `total` vs `total_count` naming — standardize on `total_count` and update `/list_relationships` response (treating the current `total` field name as uncontracted since it was never declared)
- Run `npm run openapi:generate` and commit the generated types
- Run `npm run openapi:bc-diff` and document any breaking changes in the in-progress release supplement

Issue #277, opened 2026-05-19
delete_relationship requires caller to know relationship_type with no discovery path


castor-agent:
## Summary

`delete_relationship` requires `relationship_type` as a mandatory parameter, but there is no query that lets a caller discover the relationship type(s) between two specific entities without pulling the full neighborhood of one of them. `list_relationships` has an empty parameter schema (appears to be a stub) and is not usable. This creates a gap: callers who need to delete a relationship but don't already know its type have no reliable discovery path short of a full neighborhood pull.

## Reproduction

1. Attempt to use `list_relationships` to find relationships between entity A and entity B. Observe that the tool schema has no documented parameters (`parameters: {}`), making it non-functional.
2. To discover the relationship type, call `retrieve_graph_neighborhood` on one entity — but this fails for well-connected nodes due to response truncation (#275).
3. If the guessed `relationship_type` is wrong, `delete_relationship` errors with no indication of what the actual type is.

## Impact

- Deleting erroneous edges requires knowing the type in advance or doing a full neighborhood pull
- `list_relationships` is currently unusable (empty schema, no documented parameters)
- If type is guessed incorrectly, the error message does not suggest the correct type

## Expected behavior

One of:
1. `list_relationships` should accept `source_entity_id` and/or `target_entity_id` filter parameters and return all relationships between two entities including their types
2. Or `delete_relationship` should accept an optional `relationship_type: "*"` wildcard to delete all relationship types between two entities
3. Or the error response for a wrong relationship_type should include a `hint` listing the actual relationship types that exist between the two entities

## Suggested fix

Fix `list_relationships` to accept `source_entity_id`, `target_entity_id`, and `relationship_type` filter parameters. This would also subsume the need for full neighborhood pulls in the common case of "what edges connect these two nodes?"

PR #2274, opened 2026-09-01
fix(mcp): honor source_entity_id/target_entity_id in list_relationships

### LINKED ISSUE #1973: list_relationships returns 0 when filtered by source_entity_id (incl. source+target pair and source+type), but entity_id+direction finds the same live edges
## Summary

`list_relationships` returns **zero results** for relationships that demonstrably exist and are live, when filtered by `source_entity_id` (with or without `target_entity_id` / `relationship_type`). The same relationships are returned correctly when the identical entity is queried via the legacy `entity_id` + `direction` filter. So the newer, documented filter parameters miss live edges that the legacy path finds.

This matters because `source_entity_id` + `target_entity_id` is the **documented way to discover an edge's `relationship_type` before calling `delete_relationship`**, and `source_entity_id` + `relationship_type` is the natural query for "all edges of type T out of node N" (e.g. traversing a person's `knows` / `works_at` edges). Both silently return empty.

## Reproduction

Given a live relationship `related_to` from source `S` to target `T` (created via `create_relationship`, confirmed `is_live: 1`):

1. `list_relationships(source_entity_id: S, relationship_type: "related_to", direction: "outgoing")` → `{"relationships":[],"total":0}`  ❌ (should return the edge)
2. `list_relationships(source_entity_id: S, target_entity_id: T)` → `{"relationships":[],"total":0}`  ❌ (should return the edge; this is the documented pre-delete type-discovery call)
3. `list_relationships(entity_id: S, direction: "outgoing", limit: 200)` → returns the edge (and all others), each `is_live: 1` with correct snapshot/metadata  ✅

Observed with 50 live `related_to` edges sharing one source `S`: query (3) returns `total: 50`; queries (1) and (2) return `total: 0`.

## Impact

- **Pre-delete type discovery is broken.** The tool's own description says: "To discover the type before delete_relationship, pass both source_entity_id and target_entity_id." That call returns nothing for a live edge, so a caller following the documented flow concludes the edge doesn't exist.
- **Typed out-edge traversal is broken.** "All `knows`/`works_at` out of node N" (`source_entity_id` + `relationship_type`) returns empty, forcing callers back to the un-typed `entity_id` path plus manual client-side filtering.

## Expected

`source_entity_id` (alone, with `target_entity_id`, and with `relationship_type`) must match the same live relationships that `entity_id` + `direction: outgoing` returns. The `source_entity_id`/`target_entity_id` filters and the `entity_id`+`direction` filter should be consistent views over the same relationship rows.

## Likely cause (hypothesis)

The `source_entity_id`/`target_entity_id`/`relationship_type` code path and the legacy `entity_id`/`direction` path appear to resolve against different sources of truth (e.g. one queries the live snapshot table, the other a different index/view), or the newer filter applies an extra predicate (a join, a scope, or a soft-delete/`is_live` condition) that excludes live rows. The `entity_id`+`direction` path is correct; the `source_entity_id` path is the regression.

Reported against a Fly-hosted stock Neotoma instance via the MCP.

## Swarm specification

### Product / Scope (PM)

You've hit your session limit · resets 8:10pm (Europe/Madrid)

### Engineering

You've hit your session limit · resets 8:10pm (Europe/Madrid)

### QA / Test Plan

You've hit your session limit · resets 8:10pm (Europe/Madrid)

### Security / Arch

You've hit your session limit · resets 8:10pm (Europe/Madrid)

### LINKED ISSUE #2205: list_relationships returns empty for existing live relationships (source/target filters)
## Repro (observed 2026-08-24, prod, MCP)

1. `create_relationship(PART_OF, source=<task>, target=<plan>)` → success payload with `relationship_key`, `is_live: 1`.
2. `list_relationships(source_entity_id=<task>)` → `{"relationships":[],"total":0}`.
3. `list_relationships(target_entity_id=<plan>, relationship_type="PART_OF")` → `{"relationships":[],"total":0}` — even though multiple live PART_OF edges onto that plan exist and a subsequent idempotent `create_relationship` for the same key returns `observation_count: 2` (proving the edge existed).

## Impact

Relationship existence cannot be verified through the read path, so callers double-write (harmless but wasteful thanks to idempotency) or, worse, wrongly conclude links are missing. This undermines the verify-writes-land discipline: the write path and read path disagree.

## Expected

`list_relationships` filtered by `source_entity_id` or `target_entity_id` returns live relationship rows consistent with what `create_relationship` reports.

## Swarm specification

### Product / Scope (PM)

**Problem:** After a successful `create_relationship`, MCP `list_relationships` filtered by `source_entity_id` or `target_entity_id` returns empty results even though live edges exist — breaking write verification for agents.

**In scope**
- Fix MCP `list_relationships` so `source_entity_id`, `target_entity_id`, and combined source+target filters return live edges consistent with `create_relationship` (default `include_deleted=false`, `is_live=1` only).
- Preserve existing HTTP `POST /list_relationships` behavior; align MCP handler query semantics with it (pagination, `total`/`total_count`, `relationship_type` filter, soft-delete exclusion).
- Integration tests that reproduce the reported repro sequence end-to-end.
- Cross-surface parity tests: **MCP tool** and **HTTP POST `/list_relationships`**, each using that surface's natural call shape.

**Out of scope**
- New relationship types, schema fields, or OpenAPI/MCP contract changes.
- `GET /relationships` (CLI list path) — separate surface; only in scope if arch finds it shares this defect during implementation.
- Pagination redesign, performance optimization beyond fixing the filter bug.
- Data backfill/migration (edges already exist; read path is wrong).

**Acceptance criteria**
- [ ] **Effect — source filter:** Given entities A (task) and B (plan), after `create_relationship(PART_OF, source=A, target=B)` succeeds with `is_live: 1`, `list_relationships(source_entity_id=A)` returns a row whose `relationship_key`, `source_entity_id`, `target_entity_id`, and `relationship_type` match the created edge; `total` ≥ 1.
- [ ] **Effect — target + type filter:** With multiple live `PART_OF` edges onto plan B, `list_relationships(target_entity_id=B, relationship_type=PART_OF)` returns all matching live edges including the one from A; `total` matches the live count for that filter.
- [ ] **Effect — idempotent verify:** After the above create, a second idempotent `create_relationship` for the same triple increments `observation_count` without changing the relationship key; the subsequent `list_relationships` call still returns the edge (no false-negative empty list).
- [ ] **Cross-surface parity (MCP):** An integration test drives the MCP `list_relationships` tool with `source_entity_id` / `target_entity_id` filters and asserts the observable rows above.
- [ ] **Cross-surface parity (HTTP):** An integration test drives `POST /list_relationships` with the same filters and asserts equivalent live-edge results (same keys and endpoints).
- [ ] **Soft-delete unchanged:** With `include_deleted=false` (default), soft-deleted edges remain excluded; no regression to #1570 liveness behavior.
- [ ] **No contract drift:** OpenAPI and MCP tool schemas unchanged; fix is implementation parity only.

**Priority & sequencing**
- **P0 bug** — blocks relationship verification for all MCP agents (Ateles swarm, hooks, verify-writes-land discipline).
- Unblocks reliable PART_OF / REFERS_TO linking workflows and downstream gates (arch → impl → QA).
- Should ship before any work that depends on post-write relationship reads.

**Open questions**
- None blocking — scope and expected root cause (MCP handler ignores source/target filters) are clear from repro + code inspection.

### Engineering

**Root cause (from code inspection):** the MCP `list_relationships` handler builds its query differently from the HTTP `POST /list_relationships` path — it either fails to bind `source_entity_id`/`target_entity_id` into the WHERE clause, or filters on a column/join that excludes rows the write path considers live (likely comparing against a denormalized/stale column instead of the same `is_live=1` liveness check `create_relationship` reports through). The HTTP path is the known-good reference implementation.

**Files/modules to touch**
- `src/mcp/tools/list_relationships.ts` (or equivalent MCP tool handler for `list_relationships`) — the primary defect site. Align its query-building with the HTTP handler.
- `src/api/routes/relationships.ts` (or equivalent handler backing `POST /list_relationships`) — reference implementation; do not change behavior, only extract/share logic if a common query-builder is warranted.
- Shared relationship query module (e.g. `src/db/queries/relationships.ts` or `src/services/relationshipService.ts`) — if `list_relationships` logic is duplicated between MCP and HTTP today, this is where to consolidate into one function both surfaces call, so future drift is structurally impossible.
- Test files: `test/integration/relationships/list_relationships.test.ts` (MCP surface) and `test/integration/http/list_relationships.test.ts` (HTTP surface), or the repo's equivalent integration test locations — add alongside existing relationship tests rather than new top-level files where a suite already exists.

**Data/contract changes**
- None. No schema, OpenAPI, or MCP tool-schema changes. This is an implementation-parity bug fix only — confirm the MCP tool's declared input/output schema already supports `source_entity_id`, `target_entity_id`, `relationship_type` (per the issue repro, it does; only the query execution is broken).

**Layering**
- If MCP and HTTP currently each implement their own query logic: consolidate into a single shared query function that both call, parameterized by filters (`source_entity_id?`, `target_entity_id?`, `relationship_type?`, `include_deleted?`), so this class of bug (one surface diverging from the other) cannot recur silently. This is the preferred fix shape given PM's cross-surface parity requirement.
- If MCP already calls the same underlying function as HTTP and the bug is elsewhere (e.g. a filter applied post-query in the MCP handler that HTTP doesn't apply, or a default `include_deleted`/`is_live` mismatch): fix at that narrower point instead of forcing a consolidation that isn't warranted. Confirm which case applies during implementation before choosing the fix shape.

**Build-step checklist**
1. Reproduce locally: create a relationship via MCP `create_relationship`, then call MCP `list_relationships` with `source_entity_id` and separately with `target_entity_id` + `relationship_type`; confirm both return empty against a live edge (matches issue repro).
2. Diff the MCP handler's query construction against the HTTP `POST /list_relationships` handler's query construction to isolate the exact divergence (missing WHERE binding, wrong liveness column, wrong join, or default-value mismatch on `include_deleted`).
3. Fix the MCP handler so its filter application and liveness semantics (`is_live=1`, `include_deleted=false` default) match the HTTP path exactly. Prefer consolidating into one shared query function per Layering above if duplication is confirmed as the cause.
4. Add MCP integration test: create a live relationship, assert `list_relationships(source_entity_id=...)` returns it with matching `relationship_key`/`source_entity_id`/`target_entity_id`/`relationship_type`, `total >= 1`.
5. Add MCP integration test: multiple live `PART_OF` edges onto one target entity, assert `list_relationships(target_entity_id=..., relationship_type=PART_OF)` returns all of them with `total` matching the live count.
6. Add MCP integration test: idempotent re-`create_relationship` on the same triple (asserts `observation_count` increments, key unchanged), then re-run `list_relationships` and assert the edge is still returned (guards the exact false-negative from the issue).
7. Add or confirm existing HTTP integration test coverage for `POST /list_relationships` with the same three scenarios (source filter, target+type filter, post-idempotent-write read), asserting result parity with the MCP test's expected rows (same keys).
8. Add/confirm a soft-delete regression test: soft-delete a relationship, assert default (`include_deleted=false`) `list_relationships` on both surfaces excludes it — guards against regressing #1570.
9. Run full relationship test suite (unit + integration) on both surfaces; run `code-review` self-check on the diff before opening the PR.
10. If any generated artifacts derive from touched source (e.g. test catalog, OpenAPI types, MCP docs) — run the applicable `generate:*`/`validate:*` scripts and commit output in the same PR, as the last step before `gh pr create`.
11. Open PR referencing `closes #2205`, describing root cause, the fix, and the surfaces tested.

### QA / Test Plan

**Agent-facing surface:** MCP `list_relationships` (primary defect — handler in `src/server.ts` `listRelationships` ignores `source_entity_id` / `target_entity_id` while HTTP `POST /list_relationships` in `src/actions.ts` binds them). **Eval substrate:** integration tests (real DB + real `NeotomaServer` / HTTP app), not Tier-1 agentic_eval — this is deterministic handler parity, same class as `relationship_delete_discovery_mcp.test.ts` (#277).

**Coverage gap (why #2205 escaped):** `tests/integration/mcp_actions_matrix.test.ts` lists via `entity_id` only; `tests/integration/mcp_relationship_variations.test.ts` asserts DB rows, not the MCP tool; HTTP `source_entity_id`/`target_entity_id` filters are covered in `relationship_delete_discovery.test.ts` but have **no MCP mirror**. New tests must drive the **exact repro filters** agents use post-write.

#### Primary eval deliverable

Add `tests/integration/list_relationships_mcp_source_target_filters.test.ts` (#2205). Pattern: `NeotomaServer` + `create_relationship` / `list_relationships` via private handler call (mirror `relationship_delete_discovery_mcp.test.ts`); seed `task` + `plan` entities via `store` or direct insert; **no mocked relationship queries**.

| Test id | Surface | Sequence | Assertions (agent-observable) |
|---------|---------|----------|-------------------------------|
| `2205-mcp-source-filter` | MCP | `create_relationship(PART_OF, source=task, target=plan)` → `list_relationships({ source_entity_id: task })` | `relationships.length >= 1`; row matches `{ relationship_key, source_entity_id: task, target_entity_id: plan, relationship_type: "PART_OF" }`; `total >= 1`; `create` payload had `is_live: 1` |
| `2205-mcp-target-type-filter` | MCP | Seed **3** live `PART_OF` edges onto same `plan` from distinct tasks → `list_relationships({ target_entity_id: plan, relationship_type: "PART_OF" })` | `relationships.length === 3`; `total === 3`; every row has `target_entity_id === plan` and `relationship_type === "PART_OF"`; includes edge from repro task |
| `2205-mcp-idempotent-verify` | MCP | `create_relationship` (same triple) twice → second response `observation_count === 2`, same `relationship_key` → `list_relationships({ source_entity_id: task })` | Edge still returned (guards prod false-negative); `relationship_key` unchanged |
| `2205-mcp-source-target-pair` | MCP | After create → `list_relationships({ source_entity_id: task, target_entity_id: plan })` | Exactly 1 row; discovery path parity with HTTP (#277) |
| `2205-http-parity-source` | HTTP | Same seed as `2205-mcp-source-filter` via `POST /list_relationships` | Same `relationship_key` set as MCP test (shared helper `assertLiveEdgeKeysMatch(mcpJson, httpJson)`) |
| `2205-http-parity-target-type` | HTTP | Same seed as `2205-mcp-target-type-filter` | Same keys + `total` as MCP |
| `2205-mcp-soft-delete-regression` | MCP | Create edge → `delete_relationship` → `list_relationships({ source_entity_id, include_deleted: false })` | `relationships === []`, `total === 0`; `include_deleted: true` returns 1 row (#1570, no MCP regression) |

#### Regression guards (existing paths must stay green)

- [ ] **`entity_id` + direction legacy path:** extend `mcp_actions_matrix.test.ts` `list_relationships (MCP_SPEC.md 3.16)` — after create, assert `list_relationships({ entity_id: source, direction: "outbound" })` returns the edge (proves fix didn't break the only path previously tested).
- [ ] **HTTP unchanged:** `relationship_delete_discovery.test.ts`, `relationship_liveness_pagination.test.ts`, `list_relationships_pagination.test.ts` — all pass without modification (HTTP is reference impl; if any need edits, treat as unintended behaviour change).
- [ ] **Tenant isolation:** add one MCP case to `tenant_isolation_matrix.test.ts` — user A calls `list_relationships({ source_entity_id: userB.task, target_entity_id: userB.plan })` → `relationships === []` (mirror existing `entity_id` isolation rows at L168–188).

#### Edge cases for new query branches

- [ ] **`source_entity_id` only** — returns all live outbound edges from source (no `relationship_type`).
- [ ] **`target_entity_id` only** — returns all live inbound edges to target.
- [ ] **`relationship_type` only** (schema-permitted per `ListRelationshipsRequestSchema` refine) — returns typed live edges within tenant; MCP and HTTP agree on count.
- [ ] **True empty vs false empty:** query with valid but unrelated entity id → `relationships === []`, `total === 0` (distinguishes "no matches" from "filter ignored").
- [ ] **Combined filter no-match:** `list_relationships({ source_entity_id: task, relationship_type: "REFERS_TO" })` when edge is `PART_OF` → `[]` / `total === 0`.
- [ ] **If impl extracts shared query builder** (`src/services/relationships.ts` or equivalent): one focused unit test on the builder asserting WHERE clauses for `(source?, target?, type?, include_deleted?)` — integration tests above remain the sign-off eval.

#### Contract / no-drift

- [ ] **OpenAPI + MCP schema unchanged:** `tests/contract/openapi_schema.test.ts` `list_relationships tool schema (issue #340)` passes without snapshot edits; `tests/contract/contract_mcp_cli_parity.test.ts` still lists `list_relationships`.
- [ ] **Response shape:** MCP returns `{ relationships, total, limit, offset }`; HTTP additionally `total_count` — parity helper compares `relationships[*].relationship_key` only, not alias fields.

#### CI commands (QA sign-off eval run)

```bash
# New + targeted regression
npx vitest run tests/integration/list_relationships_mcp_source_target_filters.test.ts \
  tests/integration/mcp_actions_matrix.test.ts \
  tests/integration/relationship_delete_discovery.test.ts \
  tests/integration/relationship_delete_discovery_mcp.test.ts \
  tests/integration/relationship_liveness_pagination.test.ts \
  tests/security/tenant_isolation_matrix.test.ts

# Full relationship integration lane
npm run test:integration -- --grep 'list_relationships|relationship'
npm run test:contract
```

#### Definition of done (QA gate)

- [ ] All seven primary eval cases in `list_relationships_mcp_source_target_filters.test.ts` committed in PR closing #2205
- [ ] **Pre-fix:** `2205-mcp-source-filter` and `2205-mcp-target-type-filter` fail against current `main` (repro confirmed)
- [ ] **Post-fix:** full targeted command block green in CI `agentic_evals` / integration lane
- [ ] No OpenAPI/MCP schema diff in PR
- [ ] Phoenicurus sign-off: eval ids above + CI green, or `[BLOCKING]` with failing case name

**Not in scope:** Tier-2 `relationship_lifecycle.scenario.yaml` extension (exercises `entity_id` list path only today); CLI `GET /relationships` unless arch confirms shared defect during impl.

markmhendrickson:
Fixes #2205. Also fixes #1973.

## Root cause

`ListRelationshipsRequestSchema` accepts `source_entity_id` and `target_entity_id`, and its `.refine()` explicitly permits a call carrying only those. HTTP `POST /list_relationships` branches on both (`src/actions.ts`). The MCP handler did not — it queried `parsed.entity_id` alone:

```ts
.eq("source_entity_id", parsed.entity_id)   // undefined when only source_entity_id was passed
```

So a call filtered solely by `source_entity_id` ran `.eq("source_entity_id", undefined)` and matched nothing. The write path and the read path disagreed: `create_relationship` reported `is_live: 1` for an edge that `list_relationships` then said did not exist.

Two documented flows were broken outright:

- **Pre-delete type discovery.** The tool description tells callers to pass both endpoint ids to learn an edge's type before `delete_relationship`. That call returned empty for a live edge.
- **Typed out-edge traversal.** `source_entity_id` + `relationship_type` — "all `works_at` out of node N" — returned empty.

## Why this took five reports to fix

Filed as #2205, #1973, #2156, #369, and #277 over several months. Every symptom is a **silent empty result, never an error**, and a working fallback was always one call away: `entity_id` + `direction`, `GET /entities/<id>/relationships`, or `retrieve_graph_neighborhood`. Each reporter routed around it and kept moving, so the workaround propagated as institutional knowledge while the defect stayed put.

An empty list is indistinguishable from "no edges" without a second, differently-shaped call to contradict it — which is precisely why this survived five reports and why the fix ships with a legacy-path baseline test asserting the two views agree.

## The fix

Endpoint filters take precedence over the legacy `entity_id` + `direction` pattern, matching the HTTP handler's resolution order. The ordering and pagination tail is extracted into `finalizeRelationshipList` so the two paths cannot drift apart again.

## Tests

`tests/integration/mcp_list_relationships_source_target_filters.test.ts` — **4 of 6 fail on origin/main**, all 6 pass here:

- legacy `entity_id` + `direction` baseline (passes on main — proves the edges exist and the test harness is sound)
- `source_entity_id` alone, `target_entity_id` alone
- `source_entity_id` + `target_entity_id` (pre-delete discovery)
- `source_entity_id` + `relationship_type`, with a negative case
- tenant isolation: the new query path stays scoped to the authenticated user

## Downstream

Agent code carrying the `GET /entities/<id>/relationships` / `retrieve_graph_neighborhood` workaround can move back to `list_relationships` once this lands.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -3391,6 +3391,58 @@ export class NeotomaServer {
 
     const relationships: any[] = [];
 
+    // `source_entity_id` / `target_entity_id` are explicit endpoint filters and
+    // take precedence over the legacy `entity_id` + `direction` pattern, exactly
+    // as HTTP `POST /list_relationships` resolves them (src/actions.ts).
+    //
+    // Before #2205/#1973 this handler queried `parsed.entity_id` alone, so a
+    // call filtered only by `source_entity_id` ran `.eq("source_entity_id",
+    // undefined)` and matched nothing — the read path silently disagreed with
+    // the write path, and callers routed around it via
+    // `GET /entities/<id>/relationships` or `retrieve_graph_neighborhood`.
+    if (parsed.source_entity_id || parsed.target_entity_id) {
+      let query = db.from("relationship_snapshots").select("*").eq("user_id", userId);
+
+      if (!parsed.include_deleted) {
+        query = query.eq("is_live", 1);
+      }
+      if (parsed.relationship_type) {
+        query = query.eq("relationship_type", parsed.relationship_type);
+      }
+      if (parsed.source_entity_id) {
+        query = query.eq("source_entity_id", parsed.source_entity_id);
+      }
+      if (parsed.target_entity_id) {
+        query = query.eq("target_entity_id", parsed.target_entity_id);
+      }
+
+      const { data, error } = await query;
+
+      if (!error && data) {
+        relationships.push(
+          ...data.map(
+            (r: {
+              relationship_key: string;
+              source_entity_id: string;
+              snapshot?: unknown;
+              last_observation_at?: string;
+            }) => ({
+              ...r,
+              id: r.relationship_key,
+              // With an explicit endpoint filter there is no traversal origin to
+              // orient against, so direction reports the row's own role relative
+              // to the filtered source when one was given.
+              direction: parsed.source_entity_id ? "outbound" : "inbound",
+              metadata: r.snapshot,
+              created_at: r.last_observation_at,
+            })
+          )
+        );
+      }
+
+      return this.finalizeRelationshipList(relationships, parsed);
+    }
+
     // Query relationship_snapshots instead of relationships table
     if (normalizedDirection === "outbound" || normalizedDirection === "both") {
       let outboundQuery = db
@@ -3472,14 +3524,28 @@ export class NeotomaServer {
       }
     }
 
-    // Sort by last_observation_at DESC, tiebroken by relationship_key ASC
-    // (#1571). last_observation_at is mutable and ties for rows upserted in the
-    // same millisecond (batch ingestion); without a stable tiebreaker, tied
-    // rows can move between pages across successive calls, so pagination over
-    // this order can drop or duplicate rows. relationship_key is the snapshot
-    // primary key — a stable, unique tiebreaker that fully determines order.
-    // This matches the HTTP /list_relationships ordering (actions.ts) and the
-    // determinism invariant (docs/architecture/determinism.md). See issue #368.
+    return this.finalizeRelationshipList(relationships, parsed);
+  }
+
+  /**
+   * Shared ordering + pagination tail for `list_relationships`, applied
+   * identically to the endpoint-filter path (`source_entity_id` /
+   * `target_entity_id`) and the legacy `entity_id` + `direction` path so the
+   * two cannot drift (#2205).
+   *
+   * Sort by last_observation_at DESC, tiebroken by relationship_key ASC
+   * (#1571). last_observation_at is mutable and ties for rows upserted in the
+   * same millisecond (batch ingestion); without a stable tiebreaker, tied
+   * rows can move between pages across successive calls, so pagination over
+   * this order can drop or duplicate rows. relationship_key is the snapshot
+   * primary key — a stable, unique tiebreaker that fully determines order.
+   * This matches the HTTP /list_relationships ordering (actions.ts) and the
+   * determinism invariant (docs/architecture/determinism.md). See issue #368.
+   */
+  private finalizeRelationshipList(
+    relationships: any[],
+    parsed: { offset: number; limit: number }
+  ): { content: Array<{ type: string; text: string }> } {
     relationships.sort((a, b) => {
       const aTime = new Date(a.last_observation_at || a.computed_at || 0).getTime();
       const bTime = new Date(b.last_observation_at || b.computed_at || 0).getTime();
@@ -3490,9 +3556,9 @@ export class NeotomaServer {
     });
 
     // Soft-deleted edges were already excluded at the DB via the `is_live`
-    // predicate on each direction's query (#1570), so `relationships` holds
-    // only live edges on the default path (all edges when include_deleted).
-    // `total` therefore reflects the correct post-filter count.
+    // predicate on each query (#1570), so `relationships` holds only live edges
+    // on the default path (all edges when include_deleted). `total` therefore
+    // reflects the correct post-filter count.
     const paginated = relationships.slice(parsed.offset, parsed.offset + parsed.limit);
 
     return this.buildTextResponse({
--- a/tests/integration/mcp_list_relationships_source_target_filters.test.ts
+++ b/tests/integration/mcp_list_relationships_source_target_filters.test.ts
@@ -1,0 +1,148 @@
+/**
+ * MCP `list_relationships` source/target filters (neotoma#2205, #1973).
+ *
+ * `ListRelationshipsRequestSchema` accepts `source_entity_id` and
+ * `target_entity_id`, and the HTTP handler (`POST /list_relationships`,
+ * src/actions.ts) branches on both. The MCP handler
+ * (`NeotomaServer.listRelationships`, src/server.ts) does not: it queries
+ * `parsed.entity_id` only, so a call filtered solely by `source_entity_id`
+ * runs `.eq("source_entity_id", undefined)` and matches nothing.
+ *
+ * The result is a read path that disagrees with the write path — the defect
+ * reported five times (#2205, #1973, #2156, #369, #277) and worked around, for
+ * months, with `GET /entities/<id>/relationships` or
+ * `retrieve_graph_neighborhood`.
+ *
+ * Harness mirrors tests/integration/mcp_handler_cross_user_scoping.test.ts:
+ * construct a server, set `authenticatedUserId`, call the private handler via
+ * a cast, seed rows directly through `db`.
+ */
+
+import { randomUUID } from "node:crypto";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { db } from "../../src/db.js";
+import { NeotomaServer } from "../../src/server.js";
+
+const PFX = "mcp_lr_filters";
+
+describe("MCP list_relationships source/target filters (#2205, #1973)", () => {
+  let server: NeotomaServer;
+  const userId = randomUUID();
+
+  const sourceId = `${PFX}_src_${randomUUID().slice(0, 8)}`;
+  const targetId = `${PFX}_tgt_${randomUUID().slice(0, 8)}`;
+  const otherTargetId = `${PFX}_tgt2_${randomUUID().slice(0, 8)}`;
+
+  const relType = "PART_OF";
+  const relKey = `${relType}:${sourceId}:${targetId}`;
+  const otherRelKey = `${relType}:${sourceId}:${otherTargetId}`;
+  const relKeys = [relKey, otherRelKey];
+  const entityIds = [sourceId, targetId, otherTargetId];
+
+  async function seedEntity(id: string): Promise<void> {
+    await db.from("entities").insert({
+      id,
+      user_id: userId,
+      entity_type: "test",
+      canonical_name: id,
+    });
+  }
+
+  async function seedEdge(key: string, src: string, tgt: string): Promise<void> {
+    const ts = new Date().toISOString();
+    await db.from("relationship_snapshots").insert({
+      relationship_key: key,
+      relationship_type: relType,
+      source_entity_id: src,
+      target_entity_id: tgt,
+      schema_version: "1.0",
+      snapshot: {},
+      computed_at: ts,
+      observation_count: 1,
+      last_observation_at: ts,
+      provenance: {},
+      user_id: userId,
+      is_live: 1,
+    });
+  }
+
+  /** Call the private MCP handler and parse its JSON text payload. */
+  async function listRelationships(args: Record<string, unknown>) {
+    const res = await (server as any).listRelationships(args);
+    return JSON.parse(res.content[0].text) as {
+      relationships: Array<{
+        source_entity_id: string;
+        target_entity_id: string;
+        relationship_type: string;
+      }>;
+      total?: number;
+    };
+  }
+
+  beforeAll(async () => {
+    server = new NeotomaServer();
+    (server as any).authenticatedUserId = userId;
+
+    for (const id of entityIds) await seedEntity(id);
+    await seedEdge(relKey, sourceId, targetId);
+    await seedEdge(otherRelKey, sourceId, otherTargetId);
+  });
+
+  afterAll(async () => {
+    await db.from("relationship_snapshots").delete().in("relationship_key", relKeys);
+    await db.from("entities").delete().in("id", entityIds);
+  });
+
+  it("baseline: the legacy entity_id + direction filter finds the live edges", async () => {
+    const out = await listRelationships({ entity_id: sourceId, direction: "outgoing" });
+    expect(out.relationships).toHaveLength(2);
+  });
+
+  it("source_entity_id alone returns the live out-edges", async () => {
+    const out = await listRelationships({ source_entity_id: sourceId });
+    expect(out.relationships.length).toBeGreaterThan(0);
+    for (const r of out.relationships) {
+      expect(r.source_entity_id).toBe(sourceId);
+    }
+    expect(out.relationships).toHaveLength(2);
+  });
+
+  it("target_entity_id alone returns the live in-edges", async () => {
+    const out = await listRelationships({ target_entity_id: targetId });
+    expect(out.relationships).toHaveLength(1);
+    expect(out.relationships[0]!.target_entity_id).toBe(targetId);
+  });
+
+  it("source_entity_id + target_entity_id resolves one edge (documented pre-delete type discovery)", async () => {
+    const out = await listRelationships({
+      source_entity_id: sourceId,
+      target_entity_id: targetId,
+    });
+    expect(out.relationships).toHaveLength(1);
+    expect(out.relationships[0]!.relationship_type).toBe(relType);
+  });
+
+  it("source_entity_id + relationship_type traverses typed out-edges", async () => {
+    const out = await listRelationships({
+      source_entity_id: sourceId,
+      relationship_type: relType,
+    });
+    expect(out.relationships).toHaveLength(2);
+
+    const none = await listRelationships({
+      source_entity_id: sourceId,
+      relationship_type: "REFERS_TO",
+    });
+    expect(none.relationships).toHaveLength(0);
+  });
+
+  it("does not leak edges belonging to another user", async () => {
+    (server as any).authenticatedUserId = randomUUID();
+    try {
+      const out = await listRelationships({ source_entity_id: sourceId });
+      expect(out.relationships).toHaveLength(0);
+    } finally {
+      (server as any).authenticatedUserId = userId;
+    }
+  });
+});


neotoma-agent:
**BLOCKED**

## Triage & Gate-Status Board

| Gate | State | Owner | Source |
| --- | --- | --- | --- |
| pm | signed_off | Pavo | Neotoma issue `gate_status.pm` |
| ux | pending | Accipiter | Not satisfied in Neotoma issue `gate_status` |
| arch | pending | Waxwing | Not satisfied in Neotoma issue `gate_status` |

**Owner**
- Current parent issue owner: `arch` gate / Waxwing
- Next phase after pre-impl gates clear: implementation

**Labels**
- No PR labels applied by this check

**DoD**
- [x] Parent issue resolved from PR body: `Fixes #2205`
- [x] Authoritative Neotoma issue entity read: `ent_98ea2c4fcbef0b5ebf6804d2`
- [x] `gate_status` exists, so the legacy no-`gate_status` fail-open rule does not apply
- [x] `pm` is satisfied by stored `signed_off`
- [ ] All pre-impl gates are satisfied

[BLOCKING] gate-inheritance: PR review cannot proceed until unsigned pre-impl gates are cleared: `ux` owned by Accipiter, `arch` owned by Waxwing.

**Resolution path**
- Operator override: `/confirm-gates-clear` — only @markmhendrickson may issue this command; it waives all unsigned pre-impl gates and re-triggers the PR pipeline. No other commenter can clear gates.
- Normal path: the gate owner can sign off the gate via Neotoma (`gate_status.<gate>` → `signed_off`) or the operator can waive it.

**Verdict**
- PR gate inheritance remains blocked; Vanellus is not assigned yet.

📎 Neotoma: [issue ent_98ea2c4fcbef0b5ebf6804d2](https://neotoma.markmhendrickson.com/entities/ent_98ea2c4fcbef0b5ebf6804d2)

neotoma-agent:

neotoma-agent:
review:arch
**SIGNED_OFF**

**Decision**
- Evaluate whether PR #2274 is an implementation-parity fix for the existing `list_relationships` contract, or a contract/schema change requiring OpenAPI, generated types, mapping, and instruction updates.

**Options + tradeoffs**
- Chosen: patch the MCP handler to honor already-declared `source_entity_id` / `target_entity_id` filters and share only the ordering/pagination tail inside `src/server.ts`. Coupling cost is low, blast radius is small, and reversibility is straightforward because the PR changes no public schema.
- Non-chosen: extract one shared relationship-list query service used by HTTP, MCP, and CLI. This would better prevent future drift, and may be correct if another relationship read divergence appears, but it carries higher regression risk for a P0 read-path bug and is not necessary to restore the declared contract here.

**Contract / schema / layering**
- OpenAPI already declares `POST /list_relationships` with `entity_id`, `source_entity_id`, `target_entity_id`, `relationship_type`, `include_deleted`, `limit`, and `offset`; `src/shared/action_schemas.ts` already accepts the same request shape.
- `src/shared/contract_mappings.ts` already maps `listRelationshipsForEntity` to adapter `both`, MCP tool `list_relationships`, and CLI command `relationships list`; no new operation or response field is introduced by this diff.
- `src/tool_definitions.ts` already derives the MCP input schema from OpenAPI and the MCP tool description already instructs source/target discovery before delete; the PR makes `src/server.ts` conform to that existing contract rather than widening it.
- Tenant isolation is present in the new branch via `.eq("user_id", userId)`, matching the guardrail in `docs/architecture/change_guardrails_rules.mdc` MUST 5 and the relationship tenant-isolation advisory class.
- No mutating operation is added or changed, so `idempotency_key` requirements are not applicable to this PR.

**Findings**
[NON-BLOCKING] docs: Existing CLI agent quick-reference still shows `neotoma relationships list --entity-id <entity_id>`, while the CLI command exposes `--source-entity-id` / `--target-entity-id` and no `--entity-id` flag. This is not introduced by #2274 and does not block the MCP fix, but it is an agent-instruction drift under `docs/developer/agent_instructions_sync_rules.mdc`; a follow-up should update `docs/developer/cli_agent_instructions.md` / `docs/developer/cli_reference.md` so the backup CLI path points agents at the same source/target filter contract.

**Diff-only evidence**
- Reviewed diff only; I did not execute PR code.
- The diff adds MCP coverage for `source_entity_id`, `target_entity_id`, combined source+target discovery, type filtering, legacy `entity_id` baseline, and cross-user isolation.
- Existing baseline evidence in the repo covers HTTP discovery/tenant isolation and CLI source/target list filters, so the multi-surface contract is represented without requiring this PR to churn OpenAPI or generated artifacts.

**DoD**
- [x] OpenAPI-first rule checked; no OpenAPI change required because this is not a contract-shape change.
- [x] `contract_mappings.ts` parity checked; existing mapping covers the operation.
- [x] MCP instruction/tool-description coherence checked; existing tool description already tells agents to use source/target discovery.
- [x] Tenant isolation checked in the diff branch and covered by the new MCP test.
- [x] No `idempotency_key` review needed; no mutating operation changed.

**Verdict**
SIGNED_OFF from the arch lens. No blocking findings.

---
📎 Neotoma: [neotoma#2205](https://neotoma.markmhendrickson.com/entities/ent_d7fe3ebe36524a72b68f9aeb)

neotoma-agent:
review:ux
**REQUEST_CHANGES**

Reviewed diff-only (no checkout available to execute) against the `ux` lens: discoverable naming, error messages with actionable hints, docs/examples for the new surface. Findings below are grounded in what's literally in the diff — no runtime claims.

[BLOCKING] misleading-field-value: `direction` is hardcoded and mislabels the documented pre-delete discovery flow

In `src/server.ts`, the new endpoint-filter branch sets:
```ts
direction: parsed.source_entity_id ? "outbound" : "inbound",
```
When a caller supplies **both** `source_entity_id` and `target_entity_id` — the exact "pre-delete type discovery" flow this PR's own description calls out as one of the two broken flows being fixed, and which the tool description reportedly instructs callers to use — the returned row is unconditionally labeled `"outbound"`, purely because `source_entity_id` happens to be truthy. There is no traversal origin in this call shape (both endpoints are pinned), so `"outbound"` isn't just imprecise, it's not a real answer to "which direction is this edge relative to my query." A caller consuming this field programmatically (e.g. branching on direction before calling `delete_relationship`) gets a confidently wrong signal on the flow this PR is explicitly fixing for. This is a defect in the code as written, not a runtime hypothesis — visible directly in the ternary.

Fix: when both endpoint ids are supplied, `direction` should be omitted, `null`, or a distinct value (e.g. `"pinned"`/`"exact"`) rather than reusing the traversal-semantics vocabulary (`outbound`/`inbound`) for a call shape that has no traversal direction.

[NON-BLOCKING] silent-precedence: mixing legacy and endpoint filters produces no signal that `entity_id`/`direction` were ignored

The new branch triggers on `parsed.source_entity_id || parsed.target_entity_id` and never inspects `parsed.entity_id`/`parsed.direction` again once it does. If a caller passes all four (plausible from an agent that doesn't fully internalize which filter mode it's in), `entity_id`/`direction` are silently discarded with no acknowledgment in the response. This is adjacent to the exact failure class the PR's own body identifies as the reason this bug went unreported correctly for 5 cycles ("every symptom is a silent empty result, never an error") — precedence is now correct, but the *shadowing* is still silent. Unverified whether this combination is actually reachable/common in practice; flagging because the pattern (silent behavior under ambiguous input, no hint) is the same one this PR is fixing elsewhere. Suggest: if both filter modes are present in one call, either error with an actionable message ("source_entity_id/target_entity_id take precedence over entity_id/direction; drop one set") or echo which mode was used in the response.

[NON-BLOCKING] docs-gap: no migration note for callers who adopted the `GET /entities/<id>/relationships` / `retrieve_graph_neighborhood` workaround

The PR body states this defect was rediscovered 5 times (#2205, #1973, #2156, #369, #277) over months precisely because each reporter routed around it via a workaround and moved on — meaning workaround code is now load-bearing in an unknown number of callers. The diff touches only `src/server.ts` and adds a test; no tool description, changelog, or docs update tells a caller "this now works, you can drop the workaround." Given the PR's own diagnosis of *why* this bug persisted (silent failure + easy workaround = no pressure to report correctly upstream), the same dynamic will suppress adoption of the fix unless it's surfaced somewhere a caller would see it (tool description changelog note, or at minimum the PR/release notes linked from the tool's docs surface).

---
📎 Neotoma: none linked — diff-only PR review, no entity created for this pass.

neotoma-agent:
review:pm
**REQUEST_CHANGES**

Scope check against the pm-signed acceptance criteria on #2205 (the issue this PR fixes; #1973 is a duplicate report with no separate spec, so #2205 governs).

- [x] Source filter alone returns live edges (test: `source_entity_id alone returns the live out-edges`)
- [x] Target filter alone / target+type traversal covered (`target_entity_id alone`, `source_entity_id + relationship_type`)
- [x] Pre-delete type discovery: `source_entity_id` + `target_entity_id` resolves the edge
- [x] No contract drift — fix is confined to `src/server.ts` handler internals; no schema/OpenAPI/MCP tool-schema changes
- [x] Cross-surface parity (MCP) — new integration test drives the MCP tool directly
- [ ] **Cross-surface parity (HTTP) — missing**

[BLOCKING] scope-gap: pm-signed acceptance criteria on #2205 explicitly require "Cross-surface parity (HTTP): An integration test drives `POST /list_relationships` with the same filters and asserts equivalent live-edge results (same keys and endpoints)." The diff touches only `src/server.ts` (MCP handler) and adds only `tests/integration/mcp_list_relationships_source_target_filters.test.ts`. No HTTP-side test file is present in the diff, and the PR description's "Tests" section only enumerates MCP-surface cases. Per standing policy `cross_surface_contract_parity_tested_all_surfaces` (ent_2ad0677fe23c0c1878ae43e8), any capability exposed on more than one surface needs tests on every exposing surface, each in that surface's natural call shape — this PR ships one of two. This is evident from the diff/file list itself (absent artifact), not a runtime claim.
  - Note: the PR body states the HTTP path (`src/actions.ts`) is already correct and used as the reference implementation for the fix's resolution order — so this is very likely a missing test, not a missing fix. That narrows the gap but does not close the acceptance-criteria checkbox as written.

[NON-BLOCKING] scope-gap: the "Effect — idempotent verify" criterion on #2205 (idempotent re-`create_relationship` on the same triple still returns the edge on a subsequent `list_relationships`) has no corresponding case in the new test file. If covered by pre-existing tests elsewhere, note which; otherwise this checkbox is also open. Flagging non-blocking since I cannot confirm from the diff alone whether coverage exists elsewhere in the suite.

No unrequested scope creep observed — the `finalizeRelationshipList` extraction is a same-behavior refactor in direct service of the "two paths cannot drift apart again" goal stated in both the issue and PR description, not an added feature. User-visible behavior (MCP `list_relationships` filter semantics) matches what #2205 asked for on the MCP surface.

📎 Neotoma: harness_event / plan_contribution references omitted — not yet public per Phase 3a-0; see ent_2ad0677fe23c0c1878ae43e8 (policy) in prose above.

neotoma-agent:
review:qa
**APPROVE**

Verified, not just read. On this checkout I ran the new spec against both the PR branch and an isolated revert of `src/server.ts` back to the pre-fix parent commit (831254a67, confirmed via `git diff 831254a67 0c1c0ce34^` as an empty diff), keeping only the new test file in place.

**PR-branch run:**
```
npx vitest run tests/integration/mcp_list_relationships_source_target_filters.test.ts
✓ tests/integration/mcp_list_relationships_source_target_filters.test.ts (6 tests) 21ms
Test Files  1 passed (1)
     Tests  6 passed (6)
```

**Pre-fix `server.ts` (parent commit) + same test file:**
```
FAIL … target_entity_id alone returns the live in-edges — expected 1, got 0
FAIL … source_entity_id + target_entity_id resolves one edge (documented pre-delete type discovery) — expected [] to have length 1
FAIL … source_entity_id + relationship_type traverses typed out-edges — expected [] to have length 2
Test Files  1 failed (1)
     Tests  4 failed | 2 passed (6)
```

This is an exact match for the PR description's claim ("4 of 6 fail on origin/main, all 6 pass here"). The regression test is real, reproduces the documented defect precisely, and the fix closes it.

**Coverage assessment against the fixed surface:**
- Legacy `entity_id` + `direction` baseline — covered, and used correctly as a sanity control that the harness/seed data is sound.
- `source_entity_id` alone / `target_entity_id` alone — covered, each asserts result shape is scoped to the filter.
- `source_entity_id` + `target_entity_id` (pre-delete discovery flow named in the tool description) — covered.
- `source_entity_id` + `relationship_type`, with a negative case (no match) — covered.
- Tenant isolation on the new query path — covered (`does not leak edges belonging to another user`), and this is exactly the class of regression `docs/security/advisories/2026-05-21-relationship-endpoint-tenant-isolation.md` warns about for new query endpoints on user-owned tables. Good that it's asserted here, not assumed.

**Gaps found (verified by probe, not just reasoning) — non-blocking:**

[NON-BLOCKING] test-coverage: `include_deleted` is unexercised on the new endpoint-filter branch. I wrote a throwaway probe test seeding a soft-deleted edge (`is_live: 0`) and calling `listRelationships({ source_entity_id, include_deleted: true })` — it correctly returned the soft-deleted row (`"is_live":0` in the payload, 1 result). Behavior is correct, but the committed suite has no case that would catch a regression here (e.g. if a future edit dropped the `include_deleted` check on this branch, nothing would fail). Recommend a 7th case in `mcp_list_relationships_source_target_filters.test.ts` seeding a soft-deleted edge and asserting it's excluded by default and included with `include_deleted: true`.

[NON-BLOCKING] test-coverage: the `direction` field's value when both `source_entity_id` and `target_entity_id` are given is unasserted. Probed directly: it reports `"outbound"` (since `parsed.source_entity_id` is truthy in the ternary at `src/server.ts:3435`), which is a reasonable, deterministic choice, but no test pins it. A caller depending on `direction` semantics in the pre-delete-discovery flow has no regression guard if this ternary is later inverted or generalized incorrectly.

[NON-BLOCKING] test-coverage: no pagination (`limit`/`offset`) case on the new filter branch. `finalizeRelationshipList` is shared with the legacy path and pagination is exercised there (`list_relationships_pagination.test.ts`), so the ordering/slice logic itself is covered — but nothing seeds >1 page of results specifically through `source_entity_id`/`target_entity_id` to confirm the shared tail is actually reached and correctly bounds output on this branch. Low risk given the code path is identical, but it's the one part of "the two paths cannot drift apart again" claim that isn't directly exercised from the new-filter side.

Ran `tests/security/tenant_isolation_matrix.test.ts` alongside the target suite and got 7 failures; verified these are pre-existing and unrelated — same 7 failures reproduce against the pre-fix `server.ts` with no changes to that test file in this PR. Not this PR's responsibility, noting for the record only.

No blocking findings. The regression test is real, precisely targeted, and the fix demonstrably closes the documented defect (#2205, #1973). Sign-off is warranted; the three gaps above are worth a fast follow-up but don't block.

📎 Neotoma: see `docs/security/advisories/2026-05-21-relationship-endpoint-tenant-isolation.md` for the tenant-isolation regression class this PR's new test correctly guards against.

neotoma-agent:
**REQUEST_CHANGES**

Aggregated panel verdict for PR #2274 (parent issue #2205). Head SHA reviewed: `0c1c0ce34c3426202d03685db8dd7384c7f01b48`.

## Per-lens roll-up

- **arch** (Waxwing) — SIGNED_OFF, no blocking findings. Confirms this is an implementation-parity fix against the already-declared MCP/HTTP/CLI contract; no OpenAPI/schema change needed. 1 non-blocking doc-drift note (CLI quick-reference still shows stale `--entity-id` flag).
- **ux** (lens panelist) — **REQUEST_CHANGES**. See Blocking below.
- **pm** (Pavo) — **REQUEST_CHANGES**. See Blocking below.
- **qa** (Phoenicurus) — APPROVE, no blocking findings. Verified by direct execution: new test suite passes 6/6 on the PR branch and reproduces the documented regression (4/6 fail) when reverted to the pre-fix parent commit. 3 non-blocking coverage gaps noted (include_deleted, direction-field assertion, pagination on the new filter branch).

## Blocking

- **[BLOCKING] (ux)** `direction` is hardcoded to `"outbound"` whenever `source_entity_id` is present — including the documented pre-delete discovery flow where both `source_entity_id` and `target_entity_id` are supplied together, where there is no real traversal direction. This mislabels output on the exact flow this PR claims to fix. Verified directly from the diff.
- **[BLOCKING] (pm)** #2205's pm-signed acceptance criteria require an HTTP-surface (`POST /list_relationships`) integration test asserting parity with the MCP path (policy `cross_surface_contract_parity_tested_all_surfaces`, ent_2ad0677fe23c0c1878ae43e8). The diff only adds an MCP-surface test file (`tests/integration/mcp_list_relationships_source_target_filters.test.ts`); no HTTP test appears in the diff or the PR's "Tests" section. Likely a missing *test*, not a missing *fix* (HTTP path is treated as reference-correct in the PR body), but the pm checkbox isn't satisfied as-is.

## Non-blocking (follow-up, does not block this PR)

- Mixing legacy (`entity_id`/`direction`) and new (`source_entity_id`/`target_entity_id`) filters in one call silently drops the legacy params with no signal — same silent-failure pattern this PR fixes elsewhere.
- No docs/changelog note for callers who adopted the `GET /entities/<id>/relationships` workaround.
- CLI agent quick-reference doc drift (stale `--entity-id` flag).
- QA coverage gaps: `include_deleted` on the new filter branch, `direction` value when both source+target given, pagination through the new filter branch.
- pm: idempotent-verify acceptance criterion has no matching test case in the new file (unconfirmed whether covered elsewhere).

## Merge recommendation

**Not merging.** Gate inheritance blocks this regardless of `APIS_AUTONOMY_AUTO_MERGE=1`: parent issue #2205 has `pm: changes_requested` and `arch: pending` (writeback still catching up post-panel) in `gate_status`, and this aggregation itself found blocking findings on 2 of 4 lenses. Routing back to Gryllus (impl) to address both blocking items. `pr_review` gate set to `changes_requested` on the issue.

## Verdict

**REQUEST_CHANGES**

---
📎 Neotoma: [neotoma#2205](https://neotoma.markmhendrickson.com/entities/ent_d7fe3ebe36524a72b68f9aeb)

neotoma-agent (changes requested):
All storage clean (`unknown_fields_count: 0` throughout). Comment posted, gate written back, plan_contribution linked.

---

**REQUEST_CHANGES**

Aggregated panel verdict for PR #2274 (parent issue #2205), posted at [issuecomment-5495248449](https://github.com/markmhendrickson/neotoma/pull/2274#issuecomment-5495248449).

## Per-lens roll-up

| Lens | Verdict | Blocking? |
|---|---|---|
| arch (Waxwing) | SIGNED_OFF | No |
| ux | REQUEST_CHANGES | **Yes** |
| pm (Pavo) | REQUEST_CHANGES | **Yes** |
| qa (Phoenicurus) | APPROVE, verified by direct execution | No |

## Blocking

- **[BLOCKING] (ux)** `direction` hardcoded to `"outbound"` in the source+target pre-delete discovery flow, mislabeling output on the exact flow the PR claims to fix.
- **[BLOCKING] (pm)** Missing HTTP-surface (`POST /list_relationships`) integration test required by #2205's pm-signed acceptance criteria — only an MCP-surface test is present (policy `cross_surface_contract_parity_tested_all_surfaces`, ent_2ad0677fe23c0c1878ae43e8).

## Actions taken

- `gate_status` on issue ent_d7fe3ebe36524a72b68f9aeb (#2205): `pr_review` → `changes_requested`, `pm` → `changes_requested` (arch left `pending` — writeback for that lens's own gate didn't land due to a grant restriction on the waxwing identity; my correction covers pr_review/pm, which is what the routing decision hinges on).
- `current_owner` → `gryllus` (routed back for fixes).
- Stored a `plan_contribution` recording the aggregation, linked via REFERS_TO.
- **Did not merge.** Gate inheritance (pm/arch pending) plus these blocking findings override `APIS_AUTONOMY_AUTO_MERGE=1`.

🧠 Neotoma — [Vanellus PR #2274 review aggregation](https://neotoma.markmhendrickson.com/inspector/conversations/ent_78b7e1aa7d5eb81dc152d942)
- 🆕 Created: conversation, 2× conversation_message, plan_contribution (4)
- ✏️ Updated: [issue #2205](https://neotoma.markmhendrickson.com/inspector/entities/ent_d7fe3ebe36524a72b68f9aeb) — gate_status, current_owner (1)
- 🔎 Retrieved: issue #2205 (1)

`[vanellus] merge_decision: BLOCKED — 2 blocking findings (ux: hardcoded direction field; pm: missing HTTP-surface parity test) plus pending pm/arch pre-impl gates on parent issue #2205; routed back to Gryllus, not merged.`

neotoma-agent:

markmhendrickson:
Closing as a duplicate of #2298, which has been merged as `49106f4`.

Both PRs fix the same defect from #2205 — the MCP `list_relationships` handler ignoring `source_entity_id` / `target_entity_id`. #2298 was selected as the survivor on three counts:

- **Broader fix.** It also repairs the bare `relationship_type`-only call, which fails for the same reason and is likewise permitted by the shared schema's `.refine`. This PR handles the source/target filters only.
- **CI.** #2298 is green on all 11 checks; this branch fails `baseline`.
- **Test coverage.** #2298 asserts *equivalence* against the working `entity_id` + direction path over a non-empty fixture, so a partial fix wiring one filter and not the other still fails, and no case can pass vacuously on an empty set. It also verified red-on-revert (5 of 7 cases fail).

No work is lost — the merged fix covers this branch's scope in full. Thanks for the parallel attempt; the duplication itself is being reported to the operator as a dispatch finding.

Issue #2141, opened 2026-08-09
Health check passes while every DB-backed request hangs: /health does not exercise the database


markmhendrickson:
## Symptom

On a hosted instance, `/health` returns `200` in **0.37s** while **every request that touches the database hangs until the client gives up**:

```
/health            200 in 0.37s
single entity GET  000 in 40.0s   (timeout)
page render        000 in 40.0s   (timeout)
POST /entities/query  timeout
```

Fly reports the machine `started` with `1 total, 1 passing`. Nothing alerts. From the outside the instance looks healthy; in practice it serves nothing.

## Why this is worse than an outage

A hard down is visible — the proxy 503s, checks go critical, someone notices. This fails the other way: **the one signal everybody trusts is the one signal that still works.**

The practical consequence is that a rendered page shared with a third party silently stops loading while every dashboard stays green. I hit this while verifying links on a document about to go to a partner: `/health` said fine, the page timed out at 30s, and the entity was perfectly intact when read directly.

Related but distinct from the 2026-08-09 outage on the same instance (machine `started`, health check **critical**, proxy refusing to route). That one was at least detectable. This one is not.

## Root cause hypothesis

`/health` does not exercise the database. It answers from process state alone, so it cannot distinguish "the server is up" from "the server is up and can serve a request". Any condition that stalls DB access — connection-pool exhaustion, a lock, a slow or wedged SQLite/libSQL handle — leaves the check green indefinitely.

A contributing factor visible in the logs: the instance is taking a continuous stream of `tool_invocation` writes from harness hooks across several concurrent sessions. Whether that is the cause of the stall or merely coincident with it needs measuring, but a health check that stays green through it is the defect regardless.

## Requested fix

1. **Make `/health` touch the database.** A bounded, cheap read (`SELECT 1`, or a single indexed lookup) with a short timeout. If it cannot complete, the check must fail — that is the entire point of a health check.
2. **Consider a `/health/deep` vs `/health/live` split** if a liveness probe genuinely needs to stay independent of the DB: `live` for "the process exists", `ready` for "it can serve". Fly's `[[http_service.checks]]` should point at the readiness one, or the proxy keeps routing to an instance that cannot answer.
3. **Bound the DB call itself** so a stalled query returns an error instead of hanging a request for 40s+. A request that never terminates is indistinguishable from a network failure to every caller.

## Acceptance

[...]

PR #2276, opened 2026-09-01
fix(db): bound how long a read can hold a reader-pool slot (#2217)

### LINKED ISSUE #2217: Two slow reads exhaust the 2-worker reader pool and hang every query; no statement timeout reclaims a reader

markmhendrickson:
Two slow reads could make a hosted instance serve nothing — every entity type hung, including `limit=1` on a 7-row table, while `/health` stayed green at 0.18s because it answers from process state and never takes a reader.

Closes #2217

## The obvious fix does nothing

Raising the reader count is the cheapest-looking of the issue's four suggestions, and it was measured as ineffective. Slowest trivial `SELECT 1` under the same two slow reads:

| readers | slowest `SELECT 1` |
|---|---|
| 2 (current default) | 269.9 ms |
| 4 | 288.8 ms |
| 8 | 299.8 ms |

Anyone shipping that change would have shipped nothing and believed otherwise. Two things were actually wrong.

## 1. Dispatch was blind round-robin

`readerConnection()` advanced `nextReader = (nextReader + 1) % readers.length` with no check of whether the chosen worker was executing. Each worker runs statements synchronously and strictly FIFO, so a read handed to a busy worker waits out whatever that worker is already doing.

Isolated: occupy exactly ONE worker, leave seven idle, fire eight trivial reads. Seven return in ~1 ms; the eighth takes as long as the cross join, because the cursor wrapped back onto the occupied worker. Seven idle workers were available the whole time — which is precisely why adding workers added slots to be dealt onto rather than capacity.

Reads now go to the least-loaded reader. **Depth alone is the wrong key** when statement costs differ by three orders of magnitude: a reader holding one multi-second join outranks a reader holding two 1 ms statements on depth, so depth-first ordering keeps dealing reads behind the join — the original bug wearing a different hat. So a reader busy longer than 20 ms is treated as long-running and avoided in favour of any shorter-running reader whatever its depth. That threshold sits an order of magnitude above a healthy read (~1 ms, per #2267) and two below the queries that caused the outage.

## 2. Nothing bounded runtime, and cancellation was structurally impossible

No `statement_timeout`, `query_timeout`, `sqlite3_interrupt`, or `AbortSignal` anywhere in `src/`. `busy_timeout` covers lock contention only, and under WAL a read-only reader takes no conflicting lock, so it was inert for this scenario.

Cancellation was not merely missing: the worker protocol has no cancel opcode, and a worker only reads its next message *after* the current statement runs to completion — so a cancel message could never be observed in time to matter. **Terminate-and-respawn is the only lever the synchronous driver leaves.**

Reads on the pool now carry a budget (`NEOTOMA_DB_STATEMENT_TIMEOUT_MS`, default 30 s). On expiry the reader is terminated and respawns lazily on the next read. Readers are read-only, so tearing one down loses no committed state and the caller knows nothing was written.

Deliberately scoped: **the writer is exempt, and statements inside a transaction are never timed out.** Killing the writer mid-transaction would leave the caller unable to tell whether work committed — worse than a slow query.

## 3. Client cancellation

`dbAbortContext` binds a per-request `AbortSignal`, so a client that hangs up stops holding a reader instead of running to completion for a response nobody will read. Reads only, for the reason above. No-op on backends without a worker pool.

## 4. Saturation is now visible

Logs when every reader is busy and again on recovery, rate-limited to once a second. Previously indistinguishable from a healthy idle instance from outside the process — the defining failure class here, and why the outage was only found by asking.

## Verification

Tests were confirmed **red on origin/main** before the fix, in a clean worktree:

```
× does not deal a read onto a busy worker when the round-robin cursor wraps
  → expected 1778 to be less than 250
× keeps every read fast with two slow reads in flight at the default pool size
  → expected 2071 to be less than 250
× times out a runaway statement and reclaims the reader for later reads
  → promise resolved "{ n: 31996000 }" instead of rejecting
× abandons a read when the caller's signal aborts, freeing the reader
× rejects immediately when the caller's signal is already aborted
× reports reader-pool occupancy so saturation is observable

Tests  6 failed | 7 passed (13)
```

On this branch: **13 passed**, stable across four consecutive runs.

Full-suite diff against a clean `origin/main` worktree — failing sets are **byte-identical**, no regressions:

| | baseline (origin/main) | this branch |
|---|---|---|
| Test Files | 12 failed / 531 passed | 12 failed / 531 passed |
| Tests | 22 failed / 5341 passed | 22 failed / **5349** passed |

`comm` over the normalized `FAIL` lines returns empty in both directions. The 22 are pre-existing inspector/hook/fixture cases unrelated to this change. The commit used the hook's sanctioned `SKIP_TESTS=1` with a reason (keeping type-check, lint, format, and site-copy gates live), since the suite was run separately against both trees.

## Relationship to #2267

Complementary, not duplicate. #2267 makes individual queries cheap, reducing how often a reader is held long. It does not bound how long one *can* be held, and under blind round-robin a single pathological query still landed on unrelated requests. Built on merged main.

### DIFF
--- a/docs/operations/configuration.md
+++ b/docs/operations/configuration.md
@@ -32,7 +32,8 @@ Resolution order for the data directory and variables: a project-local `.env`, t
 | `NEOTOMA_DB_BACKEND` | DB driver: `sqlite` (synchronous, zero-config) or `libsql` (concurrent — statements run off the event loop via worker-hosted driver for local files, or @libsql/client for remote sqld/Turso, so slow queries can't freeze the server; recommended for hosted/agent-heavy/shared instances) | `sqlite` |
 | `NEOTOMA_DB_URL` | libsql connection URL (`file:` for embedded local, `http(s)://`/`libsql://` for remote sqld/Turso) | `file:{NEOTOMA_SQLITE_PATH}` |
 | `NEOTOMA_DB_AUTH_TOKEN` | Auth token for remote libsql connections | unset |
-| `NEOTOMA_DB_READER_WORKERS` | Read-only worker connections for the local `libsql` backend (WAL lets them run concurrently with the writer) | `2` |
+| `NEOTOMA_DB_READER_WORKERS` | Read-only worker connections for the local `libsql` backend (WAL lets them run concurrently with the writer). Reads go to the least-loaded reader, so raising this adds capacity rather than just slots | `2` |
+| `NEOTOMA_DB_STATEMENT_TIMEOUT_MS` | Per-statement budget for reads on that reader pool. On expiry the reader is terminated (the synchronous driver cannot be interrupted mid-statement) and respawns on the next read, so one runaway query cannot hold a pool slot indefinitely. Writes and statements inside a transaction are never timed out. `0` disables | `30000` |
 | `NEOTOMA_RAW_STORAGE_DIR` | Content-addressed source files | `{dataDir}/sources` |
 | `NEOTOMA_LOGS_DIR` / `NEOTOMA_EVENT_LOG_PATH` | Log directory and event log file | under `{dataDir}/logs` |
 | `NEOTOMA_HOST_URL` / `NEOTOMA_PUBLIC_BASE_URL` | Public URL of this instance | auto-discovered or unset |
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -27,6 +27,7 @@ import {
   getAttributionDecisionFromRequest,
 } from "./middleware/aauth_verify.js";
 import { attributionContext } from "./middleware/attribution_context.js";
+import { dbAbortContext } from "./middleware/db_abort_context.js";
 import { aauthAdmission, getAAuthAdmissionFromRequest } from "./middleware/aauth_admission.js";
 import { buildSessionInfo, normalizeSessionOrigin } from "./services/session_info.js";
 import { AttributionPolicyError, enforceAttributionPolicy } from "./services/attribution_policy.js";
@@ -1829,6 +1830,11 @@ app.use(
 // may nest a more specific `runWithRequestContext` — nested scopes
 // shadow outer ones exactly as intended.
 app.use(attributionContext());
+
+// Bind a per-request AbortSignal so DB reads stop consuming a reader-pool slot
+// when the client hangs up (#2217). Reads only — a write must not be abandoned
+// half-applied. No-op on backends without a worker pool.
+app.use(dbAbortContext());
 
 // Stronger AAuth Admission plan: resolve verified AAuth identities to
 // `agent_grant` entities and stamp `req.aauthAdmission` /
--- a/src/middleware/db_abort_context.ts
+++ b/src/middleware/db_abort_context.ts
@@ -1,0 +1,44 @@
+/**
+ * Request-cancellation context for database reads (issue #2217).
+ *
+ * A client that gives up — an HTTP connection closed, an MCP call cancelled —
+ * does not stop the server-side work it started. On the worker-hosted local
+ * backend that work is holding a reader-pool slot, and the pool is small: two
+ * abandoned reads were enough to make a hosted instance serve nothing while
+ * `/health` stayed green, because nothing reclaimed their readers.
+ *
+ * This middleware binds an AbortSignal that fires when the response closes
+ * before it finished, and runs the rest of the request inside
+ * {@link withDbAbortSignal}. Reads issued downstream pick the signal up
+ * ambiently; when it fires, the reader executing them is terminated and the
+ * slot returns to the pool immediately rather than at the query's natural end.
+ *
+ * Deliberately reads-only. Writes are not abandoned: a caller hanging up
+ * mid-write must not leave a half-applied mutation, so `routeWrite` and
+ * everything inside a transaction ignore the signal (see
+ * `worker_file_database.ts`). Backends other than the worker-hosted one ignore
+ * the context entirely, so mounting this is a no-op for them.
+ */
+
+import type { NextFunction, Request, RequestHandler, Response } from "express";
+import { withDbAbortSignal } from "../repositories/worker/worker_file_database.js";
+
+export function dbAbortContext(): RequestHandler {
+  return (req: Request, res: Response, next: NextFunction): void => {
+    const controller = new AbortController();
+
+    // `close` fires on both a completed response and an aborted one. Only the
+    // latter should cancel: aborting after a normal finish would tear down a
+    // reader that is legitimately serving the NEXT request, since the signal
+    // is consulted at dispatch time.
+    const onClose = () => {
+      if (!res.writableEnded) controller.abort();
+    };
+    res.once("close", onClose);
+    res.once("finish", () => res.removeListener("close", onClose));
+
+    withDbAbortSignal(controller.signal, () => {
+      next();
+    });
+  };
+}
--- a/src/repositories/worker/worker_file_database.ts
+++ b/src/repositories/worker/worker_file_database.ts
@@ -22,6 +22,26 @@
  *     RETURNING via .get()) fails with SQLITE_READONLY and is retried on the
  *     writer — no SQL parsing heuristics.
  *
+ * Two properties bound how long that pool can be tied up (issue #2217):
+ *
+ *   - BUSY-AWARE DISPATCH. Reads go to the *least loaded* reader, not to the
+ *     next one round-robin. Blind round-robin deals work onto an occupied
+ *     worker even when siblings are idle, and since each worker runs
+ *     statements synchronously and strictly FIFO, that query waits behind the
+ *     slow one — measured at 271 ms for a trivial `SELECT 1` with seven idle
+ *     workers available. Raising the pool size does not help; only choosing
+ *     an idle worker does.
+ *   - STATEMENT TIMEOUT. The synchronous driver cannot be interrupted
+ *     mid-statement: the worker only reads its next message after the current
+ *     one runs to completion, so no cancel opcode can be observed in time.
+ *     The only real lever is `terminate()`. So a request that exceeds
+ *     `statementTimeoutMs` (or whose caller aborts) terminates its worker and
+ *     rejects with WorkerDbTimeoutError / WorkerDbAbortError; the next request
+ *     lazily respawns. Readers are read-only, so tearing one down loses no
+ *     committed state. The writer is exempt by default (killing it mid
+ *     transaction would leave the caller unable to know whether work
+ *     committed), and requests inside a transaction are never timed out.
+ *
  * Transaction semantics match the other backends: the TransactionGate
  * serializes transactions and keeps unrelated statements out of the open
  * transaction window, and AsyncLocalStorage routes statements issued on the
@@ -32,6 +52,7 @@
 import { AsyncLocalStorage } from "node:async_hooks";
 import { createRequire } from "node:module";
 import { Worker } from "node:worker_threads";
+import { logger } from "../../utils/logger.js";
 import {
   NESTED_TRANSACTION_ERROR,
   normalizeParams,
@@ -161,14 +182,61 @@ export class WorkerDbCrashError extends Error {
   }
 }
 
+/**
+ * Thrown when a statement exceeds the per-statement timeout. The worker
+ * hosting it is terminated (the synchronous driver offers no other way to stop
+ * a running statement) so the pool slot is reclaimed immediately rather than
+ * held for the query's full natural duration — the #2217 failure, where two
+ * long reads made every subsequent read queue behind them indefinitely.
+ *
+ * Only ever raised for reads on the read-only reader pool, so no committed
+ * state is at risk: the caller knows nothing was written.
+ */
+export class WorkerDbTimeoutError extends Error {
+  constructor(
+    message: string,
+    readonly timeoutMs: number
+  ) {
+    super(message);
+    this.name = "WorkerDbTimeoutError";
+  }
+}
+
+/**
+ * Thrown when the caller's AbortSignal fires while a statement is in flight —
+ * e.g. the HTTP client hung up. Same mechanism as the timeout: the reader is
+ * terminated so abandoned work stops consuming a pool slot instead of running
+ * to completion for a response nobody will read.
+ */
+export class WorkerDbAbortError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "WorkerDbAbortError";
+  }
+}
+
+type PendingEntry = {
+  resolve: (v: unknown) => void;
+  reject: (e: Error) => void;
+  /** Release the timeout timer and abort listener; idempotent. */
+  cleanup: () => void;
+};
+
+/** Options accepted per request; both bound how long a worker may be held. */
+export type WorkerRequestOptions = {
+  /** Terminate the worker and reject if the statement outlives this. */
+  timeoutMs?: number;
+  /** Terminate the worker and reject when the caller gives up. */
+  signal?: AbortSignal;
+};
+
 class WorkerConnection {
   private worker: Worker | null = null;
   private nextId = 1;
-  private pending = new Map<
-    number,
-    { resolve: (v: unknown) => void; reject: (e: Error) => void }
-  >();
+  private pending = new Map<number, PendingEntry>();
   private closed = false;
+  /** `Date.now()` when this worker last went from idle to busy; 0 while idle. */
+  private busyStartedAt = 0;
 
   constructor(
     private readonly options: {
@@ -182,6 +250,25 @@ class WorkerConnection {
   ) {}
 
   /**
+   * How many statements this worker currently owns. The worker executes them
+   * strictly FIFO and synchronously, so this is queue depth, not parallelism —
+   * which is exactly what busy-aware dispatch needs to compare.
+   */
+  get inFlight(): number {
+    return this.pending.size;
+  }
+
+  /**
+   * When the current busy stretch began, as a `Date.now()` timestamp; 0 while
+   * idle. Dispatch uses it to break depth ties: at saturation every reader
+   * usually holds one statement, and the one that has been busy longest is the
+   * one most likely to stay busy.
+   */
+  get busySince(): number {
+    return this.busyStartedAt;
+  }
+
+  /**
    * Spawn (or respawn) the worker and wire its lifecycle. Lazy: a crashed
    * worker is only recreated when the next request arrives, so a permanently
    * failing DB doesn't spin in a respawn loop while idle.
@@ -193,7 +280,11 @@ class WorkerConnection {
       const entry = this.pending.get(reply.id);
       if (!entry) return;
       this.pending.delete(reply.id);
-      if (this.pending.size === 0) worker.unref();
+      entry.cleanup();
+      if (this.pending.size === 0) {
+        this.busyStartedAt = 0;
+        worker.unref();
+      }
       if (reply.ok) {
         entry.resolve(reply.result);
       } else {
@@ -227,23 +318,87 @@ class WorkerConnection {
 
   /** Reject all in-flight requests and drop the dead worker handle. */
   private failInFlight(error: Error): void {
-    for (const entry of this.pending.values()) entry.reject(error);
+    // Snapshot first: reject() runs caller continuations that may re-enter
+    // request() and repopulate the map while we are still iterating it.
+    const entries = [...this.pending.values()];
     this.pending.clear();
+    this.busyStartedAt = 0;
     this.worker = null;
+    for (const entry of entries) {
+      entry.cleanup();
+      entry.reject(error);
+    }
+  }
+
+  /**
+   * Kill the worker hosting a request that overran its budget, and reject
+   * everything it was executing.
+   *
+   * Terminating is the only lever available: the worker runs each statement
+   * synchronously to completion before it reads its next message, so a cancel
+   * opcode could not be observed until the statement it was meant to cancel
+   * had already finished. `failInFlight` runs FIRST so the victim's siblings
+   * reject with the same reason rather than a later confusing crash error, and
+   * so `this.worker` is cleared before `terminate()` fires the `exit` handler.
+   */
+  private abandon(error: Error): void {
+    const worker = this.worker;
+    this.failInFlight(error);
+    if (worker) void worker.terminate().catch(() => {});
   }
 
   request(
     op: "run" | "get" | "all" | "exec" | "pragma",
     sql: string,
-    params: unknown[] = []
+    params: unknown[] = [],
+    options: WorkerRequestOptions = {}
   ): Promise<unknown> {
     if (this.closed) {
       return Promise.reject(new WorkerDbCrashError("DB connection is closed"));
     }
+    const { timeoutMs, signal } = options;
+    if (signal?.aborted) {
+      return Promise.reject(new WorkerDbAbortError("DB request aborted before dispatch"));
+    }
     const worker = this.ensureWorker();
     const id = this.nextId++;
     return new Promise((resolve, reject) => {
-      this.pending.set(id, { resolve, reject });
+      let timer: ReturnType<typeof setTimeout> | undefined;
+      let onAbort: (() => void) | undefined;
+      const cleanup = () => {
+        if (timer !== undefined) {
+          clearTimeout(timer);
+          timer = undefined;
+        }
+        if (onAbort) {
+          signal?.removeEventListener("abort", onAbort);
+          onAbort = undefined;
+        }
+      };
+
+      if (this.pending.size === 0) this.busyStartedAt = Date.now();
+      this.pending.set(id, { resolve, reject, cleanup });
+
+      if (timeoutMs !== undefined && timeoutMs > 0) {
+        timer = setTimeout(() => {
+          this.abandon(
+            new WorkerDbTimeoutError(
+              `DB statement exceeded ${timeoutMs}ms and its worker was terminated to reclaim the pool slot`,
+              timeoutMs
+            )
+          );
+        }, timeoutMs);
+        // Do not let a pending timeout keep the process alive on its own.
+        timer.unref?.();
+      }
+
+      if (signal) {
+        onAbort = () => {
+          this.abandon(new WorkerDbAbortError("DB request aborted by caller"));
+        };
+        signal.addEventListener("abort", onAbort, { once: true });
+      }
+
       // Hold the process open while a request is in flight.
       worker.ref();
       worker.postMessage({ id, op, sql, params });
@@ -254,8 +409,59 @@ class WorkerConnection {
     this.closed = true;
     const worker = this.worker;
     this.worker = null;
+    for (const entry of this.pending.values()) entry.cleanup();
     if (worker) await worker.terminate();
   }
+}
+
+/**
+ * How long a reader must have been busy before dispatch treats it as running
+ * something long, and prefers a deeper-queued but recently-started reader over
+ * it.
+ *
+ * Sized against the two populations this pool actually sees. A healthy read is
+ * about 1 ms (#2267 measured a 40k-row count at ~1 ms); the reads that caused
+ * the outage ran for seconds. 20 ms sits an order of magnitude above the first
+ * and two below the second, so it separates them without needing to know
+ * either precisely — and a statement genuinely near the boundary costs little
+ * either way, since queueing behind a 20 ms statement is not the failure being
+ * prevented.
+ *
+ * The threshold exists because depth is the wrong primary key when statement
+ * costs differ by three orders of magnitude: a reader holding ONE multi-second
+ * cross join outranks a reader holding TWO 1 ms statements on depth, so
+ * depth-first ordering keeps dealing reads behind the cross join — the
+ * original bug wearing a different hat.
+ */
+const LONG_RUNNING_READER_MS = 20;
+
+/** How long `a` has been busy, or 0 if idle. */
+function busyForMs(connection: WorkerConnection, now: number): number {
+  return connection.busySince > 0 ? now - connection.busySince : 0;
+}
+
+/**
+ * Order two busy readers by how attractive they are to queue behind.
+ *
+ * 1. A reader busy for less than {@link LONG_RUNNING_READER_MS} beats one busy
+ *    for longer, whatever their depths. This is what keeps traffic off a
+ *    reader stuck on a runaway query — the measured failure.
+ * 2. Both long-running: prefer the one that started more recently; it has less
+ *    elapsed work behind it and is likelier to finish first.
+ * 3. Both short-running: prefer the shallower queue, then the more recent
+ *    start. At saturation every reader typically holds exactly one statement,
+ *    so depth ties constantly, and breaking that tie by array order or a
+ *    round-robin cursor is what put a query onto the busy worker while seven
+ *    siblings sat idle.
+ */
+function compareReaderLoad(a: WorkerConnection, b: WorkerConnection): number {
+  const now = Date.now();
+  const aLong = busyForMs(a, now) >= LONG_RUNNING_READER_MS;
+  const bLong = busyForMs(b, now) >= LONG_RUNNING_READER_MS;
+  if (aLong !== bLong) return aLong ? 1 : -1;
+  if (!aLong && a.inFlight !== b.inFlight) return a.inFlight - b.inFlight;
+  // Larger busySince == started later == likely to finish sooner.
+  return b.busySince - a.busySince;
 }
 
 function isReadonlyRejection(error: unknown): boolean {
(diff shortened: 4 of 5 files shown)


neotoma-agent:
**BLOCKED**

Parent issue [neotoma#2217](https://github.com/markmhendrickson/neotoma/issues/2217) has unsigned pre-impl gates. Gate inheritance requires `pm`, `ux`, and `arch` to be `signed_off`/`waived`/`not_required` before a PR can be reviewed/merged.

**Re-checked:** PR synchronize event `2026-09-01T19:11:00Z` — authoritative `gate_status` unchanged since prior block.

**Pending pre-impl gates:**

| Gate | Owner | State |
|---|---|---|
| `pm` | Pavo | pending |
| `arch` | Waxwing | pending |

(`ux` is `not_required` on this issue — satisfied.)

**Full gate status** (from Neotoma issue entity):

| Gate | State |
|---|---|
| pm | pending |
| ux | not_required |
| arch | pending |
| impl | pending |
| pr_review | pending |
| qa | pending |
| legal | not_required |

- [x] Parent issue entity loaded from Neotoma
- [x] Authoritative `gate_status` read (not inferred from GitHub)
- [ ] Pre-impl gates clear — **blocked**
- [ ] Vanellus assigned — deferred until gates clear

**Resolution paths:**
- The gate owner signs off via Neotoma (`gate_status.<gate> → signed_off`): Pavo for `pm`, Waxwing for `arch`.
- The operator waives a gate directly.
- **Operator override:** only `@markmhendrickson` may run `/confirm-gates-clear` — this waives all unsigned pre-impl gates on the parent issue and re-triggers the PR pipeline. No other commenter can clear gates.

📎 Neotoma: [neotoma#2217](https://neotoma.markmhendrickson.com/entities/ent_02ead63b89bef8a09db4ca82)

GATE_INHERITANCE: blocked
GATE_PENDING: pm,arch

neotoma-agent:
review:pm
**SIGNED_OFF**

## Problem
Two concurrent slow reads can occupy the entire reader pool with no runtime bound or client cancellation, causing every subsequent entity read to hang indefinitely while `/health` stays green (#2217).

## Scope vs #2217 intent

**In scope (delivered)**
- Statement timeout with terminate-and-respawn reader reclaim (`NEOTOMA_DB_STATEMENT_TIMEOUT_MS`, default 30s) — load-bearing fix #1
- Busy-aware reader dispatch (least-loaded + long-running avoidance) — operator-measured prerequisite; replaces ineffective “raise pool size” approach
- HTTP client cancellation via `dbAbortContext` — issue priority #2
- Saturation visibility via structured logging + `readerPoolStats()` — partial fix #4 (logs/stats only; `/health` gauge remains #2141 follow-on)

**Out of scope / correctly omitted**
- CPU-derived default `readerCount` — measured ineffective without busy-aware dispatch; default stays 2
- `/health` or metrics-endpoint saturation gauge — eng-spec follow-on tied to #2141

**Acceptance criteria (PM sign-off)**
- [x] Idle readers are used instead of dealing reads onto an occupied worker (round-robin wrap repro test)
- [x] Runaway read is rejected on budget and reader slot is reclaimed for subsequent reads (`WorkerDbTimeoutError` test)
- [x] Caller abort terminates in-flight read and frees reader (`WorkerDbAbortError` tests)
- [x] Writer and in-transaction statements are never timed out (transaction test)
- [x] Effect-verified tests assert observable latency/reclaim outcomes, not merely input acceptance (policy `fixed_means_behavior_verified_not_contract_accepted`)
- [x] New env var documented in `docs/operations/configuration.md`
- [x] Cross-surface parity N/A — internal DB worker pool only; no MCP/REST/CLI contract change

## Findings

[NON-BLOCKING] scope-consolidation: PR bundles eng-spec “follow-on” items (#2 cancellation, partial #4 logging) plus busy-aware dispatch not in the original engineering subsection. All items directly address the #2217 failure mode, are covered by tests, and align with the human issue body + operator measurements (2026-09-01 comment). Not unrelated scope creep.

[NON-BLOCKING] config-default: default timeout is 30s vs eng-spec’s suggested 10s. Plausible given observed >25–75s hangs; unverified in this diff-only review whether 10s would false-positive on legitimate slow queries.

No `[BLOCKING]` findings from PM lens. `gate_status.pm` signed off on parent issue.

---
📎 Neotoma: [neotoma#2217](https://neotoma.markmhendrickson.com/entities/ent_12ecb34d38cba594b8ab3129)

neotoma-agent:
review:arch
**SIGNED_OFF**

## Architecture Decision Record

**Decision:** Bound reader-pool slot hold time for `WorkerFileDatabase` via (1) busy-aware dispatch replacing blind round-robin, (2) terminate-and-respawn statement timeout on read-only readers, (3) ambient `AbortSignal` propagation for abandoned HTTP reads, with writer and in-transaction paths explicitly exempt.

**Options + tradeoffs**
| Option | Coupling | Reversibility | Ops cost |
|---|---|---|---|
| **A (chosen):** Main-thread timeout + `worker.terminate()` + lazy respawn | Low — contained in repository + one middleware | High — env toggles (`NEOTOMA_DB_STATEMENT_TIMEOUT_MS=0`), no schema/API surface | Small respawn cost per timeout; saturation logs |
| **B:** Cooperative cancel opcode inside worker | Tighter worker protocol | Medium | **Does not work** — synchronous driver blocks event loop mid-statement (documented in diff) |
| **C:** Scale default `readerCount` to CPU count | None | Medium — behavior change on unset env | Measured ineffective without busy-aware dispatch (#2217 repro) |
| **D:** OpenAPI `/health` saturation gauge (fix #4) | Cross-layer contract | N/A this PR | Correct follow-on tied to #2141 |

**Chosen approach:** A + busy-aware dispatch. Root cause was dispatch semantics (dealing onto occupied workers), not pool size alone. Terminate-and-respawn is the only reliable reclaim lever given the synchronous worker driver; writer/transaction exemption preserves commit semantics.

- **No public contract change** — no `openapi.yaml`, `contract_mappings.ts`, MCP tool, CLI command, entity schema, or response-shape delta (consistent with issue #2217 engineering spec: "entirely internal to the DB access layer").
- **Layering:** Repository owns timeout/abort/dispatch (`worker_file_database.ts`); HTTP boundary owns signal binding (`db_abort_context.ts` → `withDbAbortSignal` AsyncLocalStorage); middleware mounted after `attributionContext()` in `actions.ts`. Writes route through `routeWrite`/writer untimed — correct trust boundary.
- **Config surface:** `NEOTOMA_DB_STATEMENT_TIMEOUT_MS` added alongside existing `NEOTOMA_DB_*` pattern (parsed in constructor, documented in `docs/operations/configuration.md` — matches how `NEOTOMA_DB_READER_WORKERS` is handled; no separate config schema module in repo).
- **New internal error types** (`WorkerDbTimeoutError`, `WorkerDbAbortError`) follow existing `WorkerDbCrashError` pattern — internal `Error` subclasses, not a parallel HTTP error envelope. Callers see the same generic DB-error path as today.

**Reversibility:** High. Timeout disable via env `0`; dispatch logic is localized; middleware is no-op on non-worker backends. No migration.

**Steelman (non-chosen option):** CPU-scaled default reader count (option C) is cheap and helps genuine saturation once dispatch is fixed — worth revisiting if saturation logs stay sustained after merge.

## Arch-lens checklist

- [x] **Contract-first:** No HTTP/MCP/CLI contract touched — OpenAPI/`contract_mappings` update not required (`docs/architecture/change_guardrails_rules.mdc` touchpoint matrix: internal DB layer only).
- [x] **Layering:** Repository ↔ middleware separation respected; transaction and write paths exempt from timeout/abort.
- [x] **Schema / response fields:** N/A — `readerPoolStats()` is `@internal` test/diagnostic seam, not a declared API field.
- [x] **Tenant isolation:** N/A — pool management, no entity lookup path changes.
- [x] **Idempotency:** N/A — no new mutating store operations.
- [x] **Cross-surface parity:** N/A — capability not exposed on HTTP/MCP/CLI.
- [x] **Agent-instruction coherence:** N/A — no agent-facing interface change.

## Findings

[NON-BLOCKING] **error-envelope:** `WorkerDbTimeoutError`/`WorkerDbAbortError` surface as generic internal errors (same as `WorkerDbCrashError`), not canonical `ErrorEnvelope` codes like `STATEMENT_TIMEOUT`. Consistent with existing worker-layer taxonomy; unverified whether HTTP status should be 504 vs 500 — would need runtime observation on a timed-out request.

[NON-BLOCKING] **transport-coverage:** `dbAbortContext` mounts on the Express stack only. MCP cancellation propagation is claimed in comments but not evidenced in diff — unverified whether MCP transport binds an equivalent `AbortSignal`. Follow-on if MCP sessions can hold readers after client disconnect.

[NON-BLOCKING] **observability:** Saturation is log-only (`noteSaturation`); metrics/`/health` gauge remains #2141 scope per issue spec.

Arch gate signed off on parent issue #2217.

---
📎 Neotoma: [neotoma#2217](https://neotoma.markmhendrickson.com/entities/ent_12ecb34d38cba594b8ab3129)

neotoma-agent:
review:security
**COMMENT**

Adversarial pass on reader-pool bounding (#2217). Diff-only review — no branch checkout, no runtime execution. Evidence bar: `[BLOCKING]` reserved for defects evident from the diff; availability hypotheses marked `[NON-BLOCKING] PLAUSIBLE`.

## Sinks enumerated

**Pattern:** unbounded reader-pool occupancy / blind round-robin dispatch on `WorkerFileDatabase` local-file libsql backend.

| Sink | Location | Covered by PR? |
|------|----------|----------------|
| `WorkerStatement.get()` / `.all()` → `routeRead()` | `worker_file_database.ts` | **Yes** — timeout + abort signal |
| `WorkerConnection.request()` (direct) | same file | **Partial** — only when caller passes `WorkerRequestOptions`; production reads go through `routeRead` |
| `routeWrite()` / `exec()` / `pragma()` / `prepare().run()` | same file | **No** — intentional; writer never timed out or aborted |
| Reads inside `transaction()` (`txContext`) | `routeRead()` early branch | **No** — intentional; writer, no budget |
| `isReadonlyRejection` writer retry | `routeRead()` catch | **No** — falls back to untimed writer |
| `LibsqlDatabase` (remote `http(s)://` / `libsql://`) | `libsql_driver.ts` | **No** — different backend, no pool bounding |
| `AsyncSqliteDatabase` (`NEOTOMA_DB_BACKEND=sqlite`) | `sqlite_driver.ts` via `connection.ts` | **No** — synchronous backend, fix is no-op |
| `dbAbortContext()` HTTP cancel | `actions.ts` only | **Partial** — mounted on Express app |
| `withDbAbortSignal()` manual bind | exported from `worker_file_database.ts` | **Only where mounted** — not in `index.ts` (stdio MCP) or `openclaw_entry.ts` |

All production read paths on `WorkerFileDatabase` funnel through `prepare().get/all` → `routeRead`; no uncovered sibling export bypasses the guard on that backend.

## Encodings tested

Not applicable — guard is temporal (statement budget, HTTP disconnect), not string/address normalization. Reviewed config input domain instead:

| Input | Parsed behavior | Reaches unbounded sink? |
|-------|-----------------|-------------------------|
| `NEOTOMA_DB_STATEMENT_TIMEOUT_MS` unset | defaults 30000 | No |
| `"0"` | `Math.max(0,0)` → timeout disabled | **Yes** — reads unbounded until natural completion (documented operator fail-open) |
| negative / non-numeric | `readEnvInt` → undefined → default 30000 | No |
| `NEOTOMA_DB_READER_WORKERS=0` | all reads on writer, no reader pool | Writer path has no statement timeout |

## Fail-open branches

Each path that yields **ALLOW** (continue / no reclaim) rather than bounded execution:

1. **`statementTimeoutMs === 0`** — timer not armed (`timeoutMs > 0` guard in `WorkerConnection.request`). Operator-configured; documented in `configuration.md`.
2. **`txContext.getStore()`** — transaction reads/writes on writer without timeout or abort signal. Intentional (atomicity).
3. **`routeWrite` and writer retry after `isReadonlyRejection`** — no `WorkerRequestOptions` passed. Intentional.
4. **`abortContext.getStore()` undefined** — no abort listener. Occurs on stdio MCP (`index.ts`), OpenClaw plugin (`openclaw_entry.ts`), and any code outside `dbAbortContext()` middleware. Statement timeout still applies on reader pool.
5. **Non-`file:` libsql and `sqlite` backends** — entire worker-pool mechanism absent; availability fix does not apply.
6. **`readEnvInt` invalid values** — fall through to safe defaults (fail-closed to 30s timeout, 2 readers).

## Findings

**[NON-BLOCKING] PLAUSIBLE availability — uncovered backends:** Remote `LibsqlDatabase` and default `sqlite` backend have no statement timeout, busy-aware dispatch, or abort reclaim. Deployments on `NEOTOMA_DB_BACKEND=libsql` with `file:` URLs (the reported outage shape) are covered; remote-Turso or `sqlite` backend operators retain the pre-fix DoS surface. **Confirm/refute:** deploy slow-read soak against each backend variant.

**[NON-BLOCKING] PLAUSIBLE availability — stdio/OpenClaw entry points without abort middleware:** `index.ts` and `openclaw_entry.ts` invoke `NeotomaServer` directly with no `dbAbortContext()`. Abandoned stdio clients cannot trigger reader reclaim via HTTP `close`; only the 30s statement budget applies. **Confirm/refute:** start slow `retrieve_entities` over stdio MCP, kill client mid-flight, observe reader-pool stats.

**[NON-BLOCKING] PLAUSIBLE availability — persistent HTTP MCP sessions:** `dbAbortContext` aborts on `res.close` when `!res.writableEnded`, not on per-RPC cancellation while the StreamableHTTP session stays open. A client that keeps the session alive but abandons an in-flight tool call may hold a reader until statement timeout. **Confirm/refute:** cancel single MCP `CallTool` without closing transport; measure reader hold time.

**[NON-BLOCKING] PLAUSIBLE availability — transaction-scoped slow reads:** Reads inside `db.transaction()` bypass timeout entirely (tested in PR). If any authenticated path can open a transaction wrapping an expensive read, the single writer blocks all mutations for unbounded duration. **Confirm/refute:** trace longest transaction read paths in store/merge flows; not evident as attacker-controlled from diff alone.

**[NON-BLOCKING] PLAUSIBLE availability — writer saturation via pool cap:** At full reader saturation, new reads queue on the “least bad” busy reader (`compareReaderLoad`). They still carry timeout/abort, so slots eventually reclaim, but latency amplification under adversarial concurrent load remains plausible until timeout fires (default 30s × pool depth). Mitigated vs pre-fix indefinite hold; not a regression.

No `[BLOCKING] CONFIRMED` findings from diff-only evidence. Intentional writer/transaction exemptions are documented and consistent with atomicity requirements; they do not introduce a new authz bypass or data-integrity fail-open.

## Verdict

**COMMENT** — availability hardening looks sound for the `WorkerFileDatabase` / HTTP-hosted path it targets. Residual DoS surface lives on alternate backends and entry points noted above; none are contradictions visible in the diff itself.

---
📎 Neotoma: [neotoma#2217](https://neotoma.markmhendrickson.com/entities/ent_issue_2217) · [neotoma#2276](https://github.com/markmhendrickson/neotoma/pull/2276)

neotoma-agent:
review:ux
**SIGNED_OFF**

Diff-only review — no branch checkout, no runtime execution. Evidence bar: `[BLOCKING]` reserved for defects evident from the diff.

## User goal

**Diagnose and recover when hosted DB reads saturate or hang** — without silent failure while `/health` stays green. Primary personas: instance operator (env tuning, log triage) and agent developer (HTTP/MCP client seeing DB errors).

## User-facing surface

Internal DB layer only. Operator/agent-visible surfaces in this diff:

| Surface | Where | Audience |
|---|---|---|
| `NEOTOMA_DB_STATEMENT_TIMEOUT_MS` | `docs/operations/configuration.md` | Operator |
| Updated `NEOTOMA_DB_READER_WORKERS` description | same | Operator |
| Saturation / recovery log lines | `worker_file_database.ts` `noteSaturation()` | Operator |
| `WorkerDbTimeoutError` / `WorkerDbAbortError` messages | `worker_file_database.ts` → HTTP `DB_QUERY_FAILED` envelope | Agent developer |
| HTTP client-disconnect reclaim | `db_abort_context.ts` middleware on Express stack | Agent developer (HTTP) |

## Interaction / flow

1. **Normal read** → least-loaded reader dispatch → completes within budget → no operator signal.
2. **Pool saturated** → warn log (rate-limited) with duration/depth → recovery info log when cleared.
3. **Runaway read** → worker terminated → caller receives `WorkerDbTimeoutError` message (surfaces as `DB_QUERY_FAILED` with descriptive text).
4. **Client disconnect (HTTP)** → `AbortSignal` fires on `res.close` before `finish` → reader reclaimed via `WorkerDbAbortError`.
5. **Write / in-transaction read** → timeout and abort explicitly bypassed (documented in config row and code comments).

## Discoverability & naming

- [x] `NEOTOMA_DB_STATEMENT_TIMEOUT_MS` follows existing `NEOTOMA_DB_*` convention and sits adjacent to `NEOTOMA_DB_READER_WORKERS` in the config table.
- [x] Config row documents default (`30000`), disable semantics (`0`), and writer/transaction exemption — sufficient for a senior engineer to tune without reading source.
- [x] `NEOTOMA_DB_READER_WORKERS` description updated to reflect busy-aware dispatch, correcting the prior mental model where “raise pool size” appeared ineffective.

## Error & empty states

- [x] Timeout and abort paths **reject** with typed errors and explicit messages (not silent hang). Example timeout text: `DB statement exceeded ${timeoutMs}ms and its worker was terminated to reclaim the pool slot`.
- [x] Saturation is no longer invisible — structured warn/info logs replace the pre-fix “healthy-looking idle instance” failure class.
- N/A: no UI empty states (infra-only change).

## Accessibility

N/A — no UI components in diff.

## Findings

[NON-BLOCKING] operator-guidance: Saturation warn log cites `NEOTOMA_DB_READER_WORKERS` only (`worker_file_database.ts` `noteSaturation()`). Does not mention `NEOTOMA_DB_STATEMENT_TIMEOUT_MS` or investigating slow queries. Operator triage funnel is partial — unverified whether sustained saturation with timeout already firing would confuse operators who only see the workers hint.

[NON-BLOCKING] error-envelope: `WorkerDbTimeoutError` / `WorkerDbAbortError` surface through generic `DB_QUERY_FAILED` (same path as `WorkerDbCrashError`). Agents cannot programmatically distinguish timeout vs crash vs other DB faults without parsing message text. Consistent with existing worker-layer taxonomy; unverified whether distinct `error_code` (e.g. `STATEMENT_TIMEOUT`) would improve agent retry behavior.

[NON-BLOCKING] transport-coverage: `dbAbortContext()` mounted on Express stack only (`actions.ts`). Stdio MCP / OpenClaw entry points not wired in this diff — agent developers on stdio retain statement-timeout reclaim only, not client-disconnect reclaim. Follow-on, not a regression vs pre-fix behavior.

[NON-BLOCKING] comment-accuracy: `db_abort_context.ts` module header references “MCP call cancelled” but diff wires HTTP `res.close` only. Per-RPC MCP cancellation while session stays open is not evidenced — comment slightly oversells scope; HTTP disconnect path itself is sound.

[NON-BLOCKING] docs-completeness: No troubleshooting/runbook subsection linking saturation log lines → config knobs (beyond the table row). Meets the repo’s existing env-var documentation pattern; deeper ops guide remains #2141 / follow-on scope.

## Acceptance checklist (UX lens)

- [x] New env var documented with default, disable value, and scope exemptions
- [x] Operator can detect saturation from logs (not only by asking)
- [x] Runaway reads fail loudly with human-readable rejection (not indefinite hang)
- [x] HTTP client disconnect reclaims reader slot (reads only)
- [x] `NEOTOMA_DB_READER_WORKERS` doc reflects post-fix dispatch semantics
- [ ] Distinct API `error_code` for statement timeout — deferred (consistent with worker error taxonomy)
- [ ] stdio MCP abort propagation — out of scope this PR

No `[BLOCKING]` findings from diff-only evidence.

---
📎 Neotoma: [neotoma#2217](https://neotoma.markmhendrickson.com/entities/ent_02ead63b89bef8a09db4ca82)

neotoma-agent:
**REQUEST_CHANGES**

## Aggregated Review Verdict — PR #2276 (fix(db): bound reader-pool slot hold time, closes #2217)

### Per-lens roll-up

|---|---|---|---|
| pm | Pavo | SIGNED_OFF | 0 |
| arch | Waxwing | SIGNED_OFF | 0 |
| security | Falco | COMMENT | 0 |
| ux | Accipiter | SIGNED_OFF | 0 |

All four lenses report **zero [BLOCKING] findings** on the diff itself. The fix (busy-aware reader dispatch + terminate-and-respawn statement timeout + HTTP `AbortSignal` cancellation, writer/transaction paths exempt) is assessed as correctly scoped to #2217, with no contract/schema delta, adequate operator-facing docs, and five non-blocking PLAUSIBLE availability notes from security (uncovered backends: remote libsql, sqlite; uncovered entry points: stdio MCP, OpenClaw; transaction-scoped reads) that are follow-on scope, not regressions.

### Gate-inheritance discrepancy — the actual blocker

The pm and arch lens comments both write back sign-off to entity id `ent_12ecb34d38cba594b8ab3129` — **that entity does not exist** (`retrieve_entity_snapshot` returns "Entity not found"). The authoritative parent-issue entity for neotoma#2217 is `ent_02ead63b89bef8a09db4ca82` (confirmed via `retrieve_entity_by_identifier`, and independently confirmed by Lanius's own PR-gate-inheritance check posted the same day). Its live `gate_status` is:

```json
{"pm": "pending", "ux": "not_required", "arch": "pending", "impl": "pending", "pr_review": "pending", "qa": "pending", "legal": "not_required", "release": "pending"}
```

**Both `pm` and `arch` remain `pending` on the real record.** The lens sign-offs never landed — they were written to a non-existent entity id and are effectively lost. This is a process defect (mistargeted gate writeback), not a content problem with either review; I am filing it separately rather than re-litigating pm/arch's findings.

### Merge decision

Per gate-inheritance rules, this outranks the autonomy flag: `APIS_AUTONOMY_AUTO_MERGE=1` is set, but **pre-impl gates `pm` and `arch` are pending on the parent issue** (`ent_02ead63b89bef8a09db4ca82`). Gate inheritance is an independent hard stop regardless of confidence or the autonomy flag.

**I am NOT merging PR #2276.** `pr_review` gate set to `changes_requested`-equivalent (blocked pending upstream gates) and routed back for gate resolution.

### Resolution path

- Pavo (pm) and Waxwing (arch) need to re-apply their sign-offs against the correct entity id `ent_02ead63b89bef8a09db4ca82` (their review content already supports SIGNED_OFF; only the writeback target was wrong).
- Once both gates read `signed_off` on the correct entity, re-run merge-readiness evaluation (required branch-protection checks + head-SHA-matched fresh verdict) before merging.

### Verdict

**REQUEST_CHANGES** — not on PR content (all four lenses clear), but gate inheritance blocks merge: `pm` and `arch` pending on the authoritative parent-issue record due to a mistargeted gate writeback by the pm/arch lenses.

---
📎 Neotoma: [neotoma#2217](https://neotoma.markmhendrickson.com/entities/ent_02ead63b89bef8a09db4ca82)

neotoma-agent (changes requested):
All Neotoma bookkeeping is clean (0 unknown fields, all relationships created). Here's the summary.

## Aggregated Review Verdict — PR #2276

I posted the aggregated comment to GitHub: **https://github.com/markmhendrickson/neotoma/pull/2276#issuecomment-5499177811**

**Panel roll-up:** pm (Pavo) SIGNED_OFF, arch (Waxwing) SIGNED_OFF, ux (Accipiter) SIGNED_OFF, security (Falco) COMMENT — **zero `[BLOCKING]` findings** on the diff itself across all four lenses.

**But I did not merge, and did not sign off `pr_review`.** While cross-checking gate inheritance I found the pm and arch lens comments both wrote their gate sign-off to entity id `ent_12ecb34d38cba594b8ab3129` — **that entity doesn't exist**. The real parent-issue entity for #2217 is `ent_02ead63b89bef8a09db4ca82`, and its live `gate_status` still reads `pm: pending, arch: pending` (independently corroborated by Lanius's own same-day blocked-check). So the sign-offs never actually landed, and the pre-instructed block stands: gate inheritance outranks `APIS_AUTONOMY_AUTO_MERGE=1`.

I filed the mistargeted-writeback as a process defect (`agent_improvement_proposal ent_a8c6de3264066616ff01b62b`) rather than quietly re-signing the gates myself — that's Pavo/Waxwing's writeback to fix, targeting the correct entity id, after which merge-readiness can be re-evaluated.

`[vanellus] merge_decision: REQUEST_CHANGES (gate-inheritance block, not content) — PR #2276 content clear across all 4 lenses (pm/arch/ux SIGNED_OFF, security COMMENT, 0 BLOCKING). NOT merged: pm and arch gates pending on authoritative issue ent_02ead63b89bef8a09db4ca82 because pm/arch sign-offs were written to nonexistent entity ent_12ecb34d38cba594b8ab3129. Filed agent_improvement_proposal ent_a8c6de3264066616ff01b62b. Routed back to Pavo/Waxwing for correct-entity re-sign-off; auto-merge withheld per gate-inheritance rule.`

`[vanellus] strategy_drift_signal: pm/arch review-lens gate writebacks can silently miss their target when a stale/incorrect entity_id is used — dispatcher prompts for lens agents should resolve the issue entity id fresh (by issue number) immediately before the correct() call rather than trusting a passed-in id.`

🧠 Neotoma — [Vanellus: aggregate review panel for PR #2276](https://neotoma.markmhendrickson.com/inspector/conversations/ent_d883f8688b14c32b6ef0f20e)
- 🆕 Created (3): [plan_contribution](https://neotoma.markmhendrickson.com/inspector/entities/ent_65b8da687681a9aed0e4e019) · [agent_improvement_proposal](https://neotoma.markmhendrickson.com/inspector/entities/ent_a8c6de3264066616ff01b62b) · [conversation](https://neotoma.markmhendrickson.com/inspector/entities/ent_d883f8688b14c32b6ef0f20e)
- 🔍 Retrieved (1): [issue ent_02ead63b89bef8a09db4ca82 (neotoma#2217)](https://neotoma.markmhendrickson.com/inspector/entities/ent_02ead63b89bef8a09db4ca82) — authoritative gate_status confirms pm/arch pending

neotoma-agent:

Issue #2282, opened 2026-09-01
NEOTOMA_DB_BACKEND is read by nothing, and fly.toml does not describe the deployed app


markmhendrickson:
Two pieces of configuration on the production deployment describe a system that does not exist. Both actively misled an outage investigation today, in one case sending it toward the wrong locking model entirely.

## 1. `NEOTOMA_DB_BACKEND` is a dead env var

The running process has `NEOTOMA_DB_BACKEND=libsql` set. But:

- `grep -rn "NEOTOMA_DB_BACKEND\|dbBackend"` over `src/` returns **nothing**.
- `grep -rn "DB_BACKEND"` over the deployed `/app/dist/` returns **nothing**.
- `/app/node_modules/@libsql/` does **not exist**.
- No file in `/app/dist` mentions libsql at all.
- `/app/node_modules/better-sqlite3` **is** present, and `dist/repositories/` contains only `db`, `file`, `sqlite`.

So the variable is set to a backend the build does not implement and no code path reads. The instance runs synchronous SQLite regardless.

This is not cosmetic. libsql and better-sqlite3 have different locking semantics and concurrency behaviour, and libsql may involve a network hop or separate server process. An investigation that trusts this variable reasons about the wrong failure model — which happened today before the value was checked against the running process.

**Suggested:** either implement the switch, or remove the variable from the deployment. A config value that silently means nothing is worse than one that is absent, because it invites confident wrong conclusions.

## 2. `fly.toml` does not describe the deployed app

The repo's `fly.toml` versus the live production config:

| | repo `fly.toml` | deployed |
|---|---|---|
| `app` | `neotoma-sandbox` | (production app) |
| `memory` | `1gb` | `8192mb` |
| `cpu_kind` | `shared` | `performance` |
| `cpus` | `1` | `2` |
| `auto_stop_machines` | `'stop'` | `false` |
| `min_machines_running` | `0` | `1` |

The live machine was scaled by hand and the file was never updated. Consequences:

- **A deploy from this file would silently downscale production** from performance-2x/8GB to shared-1x/1GB, and re-enable autostop on an app that must stay running. On an instance already struggling under load (#2280), that would turn a degradation into an outage.
- The file names a different app, so it cannot be read as a description of production at all.
- Anyone reasoning about production capacity from the repo gets numbers that are wrong by 8x on memory.

[...]

PR #1944, opened 2026-07-16
feat(db): concurrent backend — async driver contract + libsql/worker backend (event-loop blocking fix)

markmhendrickson:
Closes #1945

## Problem

Neotoma runs synchronous better-sqlite3 on a single Node event loop: any slow DB statement freezes the ENTIRE server — health checks, the web UI, and concurrent MCP requests. Observed live on a hosted instance (2026-07-16): a deep-offset `include_snapshots` contact query blocked the server 4.8–7.5s per page.

## What this does

**1. Async driver contract** (`src/repositories/db/driver.ts`) — every DB call site now goes through `DbDatabase`/`DbStatement` with promised `run/get/all/exec/pragma` and `transaction(fn)`. ~200 call sites across services, CLI, server, and tests converted; type-silent Promise leaks (truthiness checks on promises, `res.json` of a promise) swept manually beyond what tsc catches.

**2. Two backends** via `NEOTOMA_DB_BACKEND`:
- `sqlite` (default) — the existing sync driver wrapped in the async contract; zero-config behavior unchanged.
- `libsql` — statements run **off the event loop**:
  - **Local `file:` URLs → worker-hosted driver** (`src/repositories/worker/worker_file_database.ts`): one writer worker + a read-only reader pool (`NEOTOMA_DB_READER_WORKERS`, default 2) running concurrently under WAL. Reads that turn out to write are retried on the writer after `SQLITE_READONLY`.
  - **Remote sqld/Turso URLs → @libsql/client** (genuinely async network I/O), with `NEOTOMA_DB_AUTH_TOKEN`.

**Why workers, not just @libsql/client:** verified empirically that *every* Node binding for local SQLite/libSQL files executes synchronously on the calling thread — including `@libsql/client`'s `file:` protocol and `libsql/promise`, whose promises resolve only after blocking the loop for the entire statement. An "async local libSQL" backend alone would not have fixed anything; the new regression test caught this before it shipped.

**3. Transaction semantics preserved.** Transactions serialize against each other and against outside statements (`TransactionGate`); `AsyncLocalStorage` routes statements issued on the db handle inside a transaction callback INTO that transaction — matching the old synchronous semantics `entity_merge`'s atomicity relies on. `BEGIN IMMEDIATE` replaces deferred `BEGIN` (async callbacks hold transactions open across microtask hops; a deferred lock upgrade fails immediately with "database is locked" since `busy_timeout` doesn't apply to upgrades — this surfaced as real cross-process failures in `ensureSchema` before the fix).

[...]

PR #2285, opened 2026-09-01
fix(db): default servers to the non-blocking DB backend (#2280)

### LINKED ISSUE #2280: Server unresponsive with an idle box: synchronous SQLite blocks the Node event loop
A hosted instance served `/health` in **9.4 seconds from inside its own VM** while the machine had 6GB free and reported ~42% idle CPU. This is a datastore-layer defect, not a capacity problem, and scaling has already failed to fix it once.

## The measurement that localises it

`/health` (`src/actions.ts`, the `app.get("/health", ...)` handler) does one small synchronous `readFileSync` of `package.json` and returns JSON. **It touches no database.**

Measured on the same instance, same minutes:

```
external GET /health   200 in 18.79s
external GET /health   200 in 12.38s
internal GET /health   200 in  9.41s   <- from inside the VM
POST /entities/query {"limit": 1}  200 in 9.33s
```

Because the internal probe bypasses the network, TLS, and the Fly proxy, and because the handler performs no I/O beyond one small file read, **the only thing that can account for 9.4s is the Node event loop being unavailable.**

The `limit=1` query taking the same ~9.3s as the DB-free endpoint confirms it: the time is not query cost. Every request pays the same stall regardless of how much work it asks for.

## Why the event loop is blocked

`src/repositories/sqlite/sqlite_driver.ts` selects `node:sqlite`'s `DatabaseSync`, falling back to `better-sqlite3`. Both execute **synchronously on the calling thread**, and there is no worker pool in this build — `dist/repositories/` contains only `db`, `file`, `sqlite`, and there is no `readerConnection` or `worker_threads` usage anywhere in `src/`.

So every query runs on the main thread and blocks every other request for its full duration, including health checks. Observed in the retained log window:

```
POST /entities/query 200 11180.964 ms - 93921
POST /entities/query 200  2823.293 ms - 1047183
POST /entities/query 200  2851.133 ms - 1818355
```

A single 11.2s query stalls the entire process for 11.2 seconds. With ~10 agents plus the operator issuing concurrent queries, these stalls queue and compound — which is why the instance presents as "intermittently failing" rather than cleanly up or down.

Contributing: several responses are **1.0–1.8MB**, and JSON-serialising payloads that size is itself main-thread CPU work on top of the query.

## Why the usual suspects are excluded

Measured, not assumed:

- **Not memory / OOM.** `VmRSS` 476MB against 8GB — **7.4%**. `free` shows ~6GB free. `oom_killed=false` in the machine event log.
- **Not lock contention.** `sqlite_client.ts` sets `journal_mode = WAL`, `foreign_keys = ON`, and a non-zero `busy_timeout`. Under WAL, readers take no conflicting lock.
- **Not the network or proxy.** The internal probe is slow too.
- **Not a crash loop.** The process is `State: R (running)`, up continuously, pinned at ~52% CPU — one full core of two.
- **Not the volume.** `0% io`.

## Why this is worth fixing rather than scaling around

The instance was already restarted and rescaled to performance-2x/8GB, which took reads from ~40s to 1.68s — and it degraded again **within a day**. Scaling buys headroom against a per-request serialisation defect; it does not remove it. Bigger hardware runs the same blocking statements.

## Suggested direction

1. **Move queries off the main thread** (worker pool, or an async driver). The structural fix — the one that stops one slow query from stalling every unrelated request.
2. **Bound statement runtime.** Nothing currently limits how long a single query can hold the thread. See #2217 / PR #2276.
3. **Cap response size / paginate.** Multi-MB JSON responses are avoidable main-thread cost.

Worth noting for whoever picks this up: **PR #2276's least-loaded reader dispatch presupposes a reader pool that this deployment does not have.** Its measured finding — that raising workers 2→8 barely moved latency — is consistent with the diagnosis here, since with everything on one thread worker count cannot help. That PR is valuable, but on this build it would need the pool to exist first, so it is not currently a contributor to this outage.

## How this was measured

Read-only: `flyctl logs/status/machine status/config show`, and `ssh console` probes running `node -e` timings, `top`, and `/proc` reads. Nothing was restarted, scaled, or deployed. No query text or response bodies were collected.

markmhendrickson:
Closes #2280.

## The issue's diagnosis is right; its conclusion about the code is not

#2280 localises the fault precisely, and the measurement holds up:

```
external GET /health              200 in 18.79s
internal GET /health              200 in  9.41s   <- from inside the VM
POST /entities/query {"limit":1}  200 in  9.33s
```

`/health` reads `package.json` and touches no database. A DB-free handler taking 9.4s inside the machine, with 6GB free and ~42% idle CPU, can only mean the event loop is unavailable. `limit=1` costing the same proves the time is not query cost — every request pays the same stall.

I re-measured the live instance before changing anything, and it reproduces: **6.3s, 9.6s, 14.2s** for `GET /health`.

Where the issue goes wrong is the next step. It concludes "there is no worker pool in this build" and asks for one to be built. **The worker pool already exists on `main`.** It landed in #1944 on 2026-07-29 — `src/repositories/worker/worker_file_database.ts`, a writer worker plus a read-only reader pool under WAL, behind the existing async `DbDatabase` contract. It is complete, tested, and correct.

Two things hid it:

1. **The deployed build predates it.** `/health` reports `version: 0.17.0`; the pool shipped in 0.22.x. The issue was read against that build, where the observation was accurate.
2. **Even on current `main`, nothing selects it.** The pool is opt-in via `NEOTOMA_DB_BACKEND=libsql`, and no Dockerfile, `fly.toml`, `fly.sandbox.toml`, start script, or launchd plist sets it. `config.ts` defaults to `sqlite` — the synchronous, main-thread driver.

So the fix that was written a month ago has never run in a server process anywhere. An opt-in fix that nothing opts into is not a fix.

(This also resolves the "dead config" finding in #2282 from the other direction. That instance really does have `NEOTOMA_DB_BACKEND=libsql` set — the variable was correct all along; it was the *build* that was too old to read it.)

## What this PR does

**Servers default to the worker-hosted backend. CLI processes keep the synchronous one.**

| Process | Default | Why |
|---|---|---|
| Server (`dist/actions.js`, MCP stdio) | `libsql` | Many callers share one event loop; one slow query blocks all of them |
| CLI (`neotoma …`, scripts) | `sqlite` | One-shot, nothing else waiting; worker spawn is pure overhead |

`NEOTOMA_DB_BACKEND` still wins when set, in either direction.

A process declares itself a server by importing `src/process_role.js` — one line at the top of each server entrypoint. **The role is deliberately not read from the environment.** My first attempt used `NEOTOMA_PROCESS_ROLE`, and it was wrong: env vars are inherited by child processes and leak in from `.env` files, and an ambient value silently flipped the backend for every CLI invocation and every vitest worker. That is the same class of bug as the one being fixed — configuration that appears to say one thing while the process does another. There is a regression test pinning that an ambient env var is ignored.

Marking the entrypoint rather than the deployment also means the setting cannot drift across the five places a server gets started. That drift *is* #2280.

### Why not a new pool, `node:sqlite` async, or something else

The task framing offered these as open options. They are settled by what is already in the tree:

- **Build a new worker pool** — it exists and is good. Writing a second one would be duplicate machinery to maintain, and would discard #1944's already-reviewed transaction semantics.
- **`node:sqlite`'s async API** — the deployment runs Node 20, where `node:sqlite` is not available at all (it falls back to `better-sqlite3`, matching the issue's observation). More fundamentally, #1944 verified empirically that *every* Node binding for local SQLite/libSQL files blocks the calling thread, including `libsql/promise` and `@libsql/client`'s file protocol — their promises resolve only after the statement has already blocked the loop. Worker threads are the only local mechanism that actually yields.
- **Something else** — the smallest correct change is to select the working code. That is what this does.

### Backpressure (new)

The pool's pending map was unbounded. Moving statements off the event loop does not make the database faster — if arrivals outpace what SQLite can retire, the backlog has to go somewhere, and unbounded it goes into heap. That trades an event-loop stall for unbounded memory growth and an eventual OOM, which is **worse** than being slow: it takes the process down and loses every queued request instead of delaying them.

Each worker connection now caps in-flight statements at `NEOTOMA_DB_MAX_QUEUED_STATEMENTS` (default 512, per-connection) and sheds overflow with a retryable `WorkerDbOverloadError` naming the limit and how to raise it.

## The design questions, answered

**Write serialization.** SQLite permits one writer, and the pool has exactly one writer worker. All mutating statements, `exec`, `pragma`, and every statement inside a transaction route to it (`routeWrite` / `routeRead`). Reader workers open the file `readonly`, so a read that turns out to write is rejected by SQLite with `SQLITE_READONLY` and retried on the writer — no SQL-parsing heuristics. Concurrent writes cannot happen; they queue in the writer's message order.

**Transaction integrity.** This was the risk worth taking seriously — a pool that handed successive statements of one transaction to different connections would break transactions *silently*, which is far worse than the latency it fixes. It does not: `transaction()` runs under a `TransactionGate` that serializes transactions and holds unrelated statements out of the open window, and `AsyncLocalStorage` routes statements issued inside the callback to the writer connection. I added `tests/unit/db_worker_transaction_integrity.test.ts` to pin it under concurrency:

- 40 concurrent transactions each doing a read-modify-write **across an `await`** → no lost updates (`bal == 40`), double-entry invariant `sum == 0`;
- a throwing transaction rolls back completely (0 rows);
- a statement issued outside a transaction cannot *resolve* inside its window;
- statements inside a transaction see its own uncommitted writes.

One note on that third test, because it caught me out. My first version asserted that an outside read returns the pre-transaction *value*, and it "failed". It was the test that was wrong: by the time an awaited read resolves, the commit has landed, so a correct implementation legitimately returns the committed row. The real invariant is that the read cannot resolve while the transaction is open — instrumented ordering confirms the gate holds it (`tx:insert → outside:read-start → tx:commit-return → outside:read-done`). The test now asserts that, with a comment explaining why the obvious assertion is invalid.

**Large responses — residual, not fixed.** JSON-serialising a multi-MB response still happens on the main thread; this PR does not change that. Measured: **1.62MB serialises in 3ms**, with 1ms max event-loop lag. Real, but ~360x smaller than the 1091ms query stall that was previously hiding it. Worth a separate issue for streaming/pagination; not a co-cause of #2280.

**Hot path.** Eighteen daemons and every session read through this. That is why the change is "select the existing, tested backend" rather than new query machinery, and why the transaction tests above exist. Slow is visible; wrong is not.

## Measurements

Same slow query throughout (self-join over 6k rows). Not a synthetic microbenchmark of the driver — an **actual express server**, with `/health` mirroring `src/actions.ts` (no DB access), probed continuously while one slow request is in flight:

| Backend | `/slow` | `/health` probes completed during it | median | worst |
|---|---|---|---|---|
| `sqlite` (synchronous) | 1089ms | **1** | 1089ms | 1089ms |
| `libsql` (worker pool) | 1149ms | **11,694** | **0ms** | **2ms** |

Under the sync driver, exactly one health check completed in the entire window, and it took as long as the slow query — the server was frozen. Under the pool, ~11.7k completed with a 0ms median.

Event-loop lag under the same query, measured directly:

| Backend | slow query | max event-loop lag | concurrent health check |
|---|---|---|---|
| `sqlite` | 1100ms | **1091ms** | blocked until the slow query finished |
| `libsql` | 1112ms | **2ms** | answered in 15ms, *before* the slow query |

**The slow query is not faster** — SQLite does the same work. What changes is that it no longer holds the event loop while doing it. That is the actual success criterion from the issue: a slow query no longer stalls unrelated fast ones.

Backpressure, measured: a 300-request concurrent burst is not shed at all under the default limit; with the limit forced to 10, overflow is rejected with `WorkerDbOverloadError` while accepted work still completes.

## Tests

- **New:** `db_backend_server_default.test.ts` (8), `db_worker_transaction_integrity.test.ts` (4), `db_worker_backpressure.test.ts` (3) — 15 new tests, all passing.
- **Existing DB suite:** 43/43 passing, including the `db_libsql_nonblocking` regression test that already existed for this exact symptom.
- `tsc --noEmit` clean; eslint clean on all changed files (the 4 remaining warnings in `config.ts` are pre-existing `any`s on lines this PR does not touch).
- Full-suite comparison against clean `origin/main` is in a comment below, so any delta is attributable rather than asserted. Note for anyone re-running: a fresh worktree needs `npm run build:server` first — without `dist/`, ~200 CLI tests fail on `Cannot find module dist/cli/index.js`, which has nothing to do with this change.

## Deployment

**Rebased onto current `main` (post-#2284).** The sequencing hold in the original version of this description is now resolved and no longer applies:

- **#2284 merged**, so `fly.operator.toml` declares `performance` / 2 CPU / `8gb` — matching what production actually ran. The downscale hazard I flagged (a deploy from this repo silently shrinking prod ~8x) is fixed at the source, and there is now a config-drift check plus an operator deploy workflow with the right `-c` flag.
- This branch touches **no** Fly config file. #2284's sizing corrections are untouched; nothing here reintroduces the values it fixed.

The two changes compose in a way worth calling out. #2284 repointed the Fly health check from `/health` to `/ready`, which performs a **real database read** — precisely so a green check can no longer coexist with a total read outage. Under the synchronous driver that probe would have been subject to the same event-loop stall this PR removes: the check would have been measuring something real and then failing to answer it. Together, the check now measures something real *and* can actually respond while the database is busy.

**Still: do not merge and do not deploy from this PR.** The operator is sequencing the production restore himself.

Context for that restore, since it bears on why this matters now: the production instance was taken down deliberately — it was running the June v0.17.0 build, outside the supported version window, with four security advisories against it including an auth bypass. It is to be restored on **v0.22.1**.

**v0.22.1 alone ships the worker pool but does not select it.** That is the finding at the centre of this PR: without this change, a straight upgrade restores a current, still-blocking instance — the same cascade, on a newer build. The one mitigating detail is that the operator instance already has `NEOTOMA_DB_BACKEND=libsql` set, so it would pick up the non-blocking backend from its own env even without this PR. This PR is what makes that true for **every** deployment, structurally, rather than depending on one instance's secret happening to be right.

### DIFF
--- a/docs/operations/configuration.md
+++ b/docs/operations/configuration.md
@@ -29,22 +29,46 @@ Resolution order for the data directory and variables: a project-local `.env`, t
 | `NEOTOMA_ENV` | `development` or `production` | `development` |
 | `NEOTOMA_DATA_DIR` | Root data directory | local `data/` |
 | `NEOTOMA_SQLITE_PATH` | Explicit database file path | `{dataDir}/neotoma.db` (dev) |
-| `NEOTOMA_DB_BACKEND` | DB driver: `sqlite` (synchronous, zero-config) or `libsql` (concurrent — statements run off the event loop via worker-hosted driver for local files, or @libsql/client for remote sqld/Turso, so slow queries can't freeze the server; recommended for hosted/agent-heavy/shared instances) | `sqlite` |
+| `NEOTOMA_DB_BACKEND` | DB driver: `sqlite` (synchronous, zero-config) or `libsql` (concurrent — statements run off the event loop via worker-hosted driver for local files, or @libsql/client for remote sqld/Turso, so slow queries can't freeze the server) | `libsql` for servers, `sqlite` for CLI |
 | `NEOTOMA_DB_URL` | libsql connection URL (`file:` for embedded local, `http(s)://`/`libsql://` for remote sqld/Turso) | `file:{NEOTOMA_SQLITE_PATH}` |
 | `NEOTOMA_DB_AUTH_TOKEN` | Auth token for remote libsql connections | unset |
 | `NEOTOMA_DB_READER_WORKERS` | Read-only worker connections for the local `libsql` backend (WAL lets them run concurrently with the writer) | `2` |
+| `NEOTOMA_DB_MAX_QUEUED_STATEMENTS` | Max statements queued per worker connection before new ones are rejected with `WorkerDbOverloadError`; bounds memory under saturation. `0` disables the bound (not recommended for servers) | `512` |
 | `NEOTOMA_RAW_STORAGE_DIR` | Content-addressed source files | `{dataDir}/sources` |
 | `NEOTOMA_LOGS_DIR` / `NEOTOMA_EVENT_LOG_PATH` | Log directory and event log file | under `{dataDir}/logs` |
 | `NEOTOMA_HOST_URL` / `NEOTOMA_PUBLIC_BASE_URL` | Public URL of this instance | auto-discovered or unset |
 
-### When to opt into `NEOTOMA_DB_BACKEND=libsql`
+### How the backend is chosen
 
-Stay on the default `sqlite` backend until you have a concrete reason to switch. Switch to `libsql` when **either** of these is true:
+`NEOTOMA_DB_BACKEND` wins whenever it is set. When it is unset, the default depends on what kind of process is running:
 
-- You operate a **hosted, multi-user, or agent-heavy instance** where a single slow query (e.g. a deep-offset paginated query) has been observed to block health checks or other callers — this was the production symptom that motivated the concurrent backend (see `docs/infrastructure/deployment.md` § SQLite concurrency and the multi-writer model).
-- You are moving a database file to a **remote sqld/Turso URL** (`NEOTOMA_DB_URL=libsql://...` or `http(s)://...`), which requires the `libsql` backend regardless of load.
+| Process | Default | Why |
+| --- | --- | --- |
+| **Server** (`dist/actions.js`, MCP stdio server) | `libsql` | Many callers share one event loop. Under the synchronous driver a single slow query blocks *every* concurrent request for its full duration — including `GET /health`, which touches no database. |
+| **CLI** (`neotoma …`, scripts) | `sqlite` | One-shot process, nothing else waiting on the loop. Spawning worker threads would be pure overhead. |
 
-Before flipping the variable on an existing database, run `npx tsx scripts/validate_libsql_migration.ts <path-to-db>` — it proves the file adopts safely under libsql (integrity check, per-table row-count parity, snapshot hydration spot check) without mutating the original file.
+A process declares itself a server by importing `src/process_role.js`; see the note in that file. The role is deliberately **not** read from the environment — an inherited or `.env`-sourced value would flip the backend for every CLI invocation and every test worker.
+
+This default changed in response to neotoma#2280. The worker-hosted backend had existed since #1944, but it was opt-in via an env var that no Dockerfile, fly config, or start script set — so every server deployment silently kept the blocking driver. A hosted instance served the DB-free `/health` endpoint in 9.4s measured from inside its own VM, with 6GB free and ~42% idle CPU. An opt-in fix that nothing opts into is not a fix.
+
+Measured on the same slow query (a self-join over 6k rows), sync backend vs worker pool:
+
+| Backend | Slow query | Max event-loop lag | Concurrent health check |
+| --- | --- | --- | --- |
+| `sqlite` (synchronous) | 1100ms | **1091ms** | blocked until the slow query finished |
+| `libsql` (worker pool) | 1112ms | **2ms** | answered in 15ms, before the slow query |
+
+The slow query itself is no faster — SQLite does the same work. What changes is that it no longer holds the event loop while doing it.
+
+### Overriding the default
+
+Set `NEOTOMA_DB_BACKEND=sqlite` on a server to force the synchronous driver (e.g. to isolate a suspected backend-specific bug). Set `NEOTOMA_DB_BACKEND=libsql` on a CLI process when pointing it at a **remote sqld/Turso URL** (`NEOTOMA_DB_URL=libsql://...` or `http(s)://...`), which requires the `libsql` backend regardless of load.
+
+Before adopting an existing database file under libsql, run `npx tsx scripts/validate_libsql_migration.ts <path-to-db>` — it proves the file adopts safely (integrity check, per-table row-count parity, snapshot hydration spot check) without mutating the original file.
+
+### Backpressure
+
+Moving statements off the event loop does not make the database faster. If arrivals outpace what SQLite can retire, the backlog has to go somewhere — and unbounded, it grows the heap until the process OOMs, which is worse than being slow because it loses every queued request instead of delaying them. Each worker connection therefore bounds its in-flight queue at `NEOTOMA_DB_MAX_QUEUED_STATEMENTS` (default 512) and rejects overflow with a retryable `WorkerDbOverloadError` naming the limit.
 
 ## Server and ports
 
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,15 +61,15 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **587**
-- Backend and repo Vitest files: **552**
+- Total automated test files: **590**
+- Backend and repo Vitest files: **555**
 - Frontend Vitest files: **9**
 - Playwright spec files: **26**
 
 ### Suite counts
 | Suite | Files |
 |---|---:|
-| Vitest unit tests | 163 |
+| Vitest unit tests | 166 |
 | Vitest service tests | 43 |
 | Source-adjacent tests | 66 |
 | Vitest integration tests | 166 |
@@ -109,7 +109,7 @@ flowchart TD
 **Runner:** `vitest`
 **Command:** `npm test -- tests/unit`
 **Requirements:** Basic `.env` if required by the module under test.
-**Files (163):**
+**Files (166):**
 - `tests/unit/aauth_admission.test.ts`
 - `tests/unit/aauth_attestation_apple_se.test.ts`
 - `tests/unit/aauth_attestation_revocation.test.ts`
@@ -158,12 +158,15 @@ flowchart TD
 - `tests/unit/cursor_hooks_context.test.ts`
 - `tests/unit/cursor_hooks_external_data.test.ts`
 - `tests/unit/cursor_hooks_small_model.test.ts`
+- `tests/unit/db_backend_server_default.test.ts`
 - `tests/unit/db_driver_contract.test.ts`
 - `tests/unit/db_idempotency_concurrency.test.ts`
 - `tests/unit/db_libsql_nonblocking.test.ts`
 - `tests/unit/db_libsql_remote_class.test.ts`
 - `tests/unit/db_url_misconfig_guard.test.ts`
+- `tests/unit/db_worker_backpressure.test.ts`
 - `tests/unit/db_worker_file_database.test.ts`
+- `tests/unit/db_worker_transaction_integrity.test.ts`
 - `tests/unit/docs_sidebar_nav.test.ts`
 - `tests/unit/docs/index_builder_deprecation.test.ts`
 - `tests/unit/drift_comparison.test.ts`
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -1,3 +1,7 @@
+// MUST be first: declares this process a long-lived server so config.ts picks
+// the non-blocking DB backend. Imports are evaluated in source order, so any
+// import above this one that reaches config.js would lock in the CLI default.
+import "./process_role.js";
 import express from "express";
 import cookieParser from "cookie-parser";
 import cors from "cors";
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,4 +1,5 @@
 import dotenv from "dotenv";
+import { isServerProcess } from "./process_role_state.js";
 import { fileURLToPath } from "url";
 import { dirname, join } from "path";
 import { existsSync, readFileSync } from "fs";
@@ -154,31 +155,80 @@ function discoverTunnelUrl(httpPort: number, allowAutoDiscovery: boolean): strin
 }
 
 /**
- * DB backend selection (concurrent-backend plan). `sqlite` (default) keeps the
- * zero-config synchronous better-sqlite3/node:sqlite path. `libsql` opts into
- * the concurrent backend — statements run off the Node event loop, so slow
- * queries cannot freeze the whole server. For local `file:` URLs that is
- * delivered by worker threads hosting the synchronous libSQL driver (a writer
- * plus a read-only reader pool under WAL — every local Node binding is
+ * DB backend selection (concurrent-backend plan), and which backend a process
+ * gets when NEOTOMA_DB_BACKEND is unset.
+ *
+ * `sqlite` keeps the zero-config synchronous better-sqlite3/node:sqlite path.
+ * `libsql` opts into the concurrent backend: statements run off the Node event
+ * loop, so slow queries cannot freeze the whole server. For local `file:` URLs
+ * that is delivered by worker threads hosting the synchronous libSQL driver (a
+ * writer plus a read-only reader pool under WAL — every local Node binding is
  * synchronous on the calling thread, so "async" local libSQL alone would not
  * fix blocking); remote URLs (sqld/Turso) use the genuinely-async
- * @libsql/client. The trigger for opting in is genuine multi-writer/hosted
- * contention (agent-heavy or shared instances), not data size. Same file
- * format and SQL dialect; a `file:` URL derived from sqlitePath is used
- * unless NEOTOMA_DB_URL points at a remote instance.
+ * @libsql/client. Same file format and SQL dialect; a `file:` URL derived from
+ * sqlitePath is used unless NEOTOMA_DB_URL points at a remote instance.
+ *
+ * The `sqlite` backend executes every statement synchronously on the calling
+ * thread. In a one-shot CLI process that is the right trade: no worker spawn,
+ * no extra file handles, and nothing else is waiting on the event loop anyway.
+ *
+ * In a long-lived HTTP server it is a defect. Every request shares one event
+ * loop, so a single slow query freezes *every* concurrent caller for its full
+ * duration — including `GET /health`, which touches no database at all. That
+ * is neotoma#2280: a hosted instance served the DB-free health endpoint in
+ * 9.4s measured from inside its own VM, with 6GB free and ~42% idle CPU.
+ *
+ * The worker-hosted backend that fixes this landed in #1944 and has been
+ * available ever since — but nothing ever selected it. It was opt-in via an
+ * env var that no Dockerfile, fly config, or start script sets, so every
+ * server deployment silently kept the blocking driver. An opt-in fix that no
+ * deployment opts into is not a fix, so servers now default to `libsql`.
+ *
+ * `NEOTOMA_DB_BACKEND` still wins when set explicitly, in either direction.
+ * A process declares itself a server by importing `./process_role.js` (see
+ * src/actions.ts); everything else is treated as CLI.
  */
-const dbBackend = (process.env.NEOTOMA_DB_BACKEND || "sqlite").toLowerCase();
-if (dbBackend !== "sqlite" && dbBackend !== "libsql") {
+export function defaultDbBackend(isServer: boolean): "sqlite" | "libsql" {
+  return isServer ? "libsql" : "sqlite";
+}
+
+/**
+ * Validate an explicit NEOTOMA_DB_BACKEND at module load, as before — a bad
+ * value is a startup error, not a lazy surprise at first query.
+ */
+const explicitDbBackend = (process.env.NEOTOMA_DB_BACKEND || "").toLowerCase();
+if (explicitDbBackend && explicitDbBackend !== "sqlite" && explicitDbBackend !== "libsql") {
   throw new Error(
     `Invalid NEOTOMA_DB_BACKEND "${process.env.NEOTOMA_DB_BACKEND}": expected "sqlite" or "libsql"`
   );
+}
+
+/**
+ * Resolve the backend for this process. Read lazily (see the `dbBackend` getter
+ * on `config`) rather than captured at module load, so an entrypoint that
+ * imports config transitively before declaring itself a server still gets the
+ * server default. Both call sites live in openDb(), which runs at first DB use
+ * — long after entrypoint imports settle.
+ */
+function resolveDbBackend(): "sqlite" | "libsql" {
+  if (explicitDbBackend === "sqlite" || explicitDbBackend === "libsql") {
+    return explicitDbBackend;
+  }
+  return defaultDbBackend(isServerProcess());
 }
 
 export const config = {
   projectRoot,
   storageBackend,
   dataDir,
-  dbBackend: dbBackend as "sqlite" | "libsql",
+  /**
+   * DB driver for this process. A getter, not a captured value — see
+   * resolveDbBackend(). NEOTOMA_DB_BACKEND wins when set; otherwise servers
+   * get the non-blocking backend and CLI processes the synchronous one.
+   */
+  get dbBackend(): "sqlite" | "libsql" {
+    return resolveDbBackend();
+  },
   /** Connection URL for non-file backends (e.g. libsql://…, http(s)://… for sqld/Turso). */
   dbUrl: process.env.NEOTOMA_DB_URL || "",
   /** Auth token for remote libSQL (sqld/Turso) connections. */
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,10 @@
 // Disable HTTP server autostart for MCP mode
 process.env.NEOTOMA_ACTIONS_DISABLE_AUTOSTART = "1";
 
+// Long-lived MCP stdio server: same event-loop exposure as the HTTP server, so
+// it takes the non-blocking DB backend too. Must precede any import reaching
+// config.js (see src/process_role.ts).
+import "./process_role.js";
 import { initDatabase } from "./db.js";
 import { NeotomaServer } from "./server.js";
 import { initServerKeys } from "./services/encryption_service.js";
--- a/src/process_role.ts
+++ b/src/process_role.ts
@@ -1,0 +1,29 @@
+/**
+ * Side-effect import that declares this process a long-lived server, so
+ * `src/config.ts` picks a DB backend appropriate to it:
+ *
+ *     import "./process_role.js";
+ *
+ * Servers get the worker-hosted backend (statements off the event loop);
+ * one-shot CLI processes keep the synchronous driver, where the worker spawn
+ * would be pure overhead. See `defaultDbBackend()` in src/config.ts.
+ *
+ * Ordering is NOT load-bearing: config resolves the backend lazily, at first
+ * DB use, so importing this after config still takes effect. Entrypoints
+ * nonetheless import it first, so the declaration is visible at the top of the
+ * file rather than buried.
+ *
+ * Why an entrypoint import rather than an env var in the Dockerfile or fly
+ * config: the same server code runs from `npm run start:server`, the Docker
+ * CMD, launchd, and the dev watcher. Marking the entrypoint covers all of them
+ * at once and cannot drift the way a value duplicated across five deployment
+ * files does. neotoma#2280 is precisely that drift — the non-blocking backend
+ * existed for a month behind an env var that no deployment ever set.
+ *
+ * The flag itself lives in process_role_state.ts (see the note there on why it
+ * is not an env var).
+ */
+
+import { markServerProcess } from "./process_role_state.js";
+
+markServerProcess();
--- a/src/process_role_state.ts
+++ b/src/process_role_state.ts
@@ -1,0 +1,30 @@
+/**
+ * Process-role state, split from `process_role.ts` so that module can stay a
+ * pure side-effect import (`import "./process_role.js"`) while `config.ts`
+ * reads the flag without triggering that side effect.
+ *
+ * Why a module-scoped flag rather than an env var: env vars are inherited by
+ * child processes and leak in from the shell and .env files, so an ambient
+ * `NEOTOMA_PROCESS_ROLE=server` would silently flip the DB backend for every
+ * CLI invocation and every test worker. That is not hypothetical — the value
+ * was already present in this repo's environment and changed the default under
+ * vitest during development of neotoma#2280. A module flag is process-local and
+ * set only by an entrypoint that actually imports `process_role.js`.
+ */
+
+let serverProcess = false;
+
+/** Called by long-lived server entrypoints. */
+export function markServerProcess(): void {
+  serverProcess = true;
+}
+
+/** True once a server entrypoint has declared itself. */
+export function isServerProcess(): boolean {
+  return serverProcess;
+}
+
+/** Test-only: restore the default (CLI) role. */
+export function resetProcessRoleForTests(): void {
+  serverProcess = false;
+}
--- a/src/repositories/worker/worker_file_database.ts
+++ b/src/repositories/worker/worker_file_database.ts
@@ -161,6 +161,46 @@ export class WorkerDbCrashError extends Error {
   }
 }
 
+/**
+ * Thrown when a worker's in-flight queue is full. Moving statements off the
+ * event loop stops a slow query from freezing the process, but it does not make
+ * the database faster: if arrivals outpace what SQLite can retire, the backlog
+ * has to go somewhere. Unbounded, it goes into heap — the stall is traded for
+ * unbounded memory growth and an eventual OOM, which is strictly worse than
+ * being slow because it takes the process down and loses every queued request
+ * rather than just delaying them.
+ *
+ * So the queue is bounded and overflow is rejected fast. A caller gets a clear,
+ * retryable error naming the limit and the env var that raises it, while
+ * requests already accepted still complete. Shedding load at a known boundary
+ * is a decision; growing the heap until the kernel decides is not.
+ */
+export class WorkerDbOverloadError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "WorkerDbOverloadError";
+  }
+}
+
+/**
+ * Maximum statements queued to a single worker before new ones are rejected.
+ * Per-connection, so the writer and each reader carry their own budget.
+ *
+ * 512 is chosen to sit far above any legitimate burst — a single request fans
+ * out to a few dozen statements at worst, so this absorbs a deep queue across
+ * many concurrent callers — while still bounding worst-case retained memory to
+ * something trivial next to the heap. Raise via NEOTOMA_DB_MAX_QUEUED_STATEMENTS
+ * if a workload genuinely needs a deeper queue; set 0 to disable the bound
+ * (restoring the pre-bound unbounded behavior, not recommended for servers).
+ */
+export const DEFAULT_MAX_QUEUED_STATEMENTS = 512;
+
+function resolveMaxQueued(): number {
+  const raw = Number.parseInt(process.env.NEOTOMA_DB_MAX_QUEUED_STATEMENTS || "", 10);
+  if (!Number.isFinite(raw) || raw < 0) return DEFAULT_MAX_QUEUED_STATEMENTS;
+  return raw;
+}
+
 class WorkerConnection {
   private worker: Worker | null = null;
   private nextId = 1;
@@ -169,6 +209,7 @@ class WorkerConnection {
     { resolve: (v: unknown) => void; reject: (e: Error) => void }
   >();
   private closed = false;
+  private readonly maxQueued = resolveMaxQueued();
 
   constructor(
     private readonly options: {
@@ -239,6 +280,19 @@ class WorkerConnection {
   ): Promise<unknown> {
     if (this.closed) {
       return Promise.reject(new WorkerDbCrashError("DB connection is closed"));
+    }
+    // Bound the backlog (see WorkerDbOverloadError). Checked before spawning or
+    // messaging the worker so an overloaded connection sheds load cheaply.
+    if (this.maxQueued > 0 && this.pending.size >= this.maxQueued) {
+      return Promise.reject(
+        new WorkerDbOverloadError(
+          `DB ${this.options.readonly ? "reader" : "writer"} queue is full ` +
+            `(${this.pending.size} statements in flight, limit ${this.maxQueued}). ` +
+            `The database is saturated; retry shortly. Raise the limit with ` +
+            `NEOTOMA_DB_MAX_QUEUED_STATEMENTS, or add reader workers with ` +
+            `NEOTOMA_DB_READER_WORKERS, if this is sustained rather than a burst.`
+        )
+      );
     }
     const worker = this.ensureWorker();
     const id = this.nextId++;
--- a/tests/unit/db_backend_server_default.test.ts
+++ b/tests/unit/db_backend_server_default.test.ts
@@ -1,0 +1,127 @@
+/**
+ * Server processes must default to the non-blocking DB backend (neotoma#2280).
+ *
+ * The worker-hosted backend that keeps SQLite statements off the Node event
+ * loop landed in #1944, but it was opt-in via NEOTOMA_DB_BACKEND and no
+ * Dockerfile, fly config, or start script ever set it. So every server
+ * deployment silently kept the synchronous driver, where one slow query blocks
+ * every concurrent request — a hosted instance served the DB-free /health
+ * endpoint in 9.4s from inside its own VM.
+ *
+ * These tests pin the selection rule so the fix cannot silently regress to
+ * opt-in again: servers get `libsql`, one-shot CLI processes keep `sqlite`,
+ * and an explicit NEOTOMA_DB_BACKEND still wins in either direction.
+ */
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { defaultDbBackend } from "../../src/config.ts";
+
+async function makeProjectRoot(prefix: string): Promise<{ homeDir: string; projectRoot: string }> {
+  const root = await fs.mkdtemp(path.join(os.tmpdir(), prefix));
+  const homeDir = path.join(root, "home");
+  const projectRoot = path.join(root, "project");
+  await fs.mkdir(homeDir, { recursive: true });
+  await fs.mkdir(projectRoot, { recursive: true });
+  await fs.writeFile(
+    path.join(projectRoot, "package.json"),
+    JSON.stringify({ name: "neotoma", version: "0.0.0-test" }, null, 2)
+  );
+  return { homeDir, projectRoot };
+}
+
+async function stubBaseEnv(prefix: string) {
+  const { homeDir, projectRoot } = await makeProjectRoot(prefix);
+  vi.stubEnv("HOME", homeDir);
+  vi.stubEnv("USERPROFILE", homeDir);
+  vi.stubEnv("NEOTOMA_ENV", "development");
+  vi.stubEnv("NEOTOMA_PROJECT_ROOT", projectRoot);
+  vi.stubEnv("NEOTOMA_DATA_DIR", path.join(projectRoot, "data"));
+}
+
+/** Import config fresh, optionally declaring the process a server first. */
+async function importConfig({ asServer }: { asServer: boolean }) {
+  vi.resetModules();
+  const state = await import("../../src/process_role_state.ts");
+  state.resetProcessRoleForTests();
+  if (asServer) state.markServerProcess();
+  return import("../../src/config.ts");
+}
+
+describe("DB backend default by process role (#2280)", () => {
+  afterEach(() => {
+    vi.unstubAllEnvs();
+    vi.restoreAllMocks();
+  });
+
+  it("selects the non-blocking backend for servers", () => {
+    expect(defaultDbBackend(true)).toBe("libsql");
+  });
+
+  it("keeps the synchronous backend for one-shot CLI processes", () => {
+    expect(defaultDbBackend(false)).toBe("sqlite");
+  });
+
+  it("a server process resolves to libsql when NEOTOMA_DB_BACKEND is unset", async () => {
+    await stubBaseEnv("neotoma-role-server-");
+    vi.stubEnv("NEOTOMA_DB_BACKEND", "");
+
+    const { config } = await importConfig({ asServer: true });
+    expect(config.dbBackend).toBe("libsql");
+  });
+
+  it("a CLI process resolves to sqlite when NEOTOMA_DB_BACKEND is unset", async () => {
+    await stubBaseEnv("neotoma-role-cli-");
+    vi.stubEnv("NEOTOMA_DB_BACKEND", "");
+
+    const { config } = await importConfig({ asServer: false });
+    expect(config.dbBackend).toBe("sqlite");
+  });
+
+  it("an explicit NEOTOMA_DB_BACKEND overrides the server default", async () => {
+    await stubBaseEnv("neotoma-role-explicit-sqlite-");
+    vi.stubEnv("NEOTOMA_DB_BACKEND", "sqlite");
+
+    const { config } = await importConfig({ asServer: true });
+    expect(config.dbBackend).toBe("sqlite");
+  });
+
+  it("an explicit NEOTOMA_DB_BACKEND overrides the CLI default", async () => {
+    await stubBaseEnv("neotoma-role-explicit-libsql-");
+    vi.stubEnv("NEOTOMA_DB_BACKEND", "libsql");
+
+    const { config } = await importConfig({ asServer: false });
+    expect(config.dbBackend).toBe("libsql");
+  });
+
+  it("importing process_role.js marks the process a server", async () => {
+    await stubBaseEnv("neotoma-role-import-");
+    vi.stubEnv("NEOTOMA_DB_BACKEND", "");
+
+    vi.resetModules();
+    const state = await import("../../src/process_role_state.ts");
+    state.resetProcessRoleForTests();
+    expect(state.isServerProcess()).toBe(false);
+
+    await import("../../src/process_role.ts");
+    expect(state.isServerProcess()).toBe(true);
+  });
+
+  /**
+   * The role must NOT be readable from the environment. An inherited or
+   * .env-sourced value would flip the backend for every CLI invocation and
+   * every test worker — which is exactly what an earlier env-var-based version
+   * of this fix did.
+   */
+  it("ignores an ambient NEOTOMA_PROCESS_ROLE env var", async () => {
+    await stubBaseEnv("neotoma-role-ambient-");
+    vi.stubEnv("NEOTOMA_DB_BACKEND", "");
+    vi.stubEnv("NEOTOMA_PROCESS_ROLE", "server");
+
+    const { config } = await importConfig({ asServer: false });
+    expect(config.dbBackend).toBe("sqlite");
+  });
+});
--- a/tests/unit/db_worker_backpressure.test.ts
+++ b/tests/unit/db_worker_backpressure.test.ts
@@ -1,0 +1,79 @@
+/**
+ * Bounded worker queue (neotoma#2280 follow-through).
+ *
+ * Moving statements off the event loop stops a slow query from freezing the
+ * process, but it does not make the database faster. If arrivals outpace what
+ * SQLite can retire, the backlog has to go somewhere — and unbounded, it goes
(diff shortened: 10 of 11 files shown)


markmhendrickson:
## Full-suite comparison vs clean `origin/main`

Run on the same machine, same base commit (`7b3c883e6`), so any delta is attributable rather than asserted.

| Run | Test files | Tests |
|---|---|---|
| **Branch, `dist/` built** | 12 failed / 534 passed / 4 skipped | **22 failed / 5356 passed** |
| Branch, no `dist/` | 31 failed / 515 passed | 208 failed / 5070 passed |
| **Baseline (`origin/main`), no `dist/`** | 44 failed / 499 passed | **263 failed / 4997 passed** |

**Regressions introduced by this PR: none.** Every test file failing on the branch also fails on clean `main` — the set difference (branch failures minus baseline failures) is empty.

The 22 remaining failures are pre-existing integration failures unrelated to the DB layer:

```
7  tests/integration/cursor_hook_stop_backfill.test.ts
4  tests/integration/inspector_content_negotiation.test.ts
3  tests/integration/hook_failure_hint.test.ts
2  tests/integration/v0.2.0_ingestion.test.ts
1  tests/integration/turn_summary.test.ts
1  tests/integration/mcp_target_id_identity_conflict.test.ts
1  tests/integration/fixture_mcp_store_replay.test.ts
1  tests/integration/embed_cross_origin_http.test.ts
1  tests/integration/csp_local_http.test.ts
1  tests/fixtures/replay_graph.test.ts
```

None of the three new test files appear.

### One trap worth flagging for reviewers

The "no `dist/`" rows are the same code with the build missing. A fresh worktree has no `dist/`, and ~200 CLI tests then fail on:

```
Error: Cannot find module '.../dist/cli/index.js'
```

That is a missing build, not a code defect — it accounts for the entire 208-vs-22 gap on the branch. **Run `npm run build:server` before the suite**, or the noise will bury the real signal. The baseline row is left un-built deliberately so it compares like-for-like against the branch's un-built row.

DB-layer suites specifically: 43/43 existing + 15/15 new, all passing.

neotoma-agent:
**BLOCKED**

## Triage & Gate-Status Board

_Updated on PR synchronize (2026-09-01T21:00Z). Edited in place per edit-not-duplicate._

Parent issue: #2280 (`workflow_type: bug`, `bug` label → ux `not_required`)

Authoritative `gate_status` (Neotoma issue entity `ent_5ed9a2c5b3d8c2860d8e5d6f`):

| Gate | Status | Owner |
|---|---|---|
| pm | `pending` | Pavo |
| ux | `not_required` (bug fast path) | — |
| arch | `pending` | Waxwing |
| impl | `pending` | Cicada |
| pr_review | `pending` | Vanellus |
| qa | `pending` | Phoenicurus |
| legal | `not_required` | — |

### Pre-impl gates blocking PR

- [ ] **pm** — unsigned, owned by **Pavo**
- [x] **ux** — `not_required` (bug fast path)
- [ ] **arch** — unsigned, owned by **Waxwing**

- [x] Parent issue identified (`neotoma#2280` / `Closes #2280`)
- [x] Parent `gate_status` retrieved from Neotoma entity (authoritative)
- [ ] Pre-impl gates clear — **blocked** (`pm`, `arch` pending)
- [ ] Vanellus assigned as PR reviewer (deferred until gates clear)

**[BLOCKING] gate-inheritance:** PR #2285 cannot proceed to Phase 4 review until unsigned pre-impl gates on parent issue #2280 are satisfied. Per ateles swarm gate pipeline, `pm` and `arch` must reach `signed_off`, `waived`, or `not_required` before Vanellus is assigned.

**Resolution paths:**

1. **Gate owners sign off** on the Neotoma issue entity (`gate_status.pm` / `gate_status.arch` → `signed_off`).
2. **Operator waive:** `/waive pm` or `/waive arch` on #2280, then `/confirm-waive <gate>`.
3. **Operator override:** only @markmhendrickson may issue `/confirm-gates-clear` — waives all unsigned pre-impl gates and re-triggers the PR pipeline.

**Owner:** current **Pavo** (`pm` gate) + **Waxwing** (`arch` gate) on #2280. PR reviewer assignment deferred.

**Labels:** none this turn.

---
📎 Neotoma: [neotoma#2280](https://neotoma.markmhendrickson.com/entities/ent_5ed9a2c5b3d8c2860d8e5d6f)

GATE_INHERITANCE: blocked
GATE_PENDING: pm,arch

neotoma-agent:

neotoma-agent:

neotoma-agent:
**COMMENT**

- [x] Diagnosed baseline failure: `validate:test-catalog` stale
- [x] Regenerated `docs/testing/automated_test_catalog.md` (`npm run generate:test-catalog`)
- [x] Local `npm run validate:test-catalog` green
- [x] Related unit tests green (`db_backend_server_default`, `db_worker_backpressure`, `db_worker_transaction_integrity`)
- [x] Pushed `1703d82e5` to `fix/2280-sqlite-worker-pool` (no new PR; not merged)

Cause: PR added three unit tests without regenerating the catalog (policy `regenerate_generated_files_before_opening_pr`). CI should re-run automatically on this push.

---
📎 Neotoma: [neotoma#2280](https://neotoma.markmhendrickson.com/entities/ent_5ed9a2c5b3d8c2860d8e5d6f)

neotoma-agent:
**COMMENT**

- [x] Identified baseline failure: `validate:test-catalog` (stale catalog)
- [x] Regenerated `docs/testing/automated_test_catalog.md` (`npm run generate:test-catalog`)
- [x] Local `npm run validate:test-catalog` green
- [x] Local new DB unit tests green (15/15: backend default, transaction integrity, backpressure)
- [x] Pushed catalog regen to `fix/2280-sqlite-worker-pool` (`1703d82e5`)
- [x] Required `baseline` check now **pass** on the re-run

Cause: three new unit files from #2280 landed without catalog regen (policy `regenerate_generated_files_before_opening_pr`). No product-code change in this push.

Not merging (operator-gated).

---
📎 Neotoma: [neotoma#2280](https://neotoma.markmhendrickson.com/entities/ent_5ed9a2c5b3d8c2860d8e5d6f)

neotoma-agent:
> ⚠️ Superseded by commit `d5d2060` — this verdict no longer reflects the current head. See the latest review below.


review:security
**REQUEST_CHANGES**

Adversarial pass on process-role → non-blocking-backend selection (availability/fail-open class from #2280). Diff-only: no runtime execution this run. CONFIRMED items are closed by static path inspection.

### Sinks enumerated
Pattern: long-lived processes that open the shared DB without importing `./process_role.js` (so `isServerProcess()` stays false → `defaultDbBackend(false)` → `sqlite`).

| Call site | Role | Covered? |
| --- | --- | --- |
| `src/actions.ts` | HTTP API (`dist/actions.js`, Docker CMD) | covered — `import "./process_role.js"` first |
| `src/index.ts` | MCP stdio (`dist/index.js`, `package.json` main) | covered — `import "./process_role.js"` before config/db |
| `src/openclaw_entry.ts` | OpenClaw memory plugin (`package.json` `openclaw.extensions` → `dist/openclaw_entry.js`); `ensureServer()` → `initDatabase()` + `NeotomaServer` | **NOT covered** — no `process_role` import |
| `src/mcp_ws_bridge.ts` / `src/mcp_dev_shim.ts` | spawn child MCP | N/A on parent (no DB); child is `index.js` → covered |
| `src/seed_schemas_entry.ts` | one-shot Fly `release_command` | intentional CLI (`sqlite`) |
| `src/cli/**`, `src/core/operations.ts` | one-shot / offline CLI | intentional CLI |
| `openDb()` / `ensureDbInitialized()` in `src/repositories/db/connection.ts` | canonical backend router | covered when caller marked server |
| Direct `WorkerFileDatabase` / `openLibsqlDatabase` | driver internals + tests | not process-role sinks |

### Encodings tested (static)
Against `NEOTOMA_DB_BACKEND` and `NEOTOMA_DB_MAX_QUEUED_STATEMENTS` resolvers in the diff:

| Form | Result |
| --- | --- |
| Case fold (`SQLITE` / `LiBsQl`) | `.toLowerCase()` → accepted when exact token |
| Leading/trailing whitespace (` libsql`, `sqlite `) | not trimmed → startup throw (fail-closed) |
| Empty / unset backend | role-based default |
| Unknown backend token | startup throw (fail-closed) |
| `NEOTOMA_PROCESS_ROLE=server` | ignored by design (module flag only) — pinned in tests |
| IPv4-mapped / DNS / percent-encoding / redirects | N/A — no host/URL guard in this change |
| `NEOTOMA_DB_MAX_QUEUED_STATEMENTS` unset / non-numeric / negative | → `512` (bound retained) |
| `=0` | disables bound (documented) |
| `=0x200` with `parseInt(..., 10)` | → `0` → unbound (see finding) |

### Fail-open branches
- [x] `isServerProcess()` default `false` → `sqlite` when entrypoint omits `process_role` — **availability fail-open** (OpenClaw hole)
- [x] Explicit `NEOTOMA_DB_BACKEND=sqlite` on a marked server — intentional operator override (re-enables #2280 class)
- [x] Invalid backend → throw (deny / fail-closed)
- [x] `resolveMaxQueued`: invalid → 512; `0` → unbounded (documented opt-out)
- [ ] No evidence in-diff that `WorkerDbOverloadError` is remapped to HTTP 503/`Retry-After` (ops surface; not demonstrated)

### Findings
[BLOCKING] CONFIRMED incomplete_sink_coverage: exported OpenClaw entrypoint sits beside guarded siblings and never marks the process a server

- `src/actions.ts` / `src/index.ts` import `./process_role.js`; `src/openclaw_entry.ts` does not, yet `ensureServer()` calls `initDatabase()` and hosts a long-lived `NeotomaServer` instance (cached in `serverInstance`).
- With `NEOTOMA_DB_BACKEND` unset, `resolveDbBackend()` → `defaultDbBackend(false)` → `sqlite` — the same blocking driver this PR exists to stop selecting for long-lived multi-caller processes.
- Refute path: add `import "./process_role.js"` (or equivalent `markServerProcess()`) as the first import in `openclaw_entry.ts`, plus a regression that the OpenClaw entry resolves to `libsql` when unset.

[NON-BLOCKING] PLAUSIBLE queue_bound_bypass: `Number.parseInt(env, 10)` turns hex-looking values into `0`

- e.g. `NEOTOMA_DB_MAX_QUEUED_STATEMENTS=0x200` → `0` → unbound queue (restores pre-bound OOM class). Operator/env footgun; confirm with a unit assert if desired. Trim + reject non-decimal or treat `0x*` as invalid.

[NON-BLOCKING] PLAUSIBLE narrowed_docs: `src/repositories/db/connection.ts` header still says "`sqlite` default"

- Stale relative to role-based default; not a runtime defect. Route to docs/arch if desired.

### Narrowed protection check
- [x] New per-connection queue bound is additive (default 512); `0` is an explicit disable, not a silent widen of another control.
- [x] Explicit env override still wins — correct, but means a mis-set `NEOTOMA_DB_BACKEND=sqlite` on servers reintroduces the #2280 stall class by design.

### Verdict
`REQUEST_CHANGES` — one CONFIRMED uncovered long-lived sink (`openclaw_entry`). Marked HTTP/MCP stdio paths look correct on static inspection; queue-bound env quirk is NON-BLOCKING.

---
📎 Neotoma: [neotoma#2285](https://neotoma.markmhendrickson.com/entities/ent_1c2d6e4e5b66d9c8dc575312) · security_finding `ent_34cb32bfd21238b1e0a6c54b`

markmhendrickson:
## Rebased onto current `main` (post-#2284) — was `DIRTY`, now `MERGEABLE`

Head moved `1703d82e` → `d5d20600`. Both commits preserved, now sitting on top of #2284.

**The conflict was only in the generated test catalog**, not in code. `docs/testing/automated_test_catalog.md` is a generated inventory, and both sides had added test files to it — main gained one, this branch adds three. Hand-merging a generated file would be the wrong resolution, so I took main's version and re-ran `scripts/generate-automated-test-catalog.ts` against the actual tree:

| | Total | Backend Vitest | Unit |
|---|---|---|---|
| `origin/main` | 587 | 552 | 163 |
| this branch, regenerated | **590** | **555** | **166** |

Exactly +3, matching the three test files this PR adds — main's new test is picked up too. The generator output is what is committed, so a fresh run reproduces it byte-for-byte.

**No Fly config touched.** `git diff --name-only origin/main...HEAD` returns the same 11 files as before the rebase, none of them `fly*.toml`. #2284's sizing corrections (`performance` / 2 / `8gb`) and its `/ready` check are untouched — nothing here reintroduces the values it fixed.

### Verified on the rebased branch

- `tsc --noEmit` clean.
- DB suites **58/58**: the 15 new tests, plus `db_worker_file_database`, `db_libsql_nonblocking`, `db_driver_contract`, `config_db_backend`, `db_url_misconfig_guard`.
- **#2284's own `readiness_probe.test.ts`: 10/10** — worth checking explicitly, since `/ready` is the surface where these two PRs meet.

Full-suite comparison against the new `main` is running; I will post it if anything differs from the zero-regression result reported earlier.

### One note on how these two changes compose

#2284 repointed the Fly health check from `/health` to `/ready`, which does a real database read — so a green check can no longer coexist with a total read outage. Under the synchronous driver that probe would have been subject to the very event-loop stall this PR removes: it would have been measuring something real and then unable to answer it within the timeout. The pair is better than either alone — the check now measures something real *and* can respond while the database is busy.

neotoma-agent:
**BLOCKED**

Could not evaluate PR gate inheritance for #2280 — the gate-status read tool failed with a transport-level error, not a data result:

```
request_failed: POST /entities/query — ConnectError: [SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:1081)
```

Retried twice; same error both times. This is a connectivity/product issue on the gate-status read path, not a "gates pending" finding — I have no grounded `gate_status` snapshot for #2280 to report against, and per the grounding rule I will not infer a gate state without one.

- [ ] pm — unknown (unreadable)
- [ ] ux — unknown (unreadable)
- [ ] arch — unknown (unreadable)

**Resolution paths:**
1. Retry once connectivity recovers — I'll re-run PR gate inheritance automatically on the next dispatch, or can be re-invoked directly.
2. Operator override: `/confirm-gates-clear` — only @markmhendrickson may issue this command; it waives all unsigned pre-impl gates and re-triggers the PR pipeline.
3. Normal path: once readable, the gate owner can sign off via Neotoma (`gate_status.<gate>` → `signed_off`), or the operator can waive individually.

Note for triage: this matches a live pattern — several other recent Lanius dispatch runs have also failed in the last few hours (`~/Library/Logs/ateles/dispatch-failures/lanius-*.log`), consistent with a tooling issue on the `get_gate_status` → Neotoma `/entities/query` path rather than an issue specific to #2280.

GATE_INHERITANCE: blocked
GATE_PENDING: pm,ux,arch

markmhendrickson:
## Full suite on the rebased branch — still zero regressions

Ran against current `main` (post-#2284/#2278), `dist/` built:

| | Test files | Tests |
|---|---|---|
| Pre-rebase | 12 failed / 534 passed (550) | 22 failed / **5356** passed |
| **Post-rebase** | 12 failed / 536 passed (552) | 22 failed / **5388** passed |

**Delta: +32 passing, +2 test files, +0 failures** — all of it from what main gained in #2284/#2278.

The failing set is **byte-identical to before the rebase** — a `diff` of the two failing-file lists is empty:

```
7  tests/integration/cursor_hook_stop_backfill.test.ts
4  tests/integration/inspector_content_negotiation.test.ts
3  tests/integration/hook_failure_hint.test.ts
2  tests/integration/v0.2.0_ingestion.test.ts
1  tests/integration/turn_summary.test.ts
1  tests/integration/mcp_target_id_identity_conflict.test.ts
1  tests/integration/fixture_mcp_store_replay.test.ts
1  tests/integration/embed_cross_origin_http.test.ts
1  tests/integration/csp_local_http.test.ts
1  tests/fixtures/replay_graph.test.ts
```

All pre-existing integration failures, unrelated to the DB layer, and all present on clean `main` (verified earlier against a separate baseline worktree: 263 failed / 4997 passed with no `dist/`, versus this branch's 208 under the same no-`dist/` conditions).

All three new test files pass:

```
✓ tests/unit/db_worker_transaction_integrity.test.ts (4 tests) 241ms
✓ tests/unit/db_worker_backpressure.test.ts          (3 tests)  59ms
✓ tests/unit/db_backend_server_default.test.ts       (8 tests)   8ms
```

CI on the rebased head: both required checks pass (`security_gates`, `graph_render_smoke`), plus `frontend`, `python_sdk`, `agentic_evals`, `contract_parity`, `eval_scenarios`, `eval_combined`. `mergeable: MERGEABLE`.

neotoma-agent:
> ⚠️ Superseded by commit `d5d2060` — this verdict no longer reflects the current head. See the latest review below.


review:pm
**REQUEST_CHANGES**

PM lens review of PR #2285 against parent issue #2280. Diff-only; no runtime execution this run. Re-reviewed after rebase onto post-#2284 `main` — blocking finding unchanged.

### Scope vs #2280

- [x] **Problem alignment** — Issue symptom is event-loop stall from synchronous SQLite on long-lived servers; PR selects the existing #1944 worker pool instead of building duplicate machinery. Matches suggested direction (1).
- [x] **Correctly deferred** — Statement runtime bounds (#2217) and response-size caps remain out of scope; PR notes them as follow-ups. No creep there.
- [x] **In-scope hardening** — Backpressure (`WorkerDbOverloadError`) and transaction-integrity tests are justified follow-through for making the worker pool the server default, not unrelated feature work.
- [ ] **Server entrypoint completeness** — PR claims servers structurally default to the non-blocking backend; one published long-lived entrypoint is uncovered (see blocking finding).

### Acceptance criteria (standing PM quality gate)

Per policy `fixed_means_behavior_verified_not_contract_accepted` (ent_db0b7855d47012084477fb00):

- [x] **Effect verified (backend path)** — Existing `tests/unit/db_libsql_nonblocking.test.ts` asserts the reported effect: a concurrent trivial query completes promptly while a multi-second query runs (health-check proxy). Not merely contract acceptance.
- [x] **Effect wiring (partial)** — New `tests/unit/db_backend_server_default.test.ts` pins server vs CLI default selection and explicit-env override.
- [ ] **End-to-end server coverage** — Selection tests cover `actions.ts` and `index.ts` paths only; uncovered OpenClaw entry leaves a long-lived server on the blocking driver when `NEOTOMA_DB_BACKEND` is unset.
- [x] **Cross-surface parity** — N/A; internal driver selection by process role, not a multi-surface API capability.

### Findings

[BLOCKING] scope: published OpenClaw server entrypoint not marked — still defaults to blocking `sqlite`

`src/openclaw_entry.ts` is a long-lived server (`ensureServer()` → `initDatabase()` + `NeotomaServer`) exported via `openclaw.plugin.json` / `package.json` `openclaw.extensions`, but this PR does not import `./process_role.js` there. With `NEOTOMA_DB_BACKEND` unset, `resolveDbBackend()` → `defaultDbBackend(false)` → `sqlite` — the same #2280 stall class the PR exists to eliminate for servers. HTTP (`src/actions.ts`) and MCP stdio (`src/index.ts`) are covered; OpenClaw is not. This contradicts the PR’s “every deployment, structurally” intent.

**Fix:** add `import "./process_role.js"` as the first import in `src/openclaw_entry.ts` and a regression test that the OpenClaw entry resolves to `libsql` when unset. (Aligns with Falco security finding on the same sink.)

[NON-BLOCKING] scope: acceptance criteria could link selection + effect in one test

`db_backend_server_default.test.ts` verifies config selection only; effect relies on the pre-existing `db_libsql_nonblocking` regression. Sufficient for sign-off once sinks are complete, but an integration test through a server entrypoint import chain would tighten the DoD chain. Unverified whether that gap matters in CI beyond the two separate tests.

[NON-BLOCKING] scope: docs table lists HTTP + MCP only

`docs/operations/configuration.md` server table omits OpenClaw even if fixed — minor doc parity once the entrypoint is patched.

### PM gate

`gate_status.pm` on #2280 remains **pending** — one blocking scope gap. Re-review after OpenClaw entrypoint coverage.

---
📎 Neotoma: [neotoma#2280](https://neotoma.markmhendrickson.com/entities/ent_5ed9a2c5b3d8c2860d8e5d6f)

neotoma-agent:
> ⚠️ Superseded by commit `d5d2060` — this verdict no longer reflects the current head. See the latest review below.


review:arch
**APPROVE**

Scope check: this PR touches process-role selection (`src/process_role.ts`, `src/process_role_state.ts`), a `config.ts` default-backend getter, a bounded-queue addition in `worker_file_database.ts`, and docs/tests. No OpenAPI route, MCP tool, CLI command, response shape, schema field, or entity relationship is added or changed. Most of the arch lens (contract-first sequencing, response-field schema declaration, tenant isolation on entity lookups, idempotency_key on mutating ops) is **not applicable** to this diff — there is no client-facing contract surface here.

**Layering** — clean. `process_role.ts` (side-effect import, `markServerProcess()`) is split from `process_role_state.ts` (pure state read, `isServerProcess()`) specifically so `config.ts` can read the flag without importing the side effect. That's the right shape for this: a module-scoped flag rather than an env var, with the rationale for that choice (env vars leak into children/`.env`/test workers) documented inline and pinned by `db_backend_server_default.test.ts`'s ambient-env-var test. `config.dbBackend` moving from a captured module-load value to a lazy getter is consistent with `resolveDbBackend()` needing to run after `markServerProcess()` fires, which entrypoints call before importing anything that reaches `config.js` — ordering is explicitly documented as non-load-bearing since resolution is lazy, which is the correct belt-and-suspenders design.

**Contract-first** — N/A, no interface surface added. `NEOTOMA_DB_BACKEND` and the new `NEOTOMA_DB_MAX_QUEUED_STATEMENTS` env var are operational config, not API contract, and are documented in `docs/operations/configuration.md` in the same PR.

[NON-BLOCKING] error-envelope-mapping: `WorkerDbOverloadError` (`src/repositories/worker/worker_file_database.ts`) is a new thrown error type on a hot path (every server-side DB statement can now reject with it under saturation), but this diff does not show it being caught and translated into the standard `ErrorEnvelope`/`error_code` taxonomy documented in `docs/subsystems/errors.md` (the doc's DB-error codes are `DB_CONNECTION_FAILED`/`DB_QUERY_FAILED`/`DB_CONSTRAINT_VIOLATION`, none of which obviously fit an overload/backpressure case). This is unverified — I'm reviewing diff-only with no checkout to grep `src/actions.ts`'s top-level catch handler or confirm whether an unmapped error already falls through to a generic 500 with `DB_QUERY_FAILED` today. What would confirm it: whether `actions.ts`'s error middleware maps unknown thrown errors to a generic `error_code`, and if so whether that generic code is descriptive enough for a caller to distinguish "saturated, retry" from "query failed, don't retry" per the doc's `Retry?` column. If it doesn't already map cleanly, a follow-up should add a dedicated `error_code` (e.g. `DB_OVERLOADED`, retryable) rather than letting this fall through as a generic 500 — the error message is caller-actionable (names the env var) but only if it survives to the client instead of being flattened.

[NON-BLOCKING] test-import-path: `db_backend_server_default.test.ts` imports `defaultDbBackend` from `../../src/config.ts` (source `.ts`, not `.js`) while production code (`src/actions.ts`, `src/index.ts`) imports `./process_role.js`. This is consistent with existing repo test convention if other unit tests already import `.ts` directly under the vitest resolver — unverified without a checkout, flagging only because it's the one place the diff mixes extensions across files, not because it looks wrong.

No blocking findings from the arch lens. The process-role split is a reasonable, low-coupling, easily-reversible mechanism (deleting `process_role.ts`'s import from the two entrypoints reverts to the old always-`sqlite` default; flipping `defaultDbBackend`'s ternary reverts the polarity) — this is the kind of decision that's cheap to undo if the server/CLI default split turns out wrong in practice.

📎 Neotoma: PR reviewed diff-only per dispatcher instructions; no local checkout available to verify call-site error mapping.

markmhendrickson:
## CI green on the rebased head — `mergeStateStatus: CLEAN`

All 11 checks pass on `d5d20600b`, queried against the commit SHA directly rather than read off the PR checks view:

```
success  agentic_evals        success  graph_render_smoke
success  baseline             success  preview
success  contract_parity      success  python_sdk
success  eval_combined        success  security_gates
success  eval_scenarios       success  site_export
success  frontend
```

Both required checks — `security_gates` and `graph_render_smoke` — pass. State is `MERGEABLE` / `CLEAN`.

One methodological note, since it nearly produced a wrong claim from me: `gh pr checks` briefly showed a mix of results from the pre-rebase head (`1703d82e5`) and the rebased one, because both had runs in flight. I re-queried `repos/.../commits/d5d20600b/check-runs` to confirm every green is bound to the current SHA. The pre-rebase head did also pass fully — same code — but that is not what "this PR is green" should rest on.

**Not merging and not deploying.** The operator is sequencing the production restore.

neotoma-agent:
> ⚠️ Superseded by commit `d5d2060` — this verdict no longer reflects the current head. See the latest review below.


review:arch
**SIGNED_OFF**

## ADR — Server DB backend selection (#2280 / #2285)

**Decision** — How long-lived server processes select the non-blocking (`libsql`/worker-pool) DB backend without deployment-env drift or ambient env contamination of CLI/test workers.

**Options + tradeoffs**
| Option | Coupling | Reversibility | Fit |
|---|---|---|---|
| Env var in every deploy file (`NEOTOMA_DB_BACKEND=libsql`) | Low code change; high ops drift (the #2280 failure mode) | Easy | Already existed; never opted in |
| Env var for process role (`NEOTOMA_PROCESS_ROLE=server`) | Single knob; leaks to child processes and `.env` | Easy | Rejected in PR — observed vitest/CLI contamination |
| **Entrypoint side-effect import (`process_role.js`)** | Code marks server at bootstrap; covers all start paths | Easy — remove import or set explicit `NEOTOMA_DB_BACKEND` | Matches existing `DbDatabase` abstraction; reuses #1944 pool |
| New async driver / second pool | High implementation cost | Hard | Duplicates reviewed machinery |

**Chosen approach** — Entrypoint import + module-scoped flag (`process_role_state.ts`) + lazy `config.dbBackend` getter. Explicit `NEOTOMA_DB_BACKEND` still wins. Backpressure bounded via `NEOTOMA_DB_MAX_QUEUED_STATEMENTS` / `WorkerDbOverloadError`.

- **No external interface change** — no OpenAPI, `contract_mappings`, MCP tool, CLI command, response field, or agent-instruction surface touched. Contract-first gate: N/A (correctly absent from diff).
- **Layering** — Clean split: `process_role.ts` (side-effect), `process_role_state.ts` (readable flag), `config.ts` (lazy resolution at first DB use), `worker_file_database.ts` (queue bound at repository layer). Server entrypoints covered: `src/actions.ts`, `src/index.ts` (MCP stdio; `mcp_ws_bridge` spawns this).
- **Tenant isolation / idempotency** — No entity-lookup or new mutating API ops in diff. N/A.

**Reversibility** — High. Remove entrypoint imports or set `NEOTOMA_DB_BACKEND=sqlite`. Backpressure disable via `NEOTOMA_DB_MAX_QUEUED_STATEMENTS=0` (documented as not recommended).

## Arch-lens checklist

- [x] Contract-first — no handler/API contract changes; OpenAPI parity not required
- [x] Layering — config/repository/entrypoint boundaries respected; lazy getter avoids import-order race
- [x] Schema for response fields — N/A (internal infra)
- [x] Tenant isolation — N/A (no lookup changes)
- [x] Idempotency on mutating ops — N/A (no new store endpoints)

## Findings

[NON-BLOCKING] **entrypoint-coverage**: `src/openclaw_entry.ts` imports `db.js` without `process_role.js`. If the OpenClaw plugin runs as a long-lived server, it would retain the CLI default (`sqlite`). Not demonstrated in this diff; not blocking #2280 scope (HTTP + MCP stdio), but worth a follow-up if OpenClaw is production-deployed.

[NON-BLOCKING] **error-envelope** (unverified): `WorkerDbOverloadError` is a new internal repository error. Diff does not touch HTTP/MCP error middleware. If overload propagates to clients, it should map to the canonical retryable envelope per `docs/subsystems/errors.md` — cannot confirm from diff alone; existing global handler may already cover it.

[NON-BLOCKING] **doc-consistency**: `src/actions.ts` says `process_role` "MUST be first"; `src/process_role.ts` says import ordering is "NOT load-bearing" (lazy resolution). Comments only; behavior is consistent.

**Steelman (env-var approach)**: A single deploy-time env var is simpler for operators who read config files but not source entrypoints — valid if every deployment surface is kept in sync (the exact drift #2280 exposed).

**Key assumption**: All production server traffic enters via `dist/actions.js` or `dist/index.js` (or spawns the latter). Would revise if a third long-lived entrypoint loads DB without `process_role.js`.

---
📎 Neotoma: [neotoma#2280](https://neotoma.markmhendrickson.com/entities/ent_5ed9a2c5b3d8c2860d8e5d6f)

neotoma-agent:
> ⚠️ Superseded by commit `d5d2060` — this verdict no longer reflects the current head. See the latest review below.


review:security
**COMMENT**

## Sinks enumerated

Pattern: long-lived server processes must resolve `config.dbBackend` → `libsql` (worker pool) via `isServerProcess()` / `resolveDbBackend()` in `src/config.ts`, consumed by `src/repositories/db/connection.ts` (`openDb`, `ensureDbInitialized`).

| Call site | Role | Covered? |
|---|---|---|
| `src/actions.ts` | HTTP server entry (`dist/actions.js`, Docker CMD) | **Yes** — imports `./process_role.js` |
| `src/index.ts` | MCP stdio server entry (`dist/index.js`) | **Yes** — imports `./process_role.js` |
| `src/mcp_ws_bridge.ts` | WS bridge | **Indirect** — spawns `dist/index.js` child |
| `src/mcp_dev_shim.ts` | MCP dev shim | **Indirect** — spawns configured MCP child (default `dist/index.js`) |
| `src/openclaw_entry.ts` | OpenClaw in-process plugin server (`initDatabase` + `NeotomaServer`) | **NOT covered** — no `process_role` import |
| `src/seed_schemas_entry.ts` | One-shot Fly release command | **N/A** — CLI/one-shot; sync backend correct |
| `src/cli/index.ts` | CLI | **N/A** — intentionally `sqlite` default |
| Direct `process.env.NEOTOMA_DB_BACKEND` reads bypassing `resolveDbBackend` | — | **None found** (repo-wide search via branch tree + raw file inspection) |

## Encodings tested

Static analysis on branch `fix/2280-sqlite-worker-pool` (no checkout/runtime):

| Input | Guard behavior | Reaches sink? |
|---|---|---|
| `NEOTOMA_DB_BACKEND` case variants (`LIBSQL`, `Sqlite`) | `.toLowerCase()` before match | Explicit value honored |
| `NEOTOMA_DB_BACKEND=""` (unset/empty) | Falls through to `defaultDbBackend(isServerProcess())` | Role-based default |
| `NEOTOMA_DB_BACKEND` with leading/trailing whitespace | Fails module-load validation (not in allowlist) | Process throws at startup — **deny** |
| Ambient `NEOTOMA_PROCESS_ROLE=server` | Ignored (regression test in diff) | CLI stays `sqlite` — **correct deny of env spoof** |
| `NEOTOMA_DB_MAX_QUEUED_STATEMENTS` non-numeric / negative | `resolveMaxQueued()` → default 512 | Bounded queue |
| `NEOTOMA_DB_MAX_QUEUED_STATEMENTS=0` | Disables per-connection bound | **Unbounded queue restored** (documented operator knob) |

IPv4/percent-encoding/redirect encodings: **N/A** — backend selection is not URL/host parsing.

## Fail-open branches

| Branch | Default | Assessment |
|---|---|---|
| `NEOTOMA_DB_BACKEND` unset + `isServerProcess()===false` | `sqlite` (blocking) | Intended for CLI; **fail-open to blocking on any server entry missing `process_role`** |
| `NEOTOMA_DB_MAX_QUEUED_STATEMENTS=0` | Unbounded worker `pending` map | Documented; restores pre-PR OOM/DoS surface under saturation |
| Invalid `NEOTOMA_DB_MAX_QUEUED_STATEMENTS` | Default 512 | Fail-safe toward bounded |
| Invalid non-empty `NEOTOMA_DB_BACKEND` | Throw at module load | Fail-closed |
| `resolveMaxQueued()` at `WorkerConnection` construct time | Frozen for process lifetime | Expected; not a runtime bypass |

## Findings

- [NON-BLOCKING] PLAUSIBLE **incomplete-sink-coverage**: `src/openclaw_entry.ts` is a long-lived server entrypoint that calls `initDatabase()` and hosts `NeotomaServer` but does not import `./process_role.js`. With `NEOTOMA_DB_BACKEND` unset it resolves to synchronous `sqlite`, preserving the event-loop blocking DoS class #2280 fixes for HTTP/MCP. **Confirm:** deploy OpenClaw plugin without env override; verify `config.dbBackend` at first query. **Refute:** if OpenClaw is classified out-of-scope for “server” in ops docs.

- [NON-BLOCKING] PLAUSIBLE **fail-open-backpressure**: `NEOTOMA_DB_MAX_QUEUED_STATEMENTS=0` explicitly disables the new queue bound (`worker_file_database.ts`), restoring unbounded heap growth under DB saturation — worse than slow responses. Documented as “not recommended for servers”; no guard prevents setting it in deployment config. **Confirm:** set env to `0` under load test; observe unbounded `pending.size`.

- [NON-BLOCKING] PLAUSIBLE **availability-handling**: `WorkerDbOverloadError` is new; no diff changes map it to a retry-safe HTTP/MCP status in action handlers. Under adversarial load, clients may see generic 500s and retry, amplifying saturation. **Confirm:** overload integration test through `/ready` or MCP tool path.

## Narrowed-protection check

- [x] Explicit `NEOTOMA_DB_BACKEND` still wins over role default (unchanged operator override)
- [x] CLI default remains `sqlite` (no widening of worker spawn to one-shot processes)
- [x] Lazy `config.dbBackend` getter avoids early capture before `markServerProcess()` — improvement vs prior const
- [x] Backpressure added (mitigates unbounded-queue OOM that existed pre-PR on worker backend)

## Verdict

**COMMENT** — No diff-evident CONFIRMED security break under the diff-only evidence bar. One uncovered server sink (`openclaw_entry.ts`) and one documented fail-open knob (`MAX_QUEUED=0`) warrant follow-up; neither is demonstrated exploitable from the diff alone.

---
📎 Neotoma: neotoma#2280 · neotoma#2285

neotoma-agent:
> ⚠️ Superseded by commit `d5d2060` — this verdict no longer reflects the current head. See the latest review below.


review:ux
**SIGNED_OFF**

## Design Spec

### User-facing surface
- **Configuration:** `NEOTOMA_DB_BACKEND` (explicit override), new `NEOTOMA_DB_MAX_QUEUED_STATEMENTS`, existing `NEOTOMA_DB_READER_WORKERS`
- **Process role:** side-effect import `import "./process_role.js"` on server entrypoints (`src/actions.ts`, `src/index.ts`) — not env-configurable
- **Runtime errors:** `WorkerDbOverloadError` when worker queue saturates; existing startup throw for invalid `NEOTOMA_DB_BACKEND`
- **Docs:** `docs/operations/configuration.md` (primary operator reference)

### Interaction / flow
| Actor | Unset `NEOTOMA_DB_BACKEND` | Explicit override |
|---|---|---|
| Server (HTTP actions, MCP stdio) | `libsql` (worker pool) | env wins |
| CLI / scripts | `sqlite` (sync) | env wins |

Resolution is lazy at first DB use via `config.dbBackend` getter — entrypoint import order is not required for correctness (see friction note below).

### Discoverability & naming
- [x] `configuration.md` replaces opt-in framing with “How the backend is chosen” table and backpressure section
- [x] `process_role.ts` / `process_role_state.ts` explain why role is not an env var (prevents inherited `.env` silent flip)
- [x] `WorkerDbOverloadError` message names queue limit, retry guidance, and both tuning env vars

### Error & empty states
- **Invalid backend at startup:** `Invalid NEOTOMA_DB_BACKEND "<value>": expected "sqlite" or "libsql"` — fail loud, unchanged
- **Queue saturation:** `WorkerDbOverloadError` with reader/writer role, in-flight count, limit, `NEOTOMA_DB_MAX_QUEUED_STATEMENTS`, `NEOTOMA_DB_READER_WORKERS` — good operator-facing shape at the worker layer
- **Empty/unset backend:** role-based default; covered by `db_backend_server_default.test.ts`

### Accessibility
N/A — backend selection and overload paths are server/operator surfaces, not UI components.

## Findings

[NON-BLOCKING] docs/cross-reference: `docs/infrastructure/deployment.md` (unchanged in this PR) still instructs operators to opt in via `NEOTOMA_DB_BACKEND=libsql` and links to the removed `configuration.md` § “When to opt into `NEOTOMA_DB_BACKEND=libsql`”. After merge, deployment guidance contradicts the new server default. Sync deployment.md to “How the backend is chosen” (same PR or immediate follow-up).

[NON-BLOCKING] docs/consistency (evident in diff): `src/actions.ts` and `src/index.ts` comments say `process_role.js` “MUST be first” because imports above it “would lock in the CLI default”, but `src/process_role.ts` states ordering is **not** load-bearing because `config.dbBackend` resolves lazily. Align comments to the lazy-resolution model so future entrypoint authors are not misled.

[NON-BLOCKING] observability (unverified — no runtime in this review): diff adds no startup log of resolved `dbBackend` + process role. Operators debugging “still blocking” reports cannot confirm selection from logs alone. Would confirm by starting a server with unset backend and checking stdout; suggest one `[neotoma] db backend=libsql role=server` line at first `openDb()`.

[NON-BLOCKING] api-error-shape (unverified — not in diff): `WorkerDbOverloadError` message is actionable at the worker layer, but this PR does not show HTTP/MCP error mapping (e.g. retryable 503 vs opaque 500). Would confirm by forcing queue overflow and inspecting API response body/status. If unmapped today, follow-up should attach `retryable: true` + structured hint per existing store-policy 503 pattern.

[NON-BLOCKING] discoverability/entrypoints: only `actions.ts` and `index.ts` import `process_role.js`. Docs name those entrypoints; a one-line “new long-lived server entrypoints must import `./process_role.js`” note in `configuration.md` or `process_role.ts` header would close the gap for future surfaces.

## Acceptance checklist
- [x] Operator docs explain server vs CLI default and explicit override path
- [x] Backpressure env var documented with default and semantics
- [x] Invalid `NEOTOMA_DB_BACKEND` fails at startup with actionable message
- [x] Overload error message names limit and remediation env vars (worker layer)
- [x] Tests pin role-based defaults and reject ambient `NEOTOMA_PROCESS_ROLE` env
- [ ] `deployment.md` cross-references updated to match new default framing (follow-up)
- [ ] Entrypoint import-order comments consistent with lazy resolution (follow-up)
- [ ] Startup log or `/ready`/`/health` exposes active backend for operators (unverified; follow-up if absent)

UX gate signed off. No diff-evident blockers; follow-ups above are polish and cross-doc sync.

---
📎 Neotoma: [neotoma#2285](https://neotoma.markmhendrickson.com/entities/ent_ea8cc3f4e1b72d39f218acf1)

neotoma-agent:
> ⚠️ Superseded by commit `d5d2060` — this verdict no longer reflects the current head. See the latest review below.



**REQUEST_CHANGES**

Aggregated review panel verdict for PR #2285 against parent issue #2280.

Reviewed commit: d5d20600b3da18d6f8e00a39ce7803b1dd0ccb7e
Verdict: REQUEST_CHANGES
Blocking: 1

### Per-lens roll-up

| Lens | Verdict | Notes |
|---|---|---|
| **pm** (Pavo) | REQUEST_CHANGES | 1 blocking scope gap |
| **arch** (Waxwing) | SIGNED_OFF | Entrypoint side-effect import + lazy `config.dbBackend`; no contract changes |
| **security** (Falco) | COMMENT | No blocking findings; OpenClaw sink flagged PLAUSIBLE |
| **ux** (Accipiter) | SIGNED_OFF | Role-based defaults + backpressure docs adequate; doc polish follow-ups |

### Blocking findings

- [BLOCKING] **scope** (pm, security): `src/openclaw_entry.ts` is a published long-lived server entrypoint but does not import `./process_role.js`. With `NEOTOMA_DB_BACKEND` unset it still defaults to blocking `sqlite`, contradicting the PR's "every deployment, structurally" claim. HTTP (`src/actions.ts`) and MCP stdio (`src/index.ts`) are covered; OpenClaw is not.

**Required fix:** add `import "./process_role.js"` as the first import in `src/openclaw_entry.ts` and a regression test that the OpenClaw entry resolves to `libsql` when unset.

### Non-blocking follow-ups (post-fix)

- [NON-BLOCKING] docs: sync `docs/infrastructure/deployment.md` to new server-default framing (ux)
- [NON-BLOCKING] docs: add OpenClaw to server entrypoint table in `configuration.md` once patched (pm)
- [NON-BLOCKING] comments: align "MUST be first import" wording with lazy-resolution model (ux/arch)
- [NON-BLOCKING] observability: startup log of resolved `dbBackend` + process role (ux)
- [NON-BLOCKING] error-envelope: verify HTTP/MCP mapping for `WorkerDbOverloadError` (ux/security)

### Gate inheritance

- [ ] **pm** — `pending` (blocking scope gap above)
- [x] **arch** — `signed_off`
- [x] **ux** — `not_required` / signed off by lens
- [ ] **pr_review** — `changes_requested` (this verdict)

Gate inheritance **blocks merge** regardless of autonomy flag.

### Merge recommendation

**Do not merge.** Two independent blockers:

1. **Pre-impl gate:** `gate_status.pm` on #2280 remains `pending` until OpenClaw entrypoint coverage lands and Pavo re-reviews.
2. **Panel blocker:** PM lens `[BLOCKING]` finding on uncovered OpenClaw server sink.

Route to **Gryllus** (impl) for the OpenClaw entrypoint fix + regression test, then re-run panel.

`APIS_AUTONOMY_AUTO_MERGE=1` is set but authorization is withheld — gate inheritance outranks autonomy.

### Checklist

- [x] Core approach matches #2280 (reuse #1944 worker pool, not duplicate machinery)
- [x] Effect verified for libsql path (`db_libsql_nonblocking.test.ts`)
- [x] Server default selection tests for HTTP + MCP entrypoints
- [ ] All long-lived server entrypoints marked with `process_role.js`
- [ ] Pre-impl gates clear (pm pending)
- [ ] Fresh head-SHA-matched APPROVE with Blocking: 0

---
📎 Neotoma: [neotoma#2280](https://neotoma.markmhendrickson.com/entities/ent_5ed9a2c5b3d8c2860d8e5d6f)

neotoma-agent (changes requested):
## Aggregated Review Verdict — PR #2285

**REQUEST_CHANGES**

Aggregated review panel verdict for PR #2285 against parent issue #2280.

Reviewed commit: `d5d20600b3da18d6f8e00a39ce7803b1dd0ccb7e`
Verdict: REQUEST_CHANGES
Blocking: 1

### Per-lens roll-up

| Lens | Verdict | Notes |
|---|---|---|
| **pm** (Pavo) | REQUEST_CHANGES | 1 blocking scope gap |
| **arch** (Waxwing) | SIGNED_OFF | Entrypoint side-effect import + lazy `config.dbBackend`; no contract changes |
| **security** (Falco) | COMMENT | No blocking findings; OpenClaw sink flagged PLAUSIBLE |
| **ux** (Accipiter) | SIGNED_OFF | Role-based defaults + backpressure docs adequate; doc polish follow-ups |

### Blocking findings

- [BLOCKING] **scope** (pm, security): `src/openclaw_entry.ts` is a published long-lived server entrypoint but does not import `./process_role.js`. With `NEOTOMA_DB_BACKEND` unset it still defaults to blocking `sqlite`, contradicting the PR's "every deployment, structurally" claim. HTTP (`src/actions.ts`) and MCP stdio (`src/index.ts`) are covered; OpenClaw is not.

**Required fix:** add `import "./process_role.js"` as the first import in `src/openclaw_entry.ts` and a regression test that the OpenClaw entry resolves to `libsql` when unset.

### Non-blocking follow-ups (post-fix)

- [NON-BLOCKING] docs: sync `docs/infrastructure/deployment.md` to new server-default framing (ux)
- [NON-BLOCKING] docs: add OpenClaw to server entrypoint table in `configuration.md` once patched (pm)
- [NON-BLOCKING] comments: align "MUST be first import" wording with lazy-resolution model (ux/arch)
- [NON-BLOCKING] observability: startup log of resolved `dbBackend` + process role (ux)
- [NON-BLOCKING] error-envelope: verify HTTP/MCP mapping for `WorkerDbOverloadError` (ux/security)

### Gate inheritance

- [ ] **pm** — `pending` (blocking scope gap above)
- [x] **arch** — `signed_off`
- [x] **ux** — `not_required` / signed off by lens
- [ ] **pr_review** — `changes_requested` (this verdict)

Gate inheritance **blocks merge** regardless of autonomy flag.

### Merge recommendation

**Do not merge.** Two independent blockers:

1. **Pre-impl gate:** `gate_status.pm` on #2280 remains `pending` until OpenClaw entrypoint coverage lands and Pavo re-reviews.
2. **Panel blocker:** PM lens `[BLOCKING]` finding on uncovered OpenClaw server sink.

Route to **Gryllus** (impl) for the OpenClaw entrypoint fix + regression test, then re-run panel.

`APIS_AUTONOMY_AUTO_MERGE=1` is set but authorization is withheld — gate inheritance outranks autonomy.

### Checklist

- [x] Core approach matches #2280 (reuse #1944 worker pool, not duplicate machinery)
- [x] Effect verified for libsql path (`db_libsql_nonblocking.test.ts`)
- [x] Server default selection tests for HTTP + MCP entrypoints
- [ ] All long-lived server entrypoints marked with `process_role.js`
- [ ] Pre-impl gates clear (pm pending)
- [ ] Fresh head-SHA-matched APPROVE with Blocking: 0

Posted to PR: https://github.com/markmhendrickson/neotoma/pull/2285#issuecomment-5500504568

---
📎 Neotoma: [neotoma#2280](https://neotoma.markmhendrickson.com/entities/ent_5ed9a2c5b3d8c2860d8e5d6f)

**Gate writeback:** `pr_review` → `changes_requested`, `current_owner` → `gryllus` on issue `ent_5ed9a2c5b3d8c2860d8e5d6f`.

[vanellus] merge_decision: REQUEST_CHANGES — PR #2285 blocked; pm gate pending + [BLOCKING] OpenClaw entrypoint missing process_role.js; routed to Gryllus; not merged (gate inheritance)

🧠 Neotoma
- [neotoma#2280](https://neotoma.markmhendrickson.com/entities/ent_5ed9a2c5b3d8c2860d8e5d6f) — `pr_review: changes_requested`, `current_owner: gryllus`, `pm: pending`
- plan_contribution `ent_11515b8ad84594e3150782b4` — changes_requested summary (offline store)

neotoma-agent:

PR #2338, opened 2026-09-08
fix(db): retire an abandoned reader instead of terminating it mid-native-call (#2324)

### LINKED ISSUE #2324: Rust panic in neon binding aborts the process when a client disconnects mid-statement (#2316 fix incomplete)
## Summary

A Rust panic inside the `neon` native binding aborts the process (`SIGABRT`, `exit_code=134`) when a client disconnects while a DB request is in flight. This is the same crash class as #2316, but the fix in #2317 cannot reach it: #2317 attaches a no-op catch at the **JavaScript** layer, and this abort leaves a JS exception *pending* while control is inside **native** code. A `.catch()` cannot unwind a Rust assertion that has already fired.

## Evidence

Captured from a production instance, 2026-09-08 07:03:50–07:03:52 UTC:

```
    at EventTarget.onAbort (file:///app/dist/repositories/worker/worker_file_database.js:348:34)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:831:20)
    at EventTarget.dispatchEvent (node:internal/event_target:766:26)
    at abortSignal (node:internal/abort_controller:370:10)
    at AbortController.abort (node:internal/abort_controller:392:5)
    at ServerResponse.onClose (file:///app/dist/middleware/db_abort_context.js:32:28)
    at Object.onceWrapper (node:events:638:28)
    at ServerResponse.emit (node:events:524:28)
    at emitCloseNT (node:_http_server:1019:10)
    at Socket.onServerResponseClose (node:_http_server:277:5)
Node.js v20.20.2

thread '<unnamed>' panicked at /cargo/registry/src/index.crates.io-6f17d22bba15001f/neon-1.0.0/src/sys/external.rs:80:5:
assertion `left == right` failed
  left: PendingException
 right: Ok
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
 INFO Main child exited with signal (with signal 'SIGABRT', core dumped? false)
```

## The chain

1. Client disconnects → `ServerResponse.onClose` (`middleware/db_abort_context.ts:32`) fires the abort.
2. `worker_file_database.ts` `onAbort` (~line 348, `failInFlight`/`abandon`) rejects the in-flight request.
3. A JS exception is left **pending** while control is inside native code.
4. `neon` asserts the JS engine state is `Ok`, observes `PendingException`, and panics.
5. A Rust panic across the FFI boundary aborts the process.

## Why #2317 does not cover it

#2317's `guardAbandonedRead` attaches a no-op catch to a *derived* promise at `WorkerStatement.get`/`.all`, scoped to reads. That handles an unhandled **JS** rejection. It does not prevent an exception being pending from neon's perspective at the moment the native module re-enters the JS engine, which is what trips the assertion.

## Reproduction

Observed reliably in production whenever a request is slow enough that the client gives up mid-statement. In the captured case a `POST /entities/query` took >25s and the disconnect triggered the panic. Reader-pool saturation makes it far more likely (see #2322) by lengthening request latency, but is not the cause: after raising `NEOTOMA_DB_READER_WORKERS` from 2 to 8 the pool no longer saturates and the panic still occurs on disconnect.

## Impact

The process aborts and restarts. With Fly's `restart: always` this becomes a crash-recovery loop: the instance is intermittently reachable, `/health` answers 200 (it does not touch the DB) while the first DB-backed query returns 502. It also feeds a secondary failure — each restart triggers the platform's autostart path, which exhausts the machines-API rate limit and blocks operator-initiated starts.

## Suggested direction

The abort path must not leave a pending JS exception while the worker is inside a native call. Options worth weighing: settle the rejection outside the native re-entry (defer to a microtask/`setImmediate` past the FFI boundary), or have the worker observe the abort cooperatively rather than rejecting from an event handler that can fire mid-call.

Relates to #2316 (original abort crash), #2317 (partial fix), #2322 (reader-pool exhaustion, an amplifier).

markmhendrickson:
Closes #2324. Implements the specification posted on the issue.

## The fault

`worker.terminate()` is built on V8's `TerminateExecution`, which — per V8's own `include/v8-isolate.h` — terminates **JavaScript** frames and has no mechanism to preempt a native one. The synchronous libsql driver spends a statement inside Rust, so terminating an abandoned reader mid-statement does not stop that Rust; it puts the isolate into a tearing-down state with the native call still on the stack. The next napi call Rust makes then fails, and neon 1.0.0 asserts on the status instead of propagating it (its siblings in `fun.rs`/`buffer.rs` handle it correctly; `external.rs` omits the guard).

`libsql-js` v0.5.29 pins `neon = "1.0.0"` and builds with `panic = "abort"`, so that assertion is a bare `abort()` — SIGABRT, exit 134, no unwinding, nothing any JavaScript-level guard can observe or catch. This is why #2317's `.catch()` could not contain it, and why deferring the rejection to a microtask would not have either.

## The fix

`abandon()` conflated reclaiming the **pool slot** (all #2217 actually requires) with **killing the thread** (the implementation chosen for it). Only the first is required.

- `failInFlight` still rejects callers immediately — nothing caller-visible changes.
- The connection is retired and dropped from the pool at that instant, so dispatch never sees it again and a replacement spawns on the next read.
- The orphan finishes its statement and posts a reply that the existing `if (!entry) return` guard discards. Termination happens from the `message` handler, where the worker is provably back in JavaScript.
- A worker that never reports is terminated anyway after `NEOTOMA_DB_ORPHAN_GRACE_MS` (default 30s) — the pre-#2324 behaviour kept strictly as a last resort, so a wedged thread cannot leak forever.

### The sibling, in the same change

Both callers of `abandon()` are fixed by construction. That matters: the **statement-timeout branch shares the defect**, so a single runaway query could panic with **no client disconnect at all**.

`close()`/`terminate()` share the shape at lower severity (the process is exiting anyway) and now wait for an in-flight statement to report before terminating, bounded by the same window. A retired connection keeps its orphan handle so closing the database actually waits it out rather than returning while the thread still runs — without that, `close()` was a no-op on exactly the connections it was tracking.

## The cost, stated plainly

Peak worker threads rise, because an abandoned reader lingers until its statement finishes instead of dying at once. Measured under identical load (4 readers, 8-way concurrent abandonment):

| build | peak worker threads | settles to |
|---|---|---|
| unpatched `63a7dcf88` | 5 | 1 |
| this branch | 14 | 1 |

Roughly **triple**, transient rather than a leak. This is worse than the spec's qualitative "peak thread count rises," and on a 2-CPU machine it is real contention. It remains the better trade — a lingering thread degrades throughput, whereas the abort takes the instance down and blocks its own recovery — and the grace window bounds it. §2.2 of the spec (cooperative `sqlite3_interrupt`) is what would make it cheap, and belongs in its own issue.

## The test

The four-case survival suite already existed and asserted the right property, but **ran in no CI lane** (`test:unit` excludes `tests/integration`; the lane allowlist omitted it) and would have passed on `better-sqlite3`, which contains no neon and cannot produce the panic at all. Repaired per the spec:

1. Asserts the driver resolved to `libsql` and **fails loudly rather than skipping** — this is what makes the suite able to fail at all.
2. Asserts no `panicked at` / `PendingException` in stderr and no SIGABRT, not merely that the process is alive.
3. Adds an `.all()` case whose per-100-row `rowsNext` re-entry widens the race from one native call to hundreds.

Wired into `ci_test_lanes.yml`.

### Demonstrated failing on unpatched main

On `origin/main` at `63a7dcf88` (i.e. with #2317 already in), with `libsql` resolving:

```
8 of 8 runs reproduced the abort
  5 at neon-1.0.0/src/sys/external.rs:80   <- the exact assertion in the production trace
  3 at neon-1.0.0/src/sys/array.rs:22      <- sibling, same defect class, reached via .all()
```

With this change: **6 of 6 runs green** across all five cases.

Iteration count was measured from both directions rather than guessed — 25 iterations reproduced only 3/6, 60 reproduced 6/8, 120 reproduced 8/8 — so the case runs 120. The abort lead is jittered rather than pinned, because a single measured value is tuned to one machine and silently stops reproducing elsewhere.

## Not covered

Writes and transactions remain unabortable by design. The upstream neon defect is routed around, not repaired — neon#1061 is still open at other assertion sites, so any *other* teardown of a libsql worker mid-native-call can still abort. This change removes the trigger we control.

### DIFF
--- a/.github/workflows/ci_test_lanes.yml
+++ b/.github/workflows/ci_test_lanes.yml
@@ -166,6 +166,28 @@ jobs:
       - name: Run entity-query read-latency regressions (#2266)
         run: npx vitest run --no-file-parallelism tests/integration/entity_query_deleted_count.test.ts
 
+      # #2316/#2324 DB-worker abandonment survival. The same gap as the two steps
+      # above, and the one that cost the most: this suite already existed and
+      # already asserted the right property, but `test:unit` excludes
+      # tests/integration and this lane's allowlist omitted it — so the crash
+      # shipped to production TWICE while four passing tests guarded nothing.
+      #
+      # It gates two distinct faults. #2316 is a JS unhandled rejection (exit 1).
+      # #2324 is a Rust panic inside libsql's neon binding (SIGABRT, exit 134),
+      # reached by terminating a worker while a synchronous native call is on its
+      # stack — unreachable by any JavaScript-level guard, since libsql-js builds
+      # with `panic = "abort"`.
+      #
+      # The lane MUST run where `libsql` resolves. The suite asserts the driver
+      # and FAILS rather than skips when it does not, because `better-sqlite3`
+      # contains no neon and cannot produce the panic at all: a run on the wrong
+      # driver is green for the wrong reason, which is exactly how this shipped.
+      # ubuntu-latest + `npm ci` installs @libsql/linux-x64-gnu, so it resolves
+      # here; if that ever stops being true this step turns red loudly instead of
+      # passing quietly, which is the intended failure mode.
+      - name: Run DB-worker abandonment survival regressions (#2316/#2324)
+        run: npx vitest run --no-file-parallelism tests/integration/db_abort_client_disconnect_survival.test.ts
+
       # OFFLINE/local-transport parity gate. Curated behaviors that broke on the
       # offline/HTTP path (#1819 at_ingested, #1840 dedup, #1839 null-cleared,
       # #1842 issue-submit auth) are asserted through BOTH the MCP and offline
--- a/src/repositories/worker/worker_file_database.ts
+++ b/src/repositories/worker/worker_file_database.ts
@@ -237,7 +237,36 @@ class WorkerConnection {
   private closed = false;
   /** `Date.now()` when this worker last went from idle to busy; 0 while idle. */
   private busyStartedAt = 0;
+  /**
+   * Set once this connection has been abandoned (#2324). A retired connection
+   * is out of the pool and never dispatched to again; its worker is still
+   * running only so it can finish its native call and be terminated safely.
+   */
+  private retired = false;
+  /** Whether anything was actually executing when this connection retired. */
+  private hadInFlightAtRetire = false;
+  /** Terminates the orphaned worker; cleared once it has run. */
+  private orphanFinish: (() => void) | undefined;
+  private orphanTimer: ReturnType<typeof setTimeout> | undefined;
+  /**
+   * The orphaned worker itself, kept because `failInFlight` has already nulled
+   * `this.worker` by the time the connection retires. Without this handle
+   * `close()` would find nothing to wait on and return while the orphan was
+   * still running — the thread would then outlive the database it was reading,
+   * which is precisely what tracking retired connections exists to prevent.
+   */
+  private orphanWorker: Worker | undefined;
 
+  /**
+   * The two objects are separate on purpose. `options` is handed to the Worker
+   * verbatim as `workerData`, so everything in it must be structured-cloneable:
+   * a callback there throws `DataCloneError` at spawn time, and — because that
+   * rejection carries a NUMERIC `code` — it then trips the string handling in
+   * `isReadonlyRejection`, surfacing as an unrelated `TypeError` on the next
+   * read rather than as anything resembling its cause. Connection-local
+   * settings the worker has no use for (the grace window, the retirement
+   * callback) therefore live in `local`, which never crosses the boundary.
+   */
   constructor(
     private readonly options: {
       dbPath: string;
@@ -246,8 +275,27 @@ class WorkerConnection {
       driverPath: string | null;
       useNodeSqlite: boolean;
       stripRowMetadata: boolean;
-    }
+    },
+    private readonly local: {
+      /** Grace window for an orphaned worker before forced termination. */
+      orphanGraceMs?: number;
+      /** Called when this connection retires, so its owner drops it. */
+      onRetire?: (connection: WorkerConnection) => void;
+    } = {}
   ) {}
+
+  private get orphanGraceMs(): number {
+    return this.local.orphanGraceMs ?? DEFAULT_ORPHAN_GRACE_MS;
+  }
+
+  private get onRetire(): ((connection: WorkerConnection) => void) | undefined {
+    return this.local.onRetire;
+  }
+
+  /** Whether this connection has been abandoned and left the pool (#2324). */
+  get isRetired(): boolean {
+    return this.retired;
+  }
 
   /**
    * How many statements this worker currently owns. The worker executes them
@@ -277,6 +325,16 @@ class WorkerConnection {
     if (this.worker) return this.worker;
     const worker = new Worker(WORKER_SOURCE, { eval: true, workerData: this.options });
     worker.on("message", (reply: WorkerReply) => {
+      // A reply from a RETIRED worker means the orphan has finished its native
+      // call and is back in JavaScript — the one moment at which terminating
+      // it cannot abort the process (#2324). Its result is discarded either
+      // way: the callers were already rejected by `failInFlight`.
+      if (this.retired) {
+        const finish = this.orphanFinish;
+        this.orphanFinish = undefined;
+        finish?.();
+        return;
+      }
       const entry = this.pending.get(reply.id);
       if (!entry) return;
       this.pending.delete(reply.id);
@@ -321,6 +379,7 @@ class WorkerConnection {
     // Snapshot first: reject() runs caller continuations that may re-enter
     // request() and repopulate the map while we are still iterating it.
     const entries = [...this.pending.values()];
+    this.hadInFlightAtRetire = entries.length > 0;
     this.pending.clear();
     this.busyStartedAt = 0;
     this.worker = null;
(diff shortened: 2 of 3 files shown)


neotoma-agent:

neotoma-agent:

markmhendrickson:
**The `contract_parity` cancellations are almost certainly memory exhaustion, and I have a local measurement.**

The operator flagged Claude using 45 GB. Tracing it: a single test process from this PR's survival suite was holding **14.7 GB RSS** after 4m41s —

```
14747552  59252  04:41  node .../neotoma-wt-2338-ci/node_modules/.neotoma-abort-test-RNNecr/server-slow-all-...
```

A second in another worktree was at 1.6 GB and climbing. Killing them took the machine's total `node` footprint from **15.8 GB to 0.9 GB**.

A GitHub-hosted runner has roughly 16 GB. A step that reaches 14.7 GB in under five minutes would exhaust it, and the runner's agent dying presents exactly as *"The runner has received a shutdown signal"* — which is what the first attempt reported, at 74 seconds into that step. Both attempts show the same step `cancelled` with the following two `skipped`, on different runners. That is the signature of the machine going down, not of a test failing.

**The suspected mechanism, which is worth treating as a finding about the fix rather than only about the test:** the `.all()`-over-many-rows case was added to widen the abort window, and `iterate()` re-enters native per 100-row batch. Under retire-instead-of-terminate the abandoned worker keeps running to completion — that is the design — so if it keeps producing rows after the caller has gone, they accumulate with nothing consuming them. The `server-slow-all-` name suggests exactly that path.

If that holds, it says something real about the approach: the orphan's cost is not only the thread, it is whatever the orphan continues to allocate. The prior agent measured peak threads roughly tripling; unbounded rows from an orphaned `.all()` is a second cost that measurement would not have surfaced.

**What I would not do:** cap the row count to make the test fit. The many-rows case exists because it widens the window from one native call to hundreds, and that is what made the defect reproduce 8/8 instead of ~50%. A bound that costs the reproduction costs the control.

Local reproduction is unaffected — this machine has the headroom, which is why it went unnoticed until the memory showed up.

markmhendrickson:
## Correction: a reply is not proof the worker is idle

The first commit followed the spec's §2.1 shape — terminate the orphan from the `message` handler, "where the worker is provably back in JavaScript." **That premise is false**, and running the suite against a *saturated* reader pool proved it.

A worker posts its reply and then **immediately takes the next queued message off its queue and re-enters native code**. Terminating on reply-receipt races that next statement's native call and aborts exactly as before. Measured: with reply-triggered termination the repaired suite still panicked at `array.rs:22` in **6 of 8 runs**.

The fix is a `__drained__` probe queued *behind* everything the worker still holds. The worker runs messages strictly FIFO and each statement to completion, so acknowledging a probe queued after them is the worker's own proof that it holds no native frame and has none left to start. `close()`'s `awaitWorkerIdle` had the identical flaw — it resolved on any message — and now uses the same probe.

Two further holes, both reachable only under saturation:

- A second abort on an **already-retired** connection re-entered `abandon()` with an empty pending map, so `hadInFlightAtRetire` was false and the immediate-terminate branch fired on a worker still inside its native call — the original bug, reintroduced on the second abort.
- A retired connection could still **spawn a fresh worker** if a request reached it between selection and retirement; its replies would be discarded as orphan chatter and its callers would hang. `request()` now rejects with an internal `WorkerDbRetiredError` and `routeRead` re-dispatches to a live reader.

## The test was lightened, and that made it stricter

The earlier shape (4 readers, 8 concurrent abandonments, 120 iterations) was tuned on a 16-core machine and **killed the 2-core CI runner outright** — "the runner has received a shutdown signal", twice, ~70s in, with no test output at all. Not a flake; the workload was simply too heavy for a shared runner.

Now 2 readers, 3 concurrent abandonments, a smaller result set, 200 iterations. Two readers **saturate** the pool — and saturation is exactly what exposed every defect above. The cheaper test is the stricter one.

## Re-verified

- **6/6** runs reproduce the panic on unpatched main at `63a7dcf88`
- **8/8** runs green with this change, in ~35-50s
- Worker unit suite **13/13**

markmhendrickson:
## CI cancellation diagnosed — and it surfaced a residual hole in the fix

Investigated the two `contract_parity` cancellations (jobs `102023636280`, `102027507122`). Two separate findings; the second blocks merge.

### 1. Why the step was cancelled: the runner ran out of memory

Both attempts die ~70s after the last log line, on different runners, with no vitest output:

```
attempt 1  10:20:21  The runner has received a shutdown signal…
attempt 2  10:34:01  The operation was canceled.
```

Ruled out by evidence, not plausibility:
- **Not a job/step timeout** — there is no `timeout-minutes` anywhere in `ci_test_lanes.yml`, and the test's own budget is 420s.
- **Not infrastructure** — every other job in both attempts passed; only `contract_parity` died.
- **Not the abort/core-dump handling** — the child aborts are caught and asserted on; they do not signal the runner.

It is memory. Sampling total RSS of the child server processes through a local run of this suite:

```
13:03:09  total 1,949 MB   (1 proc)
13:03:42  total 19,729 MB  (2 procs, biggest 16,966 MB)
13:04:31  total 20,041 MB
```

A GitHub-hosted `ubuntu-latest` runner has 16 GB. The spikes exceed it, the kernel OOM-kills the runner agent, and Actions reports that as *shutdown signal* / *cancelled* rather than a test failure — exactly the observed symptom.

The source is result-set size, not thread count. `SLOW_ALL` selects `a.v` (a 192-char hex column) `LIMIT 500000`, and the worker does `.all(…).map(stripMetadata)`, holding the array twice:

```
a.v   LIMIT 500000:  170 MB retained per query
a.id  LIMIT 500000:   35 MB
a.id  LIMIT 150000:   19 MB
```

Selecting `a.id LIMIT 150000` instead drops peak child RSS from ~20 GB to **461 MB** while keeping ~1,500 native crossings (`rowsNext` fires per 100 rows) and a 65ms query — still far longer than the 3–25ms jittered abort lead, so the race window is preserved. **Verified it still reproduces on unpatched main at `63a7dcf88`: 4/4 runs panic** (`external.rs:80` ×2, `array.rs:22` ×2). The gate survives the shrink.

### 2. The blocker: the fix does not hold when a reader is abandoned with 2+ queued statements

Once the suite fit in memory and actually exercised the abandonment path, **the panic reproduced on this branch** — 4/4 runs at `neon-1.0.0/src/sys/array.rs:22`.

The correlation across every run I captured:

| run | saturation | `reclaimed 2 in-flight` | panics |
|---|---|---|---|
| `a.id` 500k (green) | 0 | 0 | 0 |
| `a.id` 150k | 33 | 102 | 5 |
| `a.v` 500k | 53 | 204 | 5 |

The green runs never logged a single reclamation — they were green because the race never fired, not because it was contained.

The mechanism, in `src/repositories/worker/worker_file_database.ts`:

- `failInFlight` (line 382) records `this.hadInFlightAtRetire = entries.length > 0` — a **boolean**. The count is discarded.
- The `message` handler (line 332) treats the **first** reply from a retired worker as proof it is back in JavaScript, and calls `finish()` → `worker.terminate()`.

But the worker executes its queue **strictly FIFO and synchronously**. A reader abandoned holding 2 statements posts a reply for statement 1 while it is already inside statement 2's native call. Terminating there is precisely the #2324 abort the fix exists to prevent.

The test hits this squarely: 8 concurrent abandonments over a 4-reader pool is exactly 2 statements per reader, which is why the logs show `reclaimed 2 in-flight` and never `1`.

The comment above that branch states the premise that does not hold:

> A reply from a RETIRED worker means the orphan has finished its native call

True for a one-statement orphan; false for the queued case.

### Where this leaves the PR

Per the brief I am not reworking the approach unilaterally — this is the "diagnosis shows the fix itself is incomplete" case, so I am reporting rather than redesigning. **Not merged, not deployed.**

The shape of the remedy looks small — await one reply *per outstanding statement* (keep the count instead of the boolean) before terminating, with the existing grace window unchanged as the backstop — but that is a change to the fix's core invariant and should be the author's call. `terminate()`/`awaitWorkerIdle` resolves on the first `message` too and likely needs the same treatment.

Worth noting the CI shrink and this defect are independent: the memory fix is needed regardless, and it is what made the residual hole visible instead of letting it hide behind an OOM.

*Evidence: unpatched main `63a7dcf88` 4/4 panic; this branch `c2be381b9` 4/4 panic once the abandonment path is actually exercised; 8/9 "green" runs logged zero reclamations.*

markmhendrickson:
Two corrections to the numbers above, now that the remaining runs finished. Both make the picture more precise; neither changes the conclusion.

**The branch panic rate is 4 of 6, not 4 of 4.** Runs 5 and 6 came back green. That is the expected shape of a race rather than a reprieve — it is 67% on code whose entire purpose is to make this unreachable, and the four failures are all `array.rs:22` with the same `reclaimed 2 in-flight` signature.

**The 8-reader experiment is inconclusive, not confirmatory.** I ran the suite with `readerWorkers: 8` against the same 8 concurrent abandonments (1 statement per reader instead of 2), expecting green. It was green 3/3 — but all three logged **zero** reclamations, so the abandonment path never ran and the result says nothing either way. I am flagging it rather than counting it as support for the diagnosis.

The diagnosis rests on the direct evidence instead: the `entries.length > 0` → boolean collapse at line 382, the terminate-on-first-reply at line 332, the worker's FIFO synchronous execution, and the run-level correlation where every panicking run logged `reclaimed 2 in-flight` and every green one logged no reclamations at all.

markmhendrickson:
**`contract_parity` is green and the survival step succeeded rather than being cancelled — the first clean run of the test that reproduces this defect.** Before merging, one question decides whether that green means what we want it to.

**The two OOM fixes are not equivalent.** The head carries `SLOW_ALL … LIMIT 120000` (down from 500000). The diagnosis in [this comment](https://github.com/markmhendrickson/neotoma/pull/2338#issuecomment-5584211458) reached the same memory ceiling a different way: select `a.id` instead of `a.v` at `LIMIT 150000` — peak child RSS 20 GB → 461 MB, **while keeping ~1,500 native crossings and a 65ms query**, still far above the 3–25ms abort lead. It verified that variant **still panics 4/4 on unpatched `main` at `63a7dcf88`**.

Row-count reduction and column-width reduction both cut memory. Only one of them is known to preserve the window the reproduction depends on.

**So: has the 120k variant been shown to fail on unpatched main?** If yes, this is ready and the question is closed. If not, a green run is consistent with two very different states — the fix working, or the test no longer reaching the path. #2317 shipped on the second of those and did not hold, which is the whole reason this suite exists.

**The other finding may already be answered by the head, and I would like that confirmed rather than assumed.** The diagnosis (against `c2be381b9`) found `hadInFlightAtRetire` collapsing a count to a boolean, so a reader abandoned holding two statements posts reply 1 while inside statement 2's native call — with a clean correlation: every panicking run logged `reclaimed 2 in-flight`, every green run logged **zero reclamations**.

Reading `801b53a32`, the `__drained__` probe appears to supersede that: the boolean now only decides *whether to send the probe*, and the probe queued behind everything FIFO is what determines when terminating is safe — so the count no longer needs to be exact. `awaitWorkerIdle` uses the same probe, which closes the second hole that diagnosis predicted. **If that reading is right, say so on the PR and the finding is discharged.**

**And the correlation is the check worth keeping either way:** a green run that logs zero reclamations has not exercised abandonment. Whatever the final row count, the run that gates this merge should show reclamations *and* no panic.

ateles-agent (approved):
Approving on the strength of the reproduction, not the argument.

The mechanism is demonstrated, not asserted: 6/6 runs reproduce the SIGABRT on unpatched main at 63a7dcf88 (hitting both `external.rs:80` — the exact assertion in the production trace — and the sibling `array.rs:22`), and 8/8 runs green with this change. CI is green across all 11 jobs, and the survival step passed in 87s with the driver assertion active, so libsql genuinely resolved on the runner rather than falling back to better-sqlite3.

Notably this PR corrects its own first commit: the spec's prescribed 'terminate from the message handler' was insufficient, because a worker posts a reply and then immediately dequeues the next message and re-enters native code. That was caught by running against a saturated pool and is now handled with a FIFO drain probe. The thread cost is reported honestly (peak 5 -> 15, draining to 1) rather than glossed.

Issue #2340, opened 2026-09-08
Redesign split_entity to be append-only: the reducer's contract, not a gated exception (operator ruling supersedes #2329's split disposition)


markmhendrickson:
**Design issue, not a bug report.** The bug — that `split_entity` mutates observation rows in place — is already recorded on #2329. This issue exists because the operator has **overruled** that issue's disposition for split.

#2329's spec offered two paths for split: gate it as a named exception to redline R1 (the spec says plainly that a gated split *is not R1 compliance*), or disable the tool pending redesign. It recommended disabling, and flagged the choice as the operator's.

The operator chose **neither**: redesign `split_entity` so that it *is* append-only. The standard applied is that **all tools should work on an append-only principle**. The spec treated redesign as out of reach because it requires changing what the reducer is a function of — but that is a design change, not an impossibility, and it is what this issue is for.

The job here is to state what an append-shaped split would have to do, and what in the reducer would have to change to permit it.

Everything below was verified against `origin/main` at `63a7dcf88`.

---

## 1. What split does today

`src/services/entity_split.ts` selects a subset of the source entity's observations by a declarative predicate, then re-points them with a single in-place `UPDATE`:

```ts
// src/services/entity_split.ts:306-314
const matchedIds = matched.map((o: ObservationRow) => o.id);
const { error: updErr } = await db
  .from("observations")
  .update({ entity_id: newEntityId })
  .in("id", matchedIds)
  .eq("user_id", userId);
if (updErr) {
  throw new Error(`Failed to re-point observations: ${updErr.message}`);
}
```

Surrounding mechanics, for the implementer:

- Predicate forms are `observed_at_gte`, `source_id_in`, and `observation_field_equals` (`:30-53`); matching is done in-process (`:264`), not in SQL.
- Guards reject a predicate matching zero rows (`:266-270`) or every row (`:271-276`) — the latter on the grounds that an all-match split is a rename.
- The target entity row is inserted first when not splitting into an existing entity (`:283-301`).
- Snapshots are invalidated and recomputed on both sides afterwards (`:328-330`).
- The audit row is written to `entity_splits` at **`:335-346`**, carrying `source_entity_id`, `new_entity_id`, the canonicalized `predicate`, `observations_rewritten`, and `idempotency_key`.
[...]

Issue #2329, opened 2026-09-08
Append-only on observations is a convention, not a constraint: merge, split, and rollback mutate rows in place


markmhendrickson:
## Summary

Append-only on `observations` is a convention held by discipline, not a constraint held by the database. There are no triggers, no `IMMUTABLE` declaration, and no `BEFORE UPDATE` guard — `grep -in "TRIGGER|IMMUTABLE|BEFORE UPDATE" src/repositories/sqlite/sqlite_client.ts` returns nothing, and the observations DDL (`src/repositories/sqlite/sqlite_client.ts:60-77`) is a plain table.

Four paths mutate observation rows in place. **Only one of them is erasure**, and that one is currently unreachable.

| Path | Evidence | Reachable |
|---|---|---|
| `UPDATE observations SET entity_id` (merge) | `src/services/entity_merge.ts:84-86` | **live** — `merge_entities` MCP tool |
| `.update({entity_id})` (split) | `src/services/entity_split.ts:307-308` | **live** — `split_entity` MCP tool |
| `.delete().in("id",…)` (rollback) | `src/services/schema_lag_repair.ts:364-367` | **live** — CLI `db --rollback <run_id>` |
| `cryptographicErasure` / `physicalDeletion` | `src/services/gdpr_deletion.ts:272-275`, `:294-328` | unreachable — `deletion_monitor` has zero importers |

## Why the first three are the interesting ones

`merge_entities` and `split_entity` are ordinary MCP tools. Any admitted agent can call them, and both rewrite `observations.entity_id` on existing rows. Each writes an audit row to a side table (`entity_merges` at `entity_merge.ts:207-217`; `entity_splits` at `entity_split.ts:334`), so the operation is reversible *in principle* — but the observation row's own history is destroyed, and reconstruction depends on a second table staying consistent with the first.

**A stale comment asserts the opposite.** `src/repositories/sqlite/sqlite_client.ts:171` says merges can be "reversed without any content rewrite on observations". `entity_merge.ts:84` performs exactly that rewrite. One of the two is wrong, and it is the comment.

Neither operation *needs* in-place mutation. A merge could append a new observation reassigning the entity and let the reducer resolve it — which is what the append-only model exists to do. The rewrite is an implementation convenience, not a semantic requirement.

[...]

PR #2342, opened 2026-09-08
feat(snapshots): resolve attachment through a declared layer (#2340)

markmhendrickson:
Implements the attachment-resolution contract approved on #2340 ([analysis](https://github.com/markmhendrickson/neotoma/issues/2340#issuecomment-5583845458)):

> A snapshot is a function of the observations **attached to** an entity, where attachment is resolved through a declared resolution layer, rather than of the observations whose `entity_id` column equals the entity.

**Scope: builds the seam and moves the fetch onto it. It does not redesign `split_entity` or `merge_entities`** — that is the follow-on work this unblocks (#2329), and keeping it separate is what makes this reviewable.

## What changed

| | |
|---|---|
| `src/services/attachment_resolution.ts` | **new** — the declared resolution layer |
| `src/services/snapshot_computation.ts` | +14 −7 in `recomputeSnapshot` |
| `tests/integration/attachment_resolution_equivalence.test.ts` | **new** — the safety argument |
| `src/reducers/observation_reducer.ts` | **unchanged**, as predicted — it reduces whatever set it is handed |

### The resolver

One resolution rule is live today: the entity-grained merge alias (`entities.merged_to_entity_id`), followed **transitively**. That is behaviour-preserving, since merge already rewrites `observations.entity_id` to the survivor — so for any entity that is not itself a tombstone the resolved set is byte-identical to the old flat fetch.

The observation-grained rules (split's difference-on-source and union-on-target; merge's append-based redirect) are deliberately **not** here. They arrive with the redesigns, as a new rule in one place rather than a new fetch in a dozen.

**Cycle guard**: a visited set. **Depth bound**: `MAX_ATTACHMENT_RESOLUTION_DEPTH = 32`. Exceeding either sets `truncated` + `truncationReason` (`"cycle"` | `"depth"`) and logs — never a silent truncation. This is load-bearing rather than decorative: `entity_merge.ts:68-74` blocks merging an already-merged entity on either side, which prevents a 2-cycle but **not** chains (A→B, then C→B, then B→D builds depth).

The read path **degrades rather than throws** on a guard hit. A snapshot read that throws takes down every `recomputeSnapshot` caller; one that reports the anomaly leaves the store readable while the inconsistency is repaired.

### The one non-obvious line in `recomputeSnapshot`

A redirected id owns no snapshot of its own:

```ts
if (attached.resolvedEntityId !== entityId) return null;
```

Without this, recomputing a merge tombstone would compute the *survivor's* snapshot and upsert it **under the tombstone's id** — manufacturing a duplicate snapshot that the flat fetch never produced (it returned nothing, because the rows had already moved). The equivalence test forced this out; case `(c2)` pins it.

## The equivalence test — before and after

This is the whole safety argument, so it was written first and run against unmodified `origin/main` (`63a7dcf88`) **before** the resolver existed.

| | equivalence (a–d) | guards |
|---|---|---|
| **Before** (flat fetch, `origin/main`) | **5 / 5 pass** | n/a — resolver did not exist |
| **After** (resolver) | **5 / 5 pass, assertions unchanged** | **4 / 4 pass** |

The equivalence assertions are byte-identical across both runs. A diff in them accompanying this change would have meant the migration altered a snapshot, which is a regression rather than a migration.

Covered: plain entity; split-then-recompute yielding the same *two* snapshots; merge-then-recompute yielding the same *one*; a tombstone still recomputing to `null`; user scoping preserved. Plus the three guarantees with no pre-change counterpart: a chain resolves **fully** (not one hop), a cycle **terminates**, depth is **bounded and reported**.

The test uses a dedicated user-scoped entity type rather than borrowing a built-in one. The reducer projects a snapshot through the *active* schema, so borrowing a shared type couples the assertions to whatever another test file happened to register for it — which it did, and it failed under full-suite parallelism before this was fixed.

## Verification against the analysis

Verified against `origin/main` at `63a7dcf88` — the same SHA the analysis used.

**Held exactly**: the flat fetch at `snapshot_computation.ts:59-63`; `entity_id` as a hashed input to `generateObservationId`; the `interpretation.ts` per-entity loop sharing one `sourceId`/`interpretationId` (so dropping `entityId` from the hash **would** collide — option 2 stays rejected on evidence); split's in-place `UPDATE`; the stale "no transaction primitive" comment at `entity_split.ts:303-305`, contradicted by `entity_merge.ts:81`; the reducer needing no change; and the identification of exactly **three** pointer-following sites, two of them unguarded.

**Drifted** — see the report on #2340 for detail. In short: the `merged_to_entity_id` file counts are lower than stated (39/23, not 44/28 — cosmetic); and the `recomputeSnapshot` caller list is **wrong in both directions** — `cli/index.ts`, `server.ts`, and `batch_correction.ts` do **not** call it, while ~7 *other* sites fetch observations flat and feed `observationReducer.computeSnapshot` directly, bypassing this seam entirely. `recomputeSnapshot` is the right first seam, but it is not the only one; the others are enumerated on the issue as follow-on scope.

**No `recomputeSnapshot` caller needed changing**, as predicted — all five call sites are identical and untouched.

## Deliberately not done

Per the approved analysis: no `redlines.md` amendment (this **strengthens** R1, so the amendment process is not engaged); no `philosophy.md` §5.1 edit (it was *proposed*, not approved — flagged on the issue instead of made); no repair of already-split data (that would adopt the rejected option 3 on historical data); no split-transaction fix (belongs with the split redesign); and `split_entity` stays **enabled and unchanged**.

## CI

`contract_parity` is **green on `main` at `63a7dcf88`**, so a red one here would be mine. Locally: type-check clean, lint 0 errors, format clean, integration **147 files / 949 tests green**, unit **311 files / 3492 tests green**.

Refs #2340, #2329

### DIFF
--- a/src/services/attachment_resolution.ts
+++ b/src/services/attachment_resolution.ts
@@ -1,0 +1,181 @@
+/**
+ * Attachment Resolution Service (Domain Layer) — #2340
+ *
+ * Implements the approved contract:
+ *
+ *   A snapshot is a function of the observations *attached to* an entity,
+ *   where attachment is resolved through a declared resolution layer, rather
+ *   than of the observations whose `entity_id` column equals the entity.
+ *
+ * This module IS that declared resolution layer. It is the single seam every
+ * snapshot fetch should go through, so that merge's union and split's
+ * difference become two resolution rules over one mechanism rather than two
+ * traversals with two cycle guards.
+ *
+ * ## What it resolves today
+ *
+ * Exactly one rule is live: the entity-grained merge alias
+ * (`entities.merged_to_entity_id`). Asking for an entity that was merged away
+ * resolves to its survivor, transitively. That is behaviour-preserving —
+ * merge already rewrites `observations.entity_id` to the survivor, so the
+ * resolved set is identical to today's flat fetch for every entity that is
+ * not itself a tombstone. See `tests/integration/attachment_resolution_
+ * equivalence.test.ts`, which pins that equivalence and passed unchanged
+ * before and after this module landed.
+ *
+ * The observation-grained rules — split's difference-on-source and
+ * union-on-target, and merge's append-based redirect — are deliberately NOT
+ * here yet. They arrive with the `split_entity` / `merge_entities` redesigns
+ * (#2329), which this module unblocks. The seam is built first, on purpose,
+ * so those changes are a new resolution rule in one place rather than a new
+ * fetch in a dozen.
+ *
+ * ## Guards
+ *
+ * Resolution is bounded by a cycle guard (a visited set) and a maximum depth.
+ * Exceeding either is reported through `truncated` / `truncationReason`,
+ * never silently truncated. This is deliberate: the two existing pointer-
+ * follows in the codebase have neither guard —
+ * `entity_queries.ts:getEntityWithProvenance` recurses into itself with no
+ * visited set, and `cli/index.ts` resolves exactly one hop — so a declared
+ * resolution layer that inherited those defects on day one would be no
+ * improvement. `entity_merge.ts` blocks merging an already-merged entity on
+ * either side, which prevents a two-cycle but does not prevent chains
+ * (A→B, then C→B, then B→D builds depth), so the bound is load-bearing.
+ *
+ * The read path degrades rather than throws: a cycle or an over-deep chain
+ * returns the best-resolved set with `truncated: true` and logs, because a
+ * snapshot read that throws takes down every caller of `recomputeSnapshot`,
+ * whereas one that reports the anomaly leaves the store readable while the
+ * inconsistency is repaired.
+ */
+
+import { db } from "../db.js";
+import { logger } from "../utils/logger.js";
+import type { Observation } from "../reducers/observation_reducer.js";
+
+/**
+ * Maximum number of alias hops followed before resolution is declared
+ * over-deep. Merge chains are pathological beyond a couple of hops; 32 is far
+ * above anything legitimate while still bounding a runaway.
+ */
+export const MAX_ATTACHMENT_RESOLUTION_DEPTH = 32;
+
+export type AttachmentTruncationReason = "cycle" | "depth";
+
+export interface AttachmentResolution {
+  /**
+   * The entity the requested id resolves to after following the declared
+   * resolution rules. Equals the requested id when nothing redirects.
+   */
+  resolvedEntityId: string;
+  /** The observations attached to the resolved entity, for this user. */
+  observations: Observation[];
+  /** The full hop path, starting at the requested id. Useful for diagnosis. */
+  path: string[];
+  /** True when a guard stopped resolution before it reached a fixed point. */
+  truncated: boolean;
+  truncationReason?: AttachmentTruncationReason;
+}
+
+/**
+ * Follow the entity-grained merge alias to a fixed point, bounded by a
+ * visited set and a depth cap. Pure traversal — no observation reads.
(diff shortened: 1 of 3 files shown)


markmhendrickson:
Follow-on work found during this change, filed rather than folded in to keep this diff reviewable:

- **#2343** — snapshot computation bypasses this seam in ~10 other places (`actions.ts`, `server.ts` ×2, `cli/index.ts`, `interpretation.ts`, `schema_lag_repair.ts` ×2, `schema_registry.ts`, `local_db_adapter.ts`, `entity_snapshot_at_time.ts`). Harmless today; **must land before either the split or merge redesign ships**, or those redesigns silently persist snapshots computed from the wrong observation set.
- **#2344** — the three existing merge-alias pointer-follows have no cycle guard and no depth bound. Independent bug, fixable in place without adopting the resolver. The argument for keeping it out of this PR is on #2340.

Full verification report — what held, what drifted, and the one consequence the analysis did not anticipate — is at https://github.com/markmhendrickson/neotoma/issues/2340#issuecomment-5584345144

neotoma-agent:
**BLOCKED**

| Gate | State | Owner |
|---|---|---|
| pm | pending | Pavo |
| ux | pending | Accipiter |
| arch | pending | Waxwing |
| impl | pending | — |
| pr_review | pending | — |
| qa | pending | — |
| legal | not_required | — |

- **Parent issue:** #2340 (`ent_35a05cabfb296aff7a862b42`)
- **Owner:** current `pavo` (Phase 1 PM) → next Accipiter + Waxwing (after PM)
- **Labels:** none applied this turn (PR event)
- **GATE_INHERITANCE:** blocked — pre-impl gates unsigned on authoritative Neotoma `gate_status`

### Unsigned pre-impl gates

- [BLOCKING] pm: owned by **Pavo** — still `pending`
- [BLOCKING] ux: owned by **Accipiter** — still `pending`
- [BLOCKING] arch: owned by **Waxwing** — still `pending`

### Resolution

- Gate owners: sign off via Neotoma (`gate_status.<gate>` → `signed_off`), or operator waives.
- Operator override (only @markmhendrickson): `/confirm-gates-clear` — waives all unsigned pre-impl gates and re-triggers the PR pipeline.

Vanellus not assigned as reviewer while inheritance is blocked.

---
📎 Neotoma: [neotoma#2340](https://neotoma.markmhendrickson.com/entities/ent_35a05cabfb296aff7a862b42)

ateles-agent (approved):
Approved. All ten CI lanes green.

The equivalence test is what makes this safe and it was done properly: 5/5 passing on `origin/main` before the change with the *final* assertions, re-run after reverting only the src changes so both runs test identical code. No snapshot moved.

The consequence the analysis did not anticipate is the most valuable thing here — `if (attached.resolvedEntityId !== entityId) return null;`. Without it, recomputing a merge tombstone writes the survivor's snapshot under the tombstone's id, a duplicate the flat fetch never produced. The general form is worth carrying into both redesigns: **resolving attachment and owning a snapshot are different questions**, and conflating them is how a resolution layer silently duplicates state.

Excluding the three pointer-follows is the right call and the argument holds: they answer "which entity should I read?", the resolver answers "which observations attach here?", and those coincide only because merge moves the rows — once split's observation-grained rules land they diverge. Migrating them now would bind them to a contract about to widen. #2344's guards-only PR is the right shape for the real bugs.

Degrading rather than throwing on a cycle or depth breach is correct — a throwing snapshot read takes down every caller — and `truncated` plus `truncationReason` means it is never silent.

Issue #2343, opened 2026-09-08
Snapshot computation bypasses the attachment-resolution seam in ~10 places


markmhendrickson:
Found while implementing the attachment-resolution contract (#2340, PR #2342). Verified against `origin/main` at `63a7dcf88`.

#2342 moves `recomputeSnapshot` onto the declared resolution layer (`src/services/attachment_resolution.ts`), so a snapshot becomes a function of the observations **attached to** an entity rather than of the rows whose `entity_id` column equals it.

**But `recomputeSnapshot` is not the only path that computes a snapshot.** Roughly ten sites fetch observations by flat `entity_id` and hand them straight to `observationReducer.computeSnapshot`, bypassing that seam entirely:

| Site | Context |
|---|---|
| `src/actions.ts:11780` | stale-snapshot auto-fix |
| `src/server.ts:1010` | stale-snapshot repair |
| `src/server.ts:6238` | the MCP `store` path |
| `src/cli/index.ts:3193` | CLI snapshot recompute |
| `src/services/interpretation.ts:777` | post-interpretation snapshot |
| `src/services/schema_lag_repair.ts:275` | schema-lag repair |
| `src/services/schema_lag_repair.ts:381` | schema-lag repair |
| `src/services/schema_registry.ts:1750` | post-migration recompute |
| `src/repositories/sqlite/local_db_adapter.ts:546` | `recomputeEntitySnapshot` — a second implementation of the same idea |
| `src/services/entity_snapshot_at_time.ts:166` | resolves one merge hop, then fetches flat |

`src/server.ts:6299` already documents its own case: *"This path computes and upserts the snapshot inline instead of going through `recomputeSnapshot()`."*

## Why this matters

Today these are mostly harmless — merge rewrites `observations.entity_id` to the survivor, so a flat fetch and a resolved fetch return the same set for any entity that is not itself a tombstone.

**They stop being harmless the moment `split_entity` or `merge_entities` becomes append-based** (#2329). Once attachment is declared by appended records rather than by the column, every one of these sites computes a snapshot from the *wrong observation set* — and several of them **upsert** that snapshot, so the wrong answer is persisted rather than merely returned.

`local_db_adapter.ts:546` deserves specific attention: it is a parallel implementation of `recomputeSnapshot` living in the repository layer, so it will drift independently unless it is folded in.

## Suggested resolution

[...]

Issue #2334, opened 2026-09-08
The MCP transport misses enforcement seams the REST path gets, and one test asserts the opposite


markmhendrickson:
## Summary

Two defects found today share one diagnosis, and the pattern is worth its own issue because it will keep producing them.

- **#2326** — `installSubscriptionBridge` is called only from the HTTP startup path, so substrate events are silently dropped under MCP stdio.
- **#2327** — `enforceAttributionPolicy("observations", …)` is never called on the MCP `store` path, so a configured `{"observations":"reject"}` policy does nothing.

In both cases a declared seam exists and **the MCP path is the one that misses it**. REST's `storeStructuredForApi` routes through `createObservation` (`actions.ts:7850`) and gets the attribution gate; MCP's `storeStructuredInternal` (`server.ts:5298`) pre-computes the observation id and inserts directly at `:6114`, bypassing the service layer that carries the gate. Same tool name, same operator configuration, different enforcement.

The cause is structural rather than incidental: **the enforcement conventions were written for the service layer, and the MCP core deliberately bypasses it** — for reasons that are individually defensible (id pre-computation for the dedup probe, avoiding a service-layer import graph). Each bypass is a local decision; the aggregate is a transport where gates are absent by default rather than present by default.

## Why this is not covered by fixing the two instances

A third instance is already visible and unaudited: `mcp_ws_bridge.ts` and `mcp_dev_shim.ts` are both `package.json` entrypoints and likely carry the same bridge gap as #2326.

A fourth is inert but real: `enforceOverridePolicy` (`observation_storage.ts:92-100`) is absent from the MCP path. It cannot be exploited today because no `agent_definition` schema is seeded — but it will bite when that schema lands, and the reason will not be written down anywhere by then.

## The test that should have caught #2327, and structurally cannot

`tests/integration/anonymous_write_policy.test.ts:5-11` asserts, in a comment, that coverage holds *"regardless of transport … HTTP / MCP stdio / MCP HTTP / CLI backup all reach the same code path"*.

It then calls `createObservation` **directly** at `:73`. It never dispatches an MCP store. So it asserts the very property that is false, by testing the one path where it is true.

That comment is plausibly why nobody looked. A test whose header claims transport-independence and whose body exercises one transport is worse than no test — it answers the question before anyone asks it.

## What would close this class

Two candidate shapes, both proposed inside the individual specs and neither yet chosen:

[...]

Issue #2325, opened 2026-09-08
store() file_path resolves server-side, so no remote client can attach a file; inlining is the only route


markmhendrickson:
## Summary

`store()`'s `file_path` parameter resolves paths **on the server**, so it cannot work at all from a remote client against a hosted instance. A client passing a real, readable local path gets:

```
MCP error -32603: File not found: /Users/<operator>/Documents/data/.../review.txt
```

The file exists and is readable — on the client. The hosted instance has no access to the caller's filesystem, and there is no client-side read step.

Since the 2026-08-04 move to a hosted instance, **every** caller is a remote client. So `file_path` is a parameter that no current caller can use successfully, and its failure mode gives no hint that the path was meant to be server-local.

Reproduced on prod, MCP-over-HTTP surface, 2026-09-08, storing a 36 KB text artifact alongside an `analysis` entity.

## Relationship to existing issues

Adjacent to **#2300** (`correct()` accepts `file_path`, reports success, writes nothing) and **#2264** (a literal `@/path` in `value` stored verbatim). Those are about a path argument being silently mishandled. **This one is about there being no remote route for file content at all** — `store()` errors correctly and honestly; the gap is that the only working alternative is to inline the bytes.

Also related to **#2191** (guest-token sharing for `file_assets`) and **#2244** (original artifacts not stored), both of which assume artifacts can get into the instance in the first place.

## Current workaround, and why it is not sufficient

Passing `file_content` inline works and is what I used. It has real limits:

- The artifact rides the JSON-RPC envelope, so size is bounded by request limits rather than by anything about the file.
- Binary artifacts need base64, inflating them ~33% inside a JSON string.
- The caller must read the whole file into memory and hold it there through the request.
- `original_filename` and `mime_type` still have to be supplied by hand, so the metadata a real upload would carry is reconstructed rather than derived.

For a 36 KB text file this is fine. For the artifacts #2244 describes — bank statement PDFs, lab result scans, transcript audio — it is the wrong shape.

## What would close it

A remote-capable path for file content, e.g. one of:

1. **A multipart upload endpoint** (`POST /files` or similar) returning a handle the caller then references from `store()` — the conventional shape, and it keeps large bodies out of the JSON-RPC envelope.
2. **A presigned-upload handshake** — the instance issues a URL, the client PUTs to it, then references the resulting id.
[...]

Issue #1840, opened 2026-06-28
store response returns empty entity_snapshot_after on deduplicated observation replay


stackhouse:
## Environment
- Version: 0.18.2
- Transport: offline (SQLite)
- OS: macOS

## Summary

When an identical observation is stored twice (same entity, same field values, same priority), the second store correctly deduplicates the observation (observation count stays at 1). However, the store response returns `entity_snapshot_after: {}` (empty object) instead of the current snapshot.

Querying the snapshot directly via `getEntitySnapshot` returns the correct, populated snapshot. The issue is response-only — no data corruption.

## Steps to Reproduce

```bash
# 1. Register schema
neotoma request --offline --operation "registerSchema" --body '{
  "entity_type": "dedup_test",
  "schema_definition": {
    "fields": { "value": {"type": "number"}, "label": {"type": "string", "required": true} }
  },
  "activate": true
}' --json

# 2. Store an observation
neotoma store --entities '[{"entity_type": "dedup_test", "name": "test", "label": "t", "value": 42}]' \
  --source-priority 5 --observation-source sensor --json
# → entity_snapshot_after: {"value": 42, "label": "t"} ✓

# 3. Store the exact same observation again
neotoma store --entities '[{"entity_type": "dedup_test", "name": "test", "label": "t", "value": 42}]' \
  --source-priority 5 --observation-source sensor --json
# → entity_snapshot_after: {} ← EMPTY (but actual snapshot is correct)
```

## Expected Behavior

`entity_snapshot_after` should return the current snapshot even when the observation is deduplicated, OR the response should include a field like `"deduplicated": true` to indicate no new observation was created.

## Actual Behavior

- `entity_snapshot_after: {}` on the deduplicated store response
- `observation_count` stays at 1 (correct dedup)
- Direct snapshot query returns the correct populated snapshot

## Impact

Low severity. Any code that validates writes by checking `entity_snapshot_after` in the store response will see a false empty-snapshot on replayed/deduplicated writes. Workaround: query the snapshot separately after writing if confirmation is needed.

markmhendrickson:
Triaged as P3 response-contract bug. The report points to a replay/dedup response assembly issue rather than snapshot corruption: direct snapshot reads are correct, but `entity_snapshot_after` can be `{}` on duplicate writes. Suggested next step: add a regression for identical observation replay and make store responses return the current snapshot plus a `replayed`/`deduplicated` marker.

PR #2347, opened 2026-09-08
fix(mcp): enforce the observations attribution gate on the MCP store path (#2327)

### LINKED ISSUE #2327: The MCP store path bypasses enforceAttributionPolicy("observations"), so a policy keyed to that path never fires
## Summary

`enforceAttributionPolicy("observations", …)` is never called on the primary MCP `store` path.

`storeStructuredInternal` (`src/server.ts:4876`) inserts observations directly at `src/server.ts:5586` rather than going through `createObservation` (`src/services/observation_storage.ts:147`), which is where the `"observations"` attribution gate lives (`src/services/observation_storage.ts:85`). Scanning `src/server.ts:4876-5700` finds no `enforceAttributionPolicy` call — only `assertCanWriteProtected` at `:5390-5393`.

The only attribution gate covering that path is inside `storeRawContent` (`src/services/raw_storage.ts:48`), reached at `src/server.ts:5077` — and it is keyed **`"sources"`**, not `"observations"`.

## Why it matters

An operator who sets a per-path policy naming the write path they most obviously mean:

```
NEOTOMA_ATTRIBUTION_POLICY_JSON={"observations":"reject"}
```

would **not** block the main `store` path. Only the `sources` key would take effect there. The policy appears configured and does nothing for the write it names.

This contradicts the module's own docstring, which asserts the seam is called "at the top of each write-path service" (`src/services/attribution_policy.ts:6-8`). It is, for six declared paths — `observations` (`observation_storage.ts:85`), `corrections` (`correction.ts:45`), `relationships` (`relationships.ts:152`, `interpretation.ts:997`), `interpretations` (`interpretation.ts:331`, `actions.ts:6204`), `timeline_events` (`timeline_events.ts:330`), `sources` (`raw_storage.ts:48`) — but the busiest caller of the observations path bypasses the service that carries the gate.

Note the provenance blob is still **stamped** (`src/server.ts:5582-5584`), so writes are attributed in the record; what does not happen is the policy being **evaluated**. This is a policy-enforcement gap, not a provenance gap.

## Related, but distinct

- **#2179** — `unverified_client` writes silently admitted: about which tier is accepted.
- **#2256** — nothing signs, so every write lands at `unverified_client`: about signing.
- **#2315** — an authenticated caller may write any entity type: about per-caller write scope.

This one is narrower and mechanical: a declared enforcement seam that one major call path does not pass through, so a policy keyed to it cannot fire.

## Suggested fix

Either route `storeStructuredInternal` through `createObservation`, or call `enforceAttributionPolicy("observations", …)` explicitly before the insert at `src/server.ts:5586`. A test asserting that a `{"observations":"reject"}` policy actually rejects an MCP `store` would keep it closed — the current gap is invisible precisely because the policy accepts by default (`DEFAULT_POLICY = { anonymous_writes: "allow" }`, `src/services/attribution_policy.ts:48-50`), so nothing fails today.

markmhendrickson:
Closes #2327. Implements the spec posted at https://github.com/markmhendrickson/neotoma/issues/2327#issuecomment-5582348834.

## The defect

`enforceAttributionPolicy("observations", …)` had **zero occurrences** in `src/server.ts`. The primary MCP `store` path (`storeStructuredInternal`) inserts observations directly rather than routing through `createObservation`, which is where that gate lives (`observation_storage.ts:85`). The only attribution gate the path evaluated was keyed `"sources"`, reached indirectly via `storeRawContent` (`raw_storage.ts:50`).

So an operator setting `NEOTOMA_ATTRIBUTION_POLICY_JSON={"observations":"reject"}` got a policy that is parsed, surfaced on `/session`, and enforced on nothing this path writes.

The parity is one-sided: REST's `storeStructuredForApi` **does** route through `createObservation` and gets the gate. Same tool name, same operator configuration, different enforcement.

Provenance *is* stamped on the MCP path (`server.ts:6128-6130`), so rows are attributable. What did not happen is the policy being evaluated. A policy-enforcement gap, not a provenance gap.

## The fix, and why it is a gate call rather than a reroute

Routing `storeStructuredInternal` through `createObservation` was considered and rejected. `createObservation` computes its own observation id with **five** arguments including `idempotency_key`; this path pre-computes with **four** (`server.ts:6083-6088`) and threads that id through its dedup probe (`:6094-6100`), the `deduplicated` flag (`:6155`, added for #1840), and the `insertedObservationIds` set (`:6138`).

Rerouting would either change content-addressed observation identity whenever an idempotency key is present — an R2 concern — or require passing the pre-computed id in, which `createObservation` has no parameter for. That is a refactor of identity semantics wearing the clothes of a one-line security fix. Convergence of the two store paths is worth doing deliberately, with id-stability tests, under its own issue.

**Placement** is before any persistence, not next to the insert. The comment already in this file states the rule: a store call is not transactional, so denial must happen before anything is persisted rather than mid-batch. Gating at the insert would reject only after `storeRawContent` had already written a source row — which is exactly what T4 pins.

Ordering is after `assertStorePolicyAllows`, not before: instance store policy governs *what* may be stored, attribution governs *who* may store it. Once per call, not per entity — `enforceAttributionPolicy` does not read entity type.

The `"sources"` gate reached via `storeRawContent` stays. Both write paths are genuinely exercised and an operator may configure them differently.

## The test, and its fail-then-pass evidence

`tests/integration/mcp_store_attribution_policy.test.ts` dispatches an **actual MCP tool call** rather than calling a service helper.

Pre-fix (`679006a3`, before the `server.ts` change):

```
FAIL  T1: rejects an MCP store when {"observations":"reject"} is configured
  AssertionError: promise resolved "{ content: [ { type: 'text', …(1) } ] }" instead of rejecting
FAIL  T4: nothing is persisted when the policy rejects
  AssertionError: promise resolved … instead of rejecting
     (payload shows a real observation row: "action":"created", "deduplicated":false)

Tests  2 failed | 2 passed (4)
```

Post-fix:

```
✓ tests/integration/mcp_store_attribution_policy.test.ts (4 tests)
Tests  4 passed (4)
```

T2 (no over-rejection with an AAuth-verified session) and T3 (default policy unchanged) pass in both runs — they guard against the fix tightening the default or over-rejecting, which it does not.

## The misleading comment, corrected

`tests/integration/anonymous_write_policy.test.ts` asserted in its docblock that coverage held *"regardless of transport … HTTP / MCP stdio / MCP HTTP / CLI backup all reach the same code path"*, and then called `createObservation` directly. That claim was false, and it is plausibly why nobody looked — a test whose header claims transport-independence and whose body exercises one transport answers the question before anyone asks it.

The docblock now states its real scope (it proves the seam is present in the *service helpers*, not that any transport reaches them) and points at the new test for per-transport coverage.

## Scope

Touches `src/server.ts` and two test files. Does not touch `attachment_resolution.ts`, `snapshot_computation.ts`, `entity_merge.ts`, `entity_split.ts`, `worker_file_database.ts`, or the file-upload surface — no collision with #2325, #2329, or #2343.

Deliberately **not** in scope, per the spec: the guest carve-out at `actions.ts:7182` and `NEOTOMA_AGENT_DEFAULT_DENY` at `agent_capabilities.ts:195`. Both are permissive *defaults*, not configured policies that do nothing. The distinction that keeps this scoped: #2327 is a lie; those are choices.

## Verification

All spec line references re-verified against `origin/main` @ `679006a3` (two commits past the spec's `63a7dcf8`; neither touched `server.ts`, `actions.ts`, or the attribution modules). Every reference held.

Part of the pattern tracked in #2334.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -5369,6 +5369,34 @@ export class NeotomaServer {
       });
     }
 
+    // Attribution policy, "observations" write path (#2327).
+    //
+    // This MCP store core does not route through `createObservation`, so the
+    // "observations" gate that lives at observation_storage.ts:85 is never
+    // reached from here. It pre-computes the observation id for its dedup
+    // probe and inserts into `observations` directly, which is deliberate —
+    // but it means a seam placed only in the shared write helper is enforced
+    // on REST and silently inert on MCP. Called explicitly, mirroring the
+    // `assertCanWriteProtected` call below and the reasoning recorded for
+    // `assertStorePolicyAllows` above.
+    //
+    // Placement, for the same reason given above: a store call is not
+    // transactional, so denial must happen before anything is persisted.
+    // Gating next to the insert would reject only after `storeRawContent`
+    // had already written a source row. Once per call, not per entity —
+    // `enforceAttributionPolicy` does not read entity type, so the answer is
+    // identical for every row in the batch.
+    //
+    // Ordering: after `assertStorePolicyAllows`, not before. Instance store
+    // policy governs *what* may be stored; attribution governs *who* may
+    // store it. Keeping this order means an entity type the instance refuses
+    // is refused identically regardless of attribution.
+    {
+      const { enforceAttributionPolicy } = await import("./services/attribution_policy.js");
+      const { getCurrentAgentIdentity } = await import("./services/request_context.js");
+      enforceAttributionPolicy("observations", getCurrentAgentIdentity());
+    }
+
     // Plan mode: resolve deterministically, report planned actions per entity,
     // and skip every write (source row, observations, snapshots, relationships).
     if (!commit) {
--- a/tests/integration/anonymous_write_policy.test.ts
+++ b/tests/integration/anonymous_write_policy.test.ts
@@ -3,12 +3,25 @@
  * the service boundary.
  *
  * The policy helper is exercised directly in unit tests; this test
- * asserts that each of the five canonical write paths actually calls
- * `enforceAttributionPolicy` and therefore reject in `reject` mode,
- * regardless of transport. It runs against the local SQLite backend
- * without going through HTTP, because the enforcement seam is
- * deliberately placed inside the service (HTTP / MCP stdio / MCP HTTP
- * / CLI backup all reach the same code path).
+ * asserts that each of the five canonical write-path *service helpers*
+ * calls `enforceAttributionPolicy` and therefore rejects in `reject`
+ * mode. It runs against the local SQLite backend without going through
+ * HTTP.
+ *
+ * SCOPE — read this before trusting the file (#2327). This test proves
+ * the seam is present in the service helpers. It does NOT prove any
+ * transport reaches those helpers. An earlier version of this docblock
+ * claimed coverage held "regardless of transport … HTTP / MCP stdio /
+ * MCP HTTP / CLI backup all reach the same code path". That claim was
+ * false and it is why nobody looked: the MCP `store` core
+ * (`storeStructuredInternal`) does not route through
+ * `createObservation`, so `{"observations":"reject"}` was silently
+ * inert on the primary MCP write path while this file passed.
+ *
+ * Per-transport enforcement is asserted where the transport is actually
+ * dispatched — see `tests/integration/mcp_store_attribution_policy.test.ts`.
+ * Adding a helper here does not extend coverage to a transport; only
+ * dispatching that transport does.
  */
 
 import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
--- a/tests/integration/mcp_store_attribution_policy.test.ts
+++ b/tests/integration/mcp_store_attribution_policy.test.ts
@@ -1,0 +1,148 @@
+/**
+ * Integration test: the MCP `store` path evaluates the "observations"
+ * attribution gate (#2327).
+ *
+ * Why this test exists separately from
+ * `tests/integration/anonymous_write_policy.test.ts`: that test asserts,
+ * in its own docblock, that enforcement holds "regardless of transport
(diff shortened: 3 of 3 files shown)


neotoma-agent:

**COMMENT**

## Triage & Gate-Status Board

**Parent issue:** #2327 — no Neotoma `issue` entity existed prior to this PR opening (checked `github_number`, `issue_number`, and `github_url`-contains lookups; zero matches). No gate-owner sign-off artifacts found on the issue thread — the sole comment is the operator's own implementation spec, not a Pavo/Accipiter/Waxwing sign-off.

Per the legacy-issue rule, this is genuine absence (no entity at all, not merely a pending field) — retroactively initialized as legacy rather than hard-blocked:

| Gate | State | Owner |
|---|---|---|
| pm | not_required (legacy) | Pavo |
| ux | not_required (bug fast-path skips ux) | Accipiter |
| arch | not_required (legacy) | Waxwing |
| impl | not_required (legacy — PR implements directly) | Cicada |
| legal | not_required | Buteo |

**Owner:** current phase → Vanellus (pr_review). Next phase → Phoenicurus + Buteo* (*legal not required) on merge.

**Labels applied:** none this pass (issue-triage label applies to new-issue events, not PR-opened events).

📎 Neotoma: [issue ent_4dfe847b3381e2177768fc0e](https://neotoma.markmhendrickson.com/entities/ent_4dfe847b3381e2177768fc0e)

---

All pre-impl gates (pm/ux/arch) are satisfied. Assigning **Vanellus** as reviewer.

GATE_INHERITANCE: clear

castor-agent (approved):
Verified against the spec at #2327. Fail-then-pass demonstrated: T1 and T4 fail pre-fix (store resolves with a real observation row instead of rejecting), all four pass post-fix. Gate placement is before any persistence, which T4 pins. All 10 CI checks green.

PR #1942, opened 2026-07-16
fix(store): guard malformed store_warnings rule (skill store 500)

markmhendrickson:
Parent #2165

## Problem

`store()` of **any** `skill` entity fails with a 500:
```
MCP error -32603: Cannot read properties of undefined (reading 'some')
```
Reproduced against prod via MCP on 2026-07-16. A minimal `{entity_type:'skill', name, content}` payload also fails, so it's not a payload issue.

## Root cause

The store-path store_warnings evaluator (in `src/server.ts` and `src/actions.ts`) calls `rule.fields.some(...)` for every declared rule. The **canonical** `store_warnings` rule shape (per `schema_registry.ts`) requires `fields: string[]`. But the `skill` schema's DB-stored rule is a legacy `condition`-shaped entry:
```json
{ "code": "MISSING_CONTENT_FIELD", "message": "skill has no content body.", "condition": { "missing_all_of": ["content"] } }
```
It has no `fields` key → `undefined.some(...)` throws → the entire store 500s. Any schema carrying a malformed rule hits the same crash.

## Fix

Three layers:

1. **Store path (`src/actions.ts`, `src/server.ts`)** — both consumers validate the outer rule collection and the complete canonical rule shape before evaluating warnings: non-null object, non-empty string field array, and string code/message. Invalid legacy metadata is skipped, valid later warnings remain available, and diagnostic labels never coerce non-string rule codes. Both advisory loops have a local exception boundary so an advisory failure cannot turn an already committed store into an error. The independent `content_field` warning continues to report `MISSING_CONTENT_FIELD`.

2. **Registration path (`src/services/schema_registry.ts`)** — `validateSchemaDefinition` now validates `store_warnings`, which it previously did not touch at all while validating six sibling rule-bearing fields. A **newly supplied** malformed rule is now **rejected** at `register()`.

3. **Carry-forward path (`updateSchemaIncremental`)** — legacy rules already stored are **normalized, not rejected**. Layer 2 alone regressed the very type this PR exists to unbreak: `updateSchemaIncremental` forwards a schema's existing `store_warnings` into `register()`, so the next incremental update to `skill` would have thrown at registration instead of at store time. `store_warnings` now gets the same pruning treatment its neighbours already had — malformed entries dropped with a log naming the entity type and rule code, removed field names pruned out, and a rule left watching nothing dropped rather than kept as one that fires vacuously on every store. A well-formed rule sitting alongside a malformed one survives.

[...]

Issue #172, opened 2026-05-16
upload --local defaults toward global npm package directory unless NEOTOMA_DATA_DIR is set


castor-agent:
## Problem
`neotoma upload --local` defaults to the global npm package install directory rather than the current working directory's data path, unless `NEOTOMA_DATA_DIR` is explicitly set in the environment.

## Impact
- Surprising: the `--local` flag implies "local to this project," but the actual default is "local to the npm install"
- Easy to write data into the wrong place without realizing
- Forces users to remember to set `NEOTOMA_DATA_DIR` for every shell session

## Expected
`upload --local` should default to the data dir associated with the current project (via project-local init, or cwd-based resolution), not the npm package install path. Environment variable should be the override, not the requirement.

## Reported by
External tester.

castor-agent:
Closing as resolved by PR #220 (merged 2026-05-18). GitHub did not auto-close at merge time despite the explicit `Closes #172` in the PR body — closing now as part of the v0.13.0 release closure-pass audit.

## What the issue asked for

`upload --local` should default to the data dir associated with the current project (via project-local init, or cwd-based resolution), not the npm package install path. Environment variable should be the override, not the requirement.

## How it was addressed

- **#172** (`neotoma upload --local` defaults to npm package dir instead of `cwd/data`): The `upload --local` action sets `NEOTOMA_DATA_DIR` to `path.join(cwd, "data")` when the variable is unset, so local ingestion lands in the project directory.

## Verification

- [x] Run `npx vitest run tests/unit/cli_bug_fixes.test.ts` — expect 13/13 passing
- [x] Manual: `cd /tmp/testproj && mkdir -p .cursor && neotoma init` — verify `.cursor/mcp.json` lands in `/tmp/testproj/.cursor/`, not a parent
- [x] Manual: run `neotoma doctor` in a project with a local `.env` — verify `data_dir_source` is `project_local` and `global_default_data_dir` is shown
- [x] Manual: set `NEOTOMA_FORCE_LOCAL_TRANSPORT=true` and run any neotoma command — verify no TCP connection attempt to `127.0.0.1`
- [x] Manual: run `neotoma upload --local <file>` from a project dir without `NEOTOMA_DATA_DIR` set — verify data lands in `<cwd>/data/`

---

_Posted as part of the v0.13.0 release closure-pass audit ([PR #257](https://github.com/markmhendrickson/neotoma/pull/257))._

PR #2388, opened 2026-09-12
fix(config): refuse the user-level data dir fallback under test (#2387)

### LINKED ISSUE #2387: CLI-invoking tests fall back to the real operator database when NEOTOMA_DATA_DIR is unset, and residue there can make a failing test pass
## What happens

Integration tests that shell out to the `neotoma` CLI resolve their database through `src/config.ts`, which falls back to `~/.config/neotoma/.env` whenever `NEOTOMA_DATA_DIR` is not already set in the child's environment. With that variable unset, the CLI resolves to **the operator's real database** — verified on this machine at `/Users/markmhendrickson/data/neotoma.db`, 204 MB, modified today. `NEOTOMA_DATA_DIR` is unset in an ordinary interactive shell.

The tests do not normally hit this, because the vitest worker inherits the variable from `vitest.global_setup.ts` and `exec` passes it to the child implicitly. So the isolation is real but **accidental**: it depends on an ambient variable that nothing in the test asserts, and any invocation path that does not inherit it writes to real data.

## Why this is worse than a stray write

Discovered while fixing CLI test failures on PR #1942. The first attempt to make the dependency explicit used `env -u NEOTOMA_DATA_DIR`, expecting the test to fail. **It passed** — because the operator's database already held residue from earlier runs of the same probes. A test that passes on leftover state in a production database is green for the wrong reason twice over: it is not testing isolation, and it is not testing the code either.

That is the precise failure `fixed_means_behavior_verified_not_contract_accepted` exists to exclude — a green result that is evidence about the runner rather than about the code. Here the runner is a live user database, so the green is evidence about accumulated residue.

## Two distinct defects

1. **The fallback destination.** A test-shaped invocation that cannot resolve an explicit data directory should fail loudly rather than silently resolving to whatever `~/.config/neotoma/.env` names. Falling back to a real database is the worst available default, because the failure is invisible — the command succeeds.
2. **The implicit inheritance.** Test isolation currently rests on an environment variable the test never states. Every CLI invocation in a test should pass its data directory explicitly and fail if it is unresolvable, so isolation is a property of the test rather than of the shell that happened to launch it.

PR #1942 fixes (2) for its own three CLI invocations. (1) is unfixed and applies to every CLI path.

## Suggested direction

- Make `src/config.ts` refuse the `~/.config/neotoma/.env` fallback when a test-mode signal is present (`NODE_ENV=test`, `VITEST`, or an explicit `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR`), exiting non-zero with the reason instead.
- Audit the remaining CLI-invoking tests for the same implicit inheritance.
- Consider whether any non-test CLI path should also refuse a silent fallback — #1504 reports `NEOTOMA_SQLITE_PATH` being advertised and ignored, and #172 reports `upload --local` defaulting toward the wrong directory unless this same variable is set, so this looks like a family rather than three separate slips.

## Not fixed here, and one thing I have not established

I have not audited which probe entities from prior runs are currently sitting in the operator's database, or whether any are still there. The residue is what made the failing test pass, so at least some existed at that moment. Worth checking before assuming the real data is untouched — but that is a data question for the operator rather than something to resolve in a test-isolation fix.

Related: #1504 (`NEOTOMA_SQLITE_PATH` advertised in CLI help but ignored by `config.ts`), #172 (`upload --local` defaults toward the global npm package directory unless `NEOTOMA_DATA_DIR` is set).

Surfaced via internal analysis on 2026-09-12; full context kept privately.

## Swarm specification

### Product / Scope (PM)

Comment posted: https://github.com/markmhendrickson/neotoma/issues/2387#issuecomment-5644559036

## Summary

**Prior art check:** Searched `gh issue list`/`gh pr list` for related work. #1504 (`NEOTOMA_SQLITE_PATH` ignored) and #172 (`upload --local` defaulting, closed) are real, correctly cross-referenced, and not duplicates. PR #1942 is confirmed still **open/unmerged**. Verified the defect directly against `src/config.ts` on this checkout: no `NODE_ENV`/`VITEST` test-mode gate exists at all — the fallback chain (lines 22–33, 108) is exactly as the issue describes.

**PM gate: SIGNED_OFF.** Scope of record:
- **In scope:** a fail-loud refusal branch in `config.ts` under a test-mode signal, plus an audit of remaining CLI-invoking tests for the same implicit-inheritance pattern.
- **Out of scope:** cleaning residue in the operator's real DB (operator's call), and extending the refusal to non-test CLI paths (a follow-up question, not this issue's requirement).
- **Acceptance criteria** require both an effect-verified test (drives the real CLI binary, revert-to-red confirmed) and cross-surface parity (MCP/HTTP startup path unaffected by the new refusal) per the standing quality gate.

**Routing decision:** Rather than the `bug` fast path straight to Cicada, I forced this through the **arch gate (Waxwing)** — the fix touches a shared, load-bearing config module and introduces a new env-var contract (`NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR`), which is closer to an interface-surface change than a localized bug, even though it doesn't strictly match the enumerated predicate. Flagged that Waxwing can waive quickly if the actual diff stays contained.

**Gate handoff completed and read-back verified:** `gate_status.pm → signed_off`, `owner_history` appended, `current_owner → waxwing`, `plan_contribution` stored (`unknown_fields_count: 0`). Verdict comment posted to the GitHub issue with the required attribution header and SIGNED_OFF chrome.

[pavo] acceptance_criteria: pm gate SIGNED_OFF on neotoma#2387. Scope: config.ts refusal branch (fail loud under test-mode signal, no silent ~/.config fallback) + audit of remaining CLI-invoking tests for implicit NEOTOMA_DATA_DIR inheritance. Acceptance criteria require effect-verified test (real CLI binary, revert-to-red) + cross-surface parity (MCP/HTTP startup unaffected). Out of scope: residue cleanup (operator's call), #1504/#172 (independent family), non-test CLI paths (follow-up). Routed current_owner → waxwing (arch), not the bug fast path, since the fix touches a shared config module and adds a new env-var contract. gate_status/owner_history/current_owner all read-back confirmed; plan_contribution ent_d24dc0dbe49e65cf2543552d stored.

### Design / UX

**User goal:** a developer running or writing a CLI-invoking test needs to know *immediately and unambiguously* that their invocation has no explicit data directory — not discover it later as silent corruption of real data.

#### Naming

- Env var: `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR` (from the issue's suggested direction) — confirmed as the right name. It reads as a requirement being *stated*, not a mode being toggled, which matches the mental model: "this invocation demands an explicit path, full stop."
- Do not introduce a second, competing flag (e.g. `NEOTOMA_TEST_MODE`) — one lever, one meaning. Test-mode auto-detection (`NODE_ENV=test`, `VITEST`) should *imply* this same refusal, not a parallel code path with its own message shape.
- Reject naming that implies severity is optional (`NEOTOMA_STRICT_DATA_DIR`, `NEOTOMA_SAFE_MODE`) — this is a correctness gate, not a strictness dial.

#### Interaction / flow

1. CLI process starts → `config.ts` begins fallback resolution.
2. Before touching `~/.config/neotoma/.env`: check if `NEOTOMA_DATA_DIR` is set. If yes → proceed normally, no behavior change.
3. If unset, check test-mode signal (`NODE_ENV=test` OR `VITEST` OR `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR` truthy).
4. If a test-mode signal is present and `NEOTOMA_DATA_DIR` is unset → **refuse immediately**, exit non-zero, print the actionable error (below) to stderr. Do not touch `~/.config/neotoma/.env` at all — not even to read it for a "would have used X" message, since reading it is the exact action being refused.
5. If no test-mode signal is present (ordinary interactive/production shell) → existing fallback behavior is unchanged. This issue does not touch the non-test path (PM: out of scope; #1504/#172 are the tracked follow-ups).

No decision point requires the user to choose anything — this is a fail-fast gate, not a prompt. The only "interaction" is reading the error and fixing the invocation.

#### Error message (agent/developer-facing, stderr, non-zero exit)

The error is read by two audiences with different context: a human debugging a failing test, and a swarm agent parsing CLI output programmatically. Both need the same three things — what happened, why, how to fix — in that order, plain text, no jargon requiring internal knowledge of `config.ts`.

```
neotoma: refusing to resolve a data directory automatically.

NEOTOMA_DATA_DIR is not set, and this process is running under a
test-mode signal (${signal_name}). Falling back to the configured
default database (~/.config/neotoma/.env) is disabled in this mode
to prevent tests from silently reading or writing real data.

Fix: set NEOTOMA_DATA_DIR explicitly before invoking this command, e.g.:
  NEOTOMA_DATA_DIR=/tmp/neotoma-test-$$ neotoma <command>

If you are not running a test and see this in error, unset VITEST /
NODE_ENV=test, or set NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR=0 to override.
```

Requirements on this message:
- **Names the signal that triggered it** (`${signal_name}` — literally which of `NODE_ENV=test` / `VITEST` / `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR` fired), not a generic "test mode detected" — so a developer with two of the three set doesn't have to guess which one to unset.
- **Gives a copy-pasteable fix line**, not just a description of the fix. This audience (senior engineer / agent) will copy-paste before reading prose.
- **Gives the escape hatch** for the false-positive case (CI runner that legitimately sets `NODE_ENV=test` for an unrelated reason but isn't invoking neotoma's own tests) — otherwise this becomes a new abandonment point for unrelated tooling.
- **Never prints the resolved fallback path it refused to use.** Printing `~/.config/neotoma/.env` or the DB path in the error is fine (it's not sensitive and helps debugging); do not print anything that looks like it partially executed the fallback before refusing.

#### Silent-failure and abandonment points

| Point | Risk | Severity | Mitigation |
|---|---|---|---|
| Test-mode signal not detected (e.g. a test runner other than vitest, no `VITEST`/`NODE_ENV` set) | Falls through to old silent-fallback behavior — the exact bug this issue fixes, just via a gap in signal detection | **P0** | Audit step (PM scope) must enumerate every CLI invocation path in the test suite and confirm at least one signal is present; do not rely on vitest-only detection |
| Developer sets `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR=0` to silence a real bug rather than a false positive | Reintroduces the original defect, now with an explicit opt-out that looks sanctioned | **P1** | Escape hatch documented as override-of-last-resort in docs, not advertised in the primary error path as a "quick fix"; keep it in the error text but as the *last* line, after the real fix |
| Error fires in CI for an unrelated reason (e.g. a script sets `NODE_ENV=test` globally for a monorepo, then shells out to `neotoma` for something unrelated to testing) | CI breaks on upgrade, developer doesn't know why | **P1** | Signal name in error message (above) makes this diagnosable in one read; changelog/release notes must call out the new refusal as a breaking behavior change |
| Non-test CLI paths (#1504, #172) are visually/behaviorally inconsistent with this new refusal — user hits silent fallback here, fail-loud there | Confusing mental model: "sometimes neotoma refuses, sometimes it doesn't" | **P2** (explicitly out of scope per PM, but flag for follow-up) | Docs must state plainly that this refusal is test-mode-only today, with a link to #1504/#172 as the tracked follow-up, so the asymmetry is explained rather than discovered |

#### Docs / examples the change must ship

1. **CLI help text** (`neotoma --help` or equivalent top-level help) gets one new line under environment variables: `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR` — one-line description, pointing to the fuller doc.
2. **A short doc section** (wherever `NEOTOMA_DATA_DIR` / `NEOTOMA_SQLITE_PATH` are currently documented — same page, adjacent section, not a new page) covering:
   - What the refusal does and when it fires (the three signals, verbatim).
   - The copy-pasteable fix example (same as in the error message — keep these two in sync verbatim so a developer who read the docs first isn't surprised by different wording in the error).
   - The override escape hatch, framed as "only if you know this is a false positive."
   - One sentence cross-linking #1504 and #172 as the known non-test-path gaps, so a reader doesn't conclude the whole CLI is now safe.
3. **A test-authoring example** in the test-writing guide (wherever contributors are told how to write a CLI-invoking test) showing the correct pattern: explicit `NEOTOMA_DATA_DIR` per test (e.g. a temp dir per test run), not relying on inherited environment. This is the doc surface that actually prevents the next instance of this bug — the error message catches it at runtime, but the example prevents the developer from needing the error at all.
4. **Changelog/release-notes entry** flagging this as a breaking change for any CI pipeline that sets `NODE_ENV=test` broadly and shells out to `neotoma` without `NEOTOMA_DATA_DIR` — per the P1 abandonment point above.

#### Accessibility

Not applicable as a visual/screen-reader concern (this is a CLI stderr message and Markdown docs, no UI component). The relevant accessibility analogue is **parseability**: keep the error message plain-text, one paragraph + one fix line + one override line, no ANSI color coding as the *only* signal of severity (exit code non-zero is the machine-readable signal; color may decorate but must not carry meaning alone), so it degrades cleanly when piped, logged, or read by an agent.

- [ ] Error message includes: what happened, which signal fired (named explicitly, not generic), a copy-pasteable fix command, and the override escape hatch — in that order.
- [ ] Error message and docs use identical wording for the fix example (no drift between the two).
- [ ] `--help` output documents `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR`.
- [ ] Docs page cross-links #1504 and #172 so the non-test-path asymmetry is explained, not silently inconsistent.
- [ ] Test-writing guide includes a worked example of per-test explicit `NEOTOMA_DATA_DIR` (e.g. temp dir pattern).
- [ ] Changelog/release notes flag the refusal as a breaking change for CI setting `NODE_ENV=test` broadly.
- [ ] No code path prints or partially uses the `~/.config/neotoma/.env` value before refusing (audit: refusal happens before any read of that file).
- [ ] Audit (PM scope) confirms every CLI-invoking test in the suite trips at least one of the three signals — not vitest-only.

markmhendrickson:
Closes #2387

## Design basis

No `docs/foundation/` design governs this diff — it is a runtime safety guard on data-directory resolution, not a change to the work model, a gate, or a principle. The contract it implements is this repo's own `docs/operations/configuration.md` § Data directory, whose stated resolution order (project-local `.env`, then the user-level config, then built-in defaults) this diff amends with one exception for test-shaped processes; that doc is updated in the same commit. `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR` is the name #2387 itself proposed.

## The mechanism, reproduced before changing anything

Verified against `origin/main` (`2fc45a45a`). The issue's reading holds:

- `hydrateDataDirFromUserEnvConfig()` returns early if `NEOTOMA_DATA_DIR` is already set; otherwise it parses the user-level config built from `$HOME` and assigns `process.env.NEOTOMA_DATA_DIR` from it, with a `catch` that swallows errors silently.
- That function is called unconditionally at module load.
- `const dataDir = process.env.NEOTOMA_DATA_DIR || join(projectRoot, "data");`

Reproduced empirically by running the **built** config module in a child process with a synthetic `HOME` whose user-level config names a stand-in directory. No real config file or database was read, written, or resolved at any point in this work.

```
[plain child, NODE_ENV unset]        exit=0  RESOLVED_DATA_DIR=<stand-in dir>
[test-shaped child, NODE_ENV=test]   exit=0  RESOLVED_DATA_DIR=<stand-in dir>
[test-shaped child, VITEST=true]     exit=0  RESOLVED_DATA_DIR=<stand-in dir>
```

A test-shaped process with no explicit data directory resolved to whatever the user-level config named, silently, exiting 0. Test isolation held only because the vitest worker inherits the variable from `vitest.global_setup.ts` and `exec` passes it to children implicitly — nothing asserted it.

## What changed

`src/config.ts` only, plus tests and the configuration doc.

- **`requiresExplicitDataDir()`** — true when `VITEST` is set, `NODE_ENV === "test"`, or `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR=1`.
- **`userEnvFallbackAllowedInTest()`** — the escape hatch, `NEOTOMA_ALLOW_USER_ENV_IN_TEST=1`.
- **`refuseUserEnvFallbackUnderTest()`** — writes an actionable message naming `NEOTOMA_DATA_DIR` and the escape hatch, then `process.exit(1)`. Not a warning, not a fallback to `projectRoot/data`.
- The refusal fires **only where the fallback would actually be taken** — test-shaped, no explicit data dir, *and* a user-level config that names one. A checkout whose user config names no data directory still reaches the `projectRoot/data` default exactly as before, which is why the non-test paths are untouched.
- The message reports the refused path's **length, not its content**: it may name a real operator directory.

`dataDir`'s shape is unchanged. `src/actions.ts` and `src/server.ts` are untouched. PR #1942's test file is untouched — it already passes its data directory explicitly.

## Tests

New `tests/unit/config_refuses_user_env_under_test.test.ts` — 7 cases. They drive the **built** config module in real child processes, because the refusal calls `process.exit` and cannot be observed in-process. Every case uses a synthetic `HOME`.

| Case | Expected |
| --- | --- |
| `NODE_ENV=test`, no data dir | refuses, exit non-zero, message names the variable, does not leak the path |
| `VITEST` set, no data dir | refuses, exit non-zero |
| `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR=1` outside a runner | refuses, exit non-zero |
| test-shaped + explicit data dir | proceeds, resolves the explicit dir |
| test-shaped + escape hatch | proceeds, resolves the user-level dir |
| **non-test, no data dir** | **still hydrates from the user-level config — the regression guard** |
| test-shaped, user config names no data dir | proceeds, resolves `projectRoot/data` |

**Effect verification against the real built CLI binary** (not just the config module), same synthetic `HOME`:

```
[real CLI, test-shaped (NODE_ENV=test)] exit=1  stderr[0]=[neotoma] Refusing to resolve the data directory from the user-level config while running under test.
[real CLI, non-test (regression guard)] exit=0  "data_dir": "<stand-in dir>", "data_dir_source": "env"
```

### Planted failure — proof the guard can go red

Removed the two-line guard from `hydrateDataDirFromUserEnvConfig()`, rebuilt, re-ran:

```
× refuses and exits non-zero when NODE_ENV=test and no data dir is set
× refuses and exits non-zero when VITEST is set and no data dir is set
× refuses when NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR=1 outside a test runner
✓ proceeds when a test-shaped process sets an explicit data dir
✓ proceeds via the escape hatch when a test opts in deliberately
✓ still hydrates from the user-level config in a non-test process
✓ falls back to projectRoot/data under test when the user config names no data dir
Test Files  1 failed (1)
```

Exactly the three refusal cases went red; the four "proceeds" cases, including the regression guard, stayed green. Guard restored, all 7 green again.

### Cross-surface parity

Full suite on this branch: **540 test files passed, 7 skipped, 0 failed.** `tsc --noEmit` clean. The MCP/HTTP startup path is unaffected — it is a non-test process, and the 540-file run includes the integration and contract suites that boot the server.

### One existing test the guard caught in the act

`tests/cli/cli_init_interactive.test.ts` → "prefers NEOTOMA_DATA_DIR from selected env file" went red on the first full CLI run: it builds a synthetic `HOME` with a user-level config naming a data directory and then imports the CLI in-process, so it hit the refusal. That is the defect's own pattern — a test relying on the user-level fallback — caught by the guard rather than by a reviewer. It is a deliberate exercise of the fallback against a synthetic home holding no real data, so it now opts in via the escape hatch, scoped with try/finally because that file has no env-cleanup hook. `tests/unit/config_data_dir_resolution.test.ts` needed the same opt-in for the same reason. Those two are the only occurrences in 547 test files.

## On #1504 and #172 — checked, and neither is a live sibling

The issue suggested these might be one family. Read both; the family framing is right about the shape but both instances are already closed out, so there is nothing shared and live to fix:

- **#1504** (`NEOTOMA_SQLITE_PATH` advertised but ignored) — **already fixed on `main`.** `src/config.ts` now reads `process.env.NEOTOMA_SQLITE_PATH` before the derived default, and `tests/unit/config_sqlite_path.test.ts` covers both the override-set and override-unset paths. The issue is still open but labelled `resolved-in-release`; it looks like a bookkeeping close rather than work.
- **#172** (`upload --local` defaulting toward the wrong directory) — **CLOSED**, with coverage in `tests/unit/cli_bug_fixes.test.ts`.

So the accurate statement is that all three belong to one family — *config resolution that silently picks a surprising directory instead of refusing* — and this PR closes the last open member of it. No scope was expanded; neither issue's code was touched.

## Not done here

Per the issue's scope: no residue was audited or removed from any real database, and nothing was deleted anywhere. Extending the refusal to non-test CLI paths remains the open follow-up question #2387 raises and is deliberately out of scope. The issue's second defect — auditing every CLI-invoking test for implicit inheritance — is now partly mechanical rather than manual: the guard fails any such test loudly, and the full-suite run above found exactly two, both addressed.

### DIFF
--- a/docs/operations/configuration.md
+++ b/docs/operations/configuration.md
@@ -18,25 +18,37 @@ All state lives under one directory you control, set by `NEOTOMA_DATA_DIR` (defa
 
 Resolution order for the data directory and variables: a project-local `.env`, then `~/.config/neotoma/.env`, then built-in defaults.
 
+### Under test, the user-level step is refused
+
+A test-shaped process — one with `VITEST` set, `NODE_ENV=test`, or `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR=1` — does **not** fall back to `~/.config/neotoma/.env` for its data directory. If it reaches that step it stops and exits non-zero, naming the variable to set.
+
+The reason is that the user-level config names the data directory holding real data. A test process, or a CLI child a test spawns without passing `NEOTOMA_DATA_DIR` through, would otherwise read and write that directory silently and succeed — so the mistake is invisible, and a test can pass on residue left in real data rather than on the code under test.
+
+So set `NEOTOMA_DATA_DIR` to a test-scoped directory, and pass it explicitly to every CLI child process a test spawns rather than relying on inheritance. To exercise the user-level fallback deliberately, set `NEOTOMA_ALLOW_USER_ENV_IN_TEST=1`.
+
+Nothing changes for normal operation: the interactive CLI, the server, and every non-test entry point resolve the data directory exactly as the order above describes.
+
 ## Environments
 
 `NEOTOMA_ENV` selects the profile: `development` (default) or `production`. The profiles use separate database files, source directories, and logs so a dev stack never touches prod data. Production also changes default ports and tightens auth expectations.
 
 ## Core variables
 
-| Variable | Purpose | Default |
-| --- | --- | --- |
-| `NEOTOMA_ENV` | `development` or `production` | `development` |
-| `NEOTOMA_DATA_DIR` | Root data directory | local `data/` |
-| `NEOTOMA_SQLITE_PATH` | Explicit database file path | `{dataDir}/neotoma.db` (dev) |
-| `NEOTOMA_DB_BACKEND` | DB driver: `sqlite` (synchronous, zero-config) or `libsql` (concurrent — statements run off the event loop via worker-hosted driver for local files, or @libsql/client for remote sqld/Turso, so slow queries can't freeze the server; recommended for hosted/agent-heavy/shared instances) | `sqlite` |
-| `NEOTOMA_DB_URL` | libsql connection URL (`file:` for embedded local, `http(s)://`/`libsql://` for remote sqld/Turso) | `file:{NEOTOMA_SQLITE_PATH}` |
-| `NEOTOMA_DB_AUTH_TOKEN` | Auth token for remote libsql connections | unset |
-| `NEOTOMA_DB_READER_WORKERS` | Read-only worker connections for the local `libsql` backend (WAL lets them run concurrently with the writer). Reads go to the least-loaded reader, so raising this adds capacity rather than just slots | `2` |
-| `NEOTOMA_DB_STATEMENT_TIMEOUT_MS` | Per-statement budget for reads on that reader pool. On expiry the reader is terminated (the synchronous driver cannot be interrupted mid-statement) and respawns on the next read, so one runaway query cannot hold a pool slot indefinitely. Writes and statements inside a transaction are never timed out. `0` disables | `30000` |
-| `NEOTOMA_RAW_STORAGE_DIR` | Content-addressed source files | `{dataDir}/sources` |
-| `NEOTOMA_LOGS_DIR` / `NEOTOMA_EVENT_LOG_PATH` | Log directory and event log file | under `{dataDir}/logs` |
-| `NEOTOMA_HOST_URL` / `NEOTOMA_PUBLIC_BASE_URL` | Public URL of this instance | auto-discovered or unset |
+| Variable                                       | Purpose                                                                                                                                                                                                                                                                                                                    | Default                      |
+| ---------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------- |
+| `NEOTOMA_ENV`                                  | `development` or `production`                                                                                                                                                                                                                                                                                              | `development`                |
+| `NEOTOMA_DATA_DIR`                             | Root data directory                                                                                                                                                                                                                                                                                                        | local `data/`                |
+| `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR`            | `1` demands an explicit `NEOTOMA_DATA_DIR`: refuse the `~/.config/neotoma/.env` fallback and exit non-zero. Implied by `VITEST` / `NODE_ENV=test`                                                                                                                                                                          | unset                        |
+| `NEOTOMA_ALLOW_USER_ENV_IN_TEST`               | `1` re-permits that fallback in a test-shaped process, for a test that means to exercise it                                                                                                                                                                                                                                | unset                        |
+| `NEOTOMA_SQLITE_PATH`                          | Explicit database file path                                                                                                                                                                                                                                                                                                | `{dataDir}/neotoma.db` (dev) |
+| `NEOTOMA_DB_BACKEND`                           | DB driver: `sqlite` (synchronous, zero-config) or `libsql` (concurrent — statements run off the event loop via worker-hosted driver for local files, or @libsql/client for remote sqld/Turso, so slow queries can't freeze the server; recommended for hosted/agent-heavy/shared instances)                                | `sqlite`                     |
+| `NEOTOMA_DB_URL`                               | libsql connection URL (`file:` for embedded local, `http(s)://`/`libsql://` for remote sqld/Turso)                                                                                                                                                                                                                         | `file:{NEOTOMA_SQLITE_PATH}` |
+| `NEOTOMA_DB_AUTH_TOKEN`                        | Auth token for remote libsql connections                                                                                                                                                                                                                                                                                   | unset                        |
+| `NEOTOMA_DB_READER_WORKERS`                    | Read-only worker connections for the local `libsql` backend (WAL lets them run concurrently with the writer). Reads go to the least-loaded reader, so raising this adds capacity rather than just slots                                                                                                                    | `2`                          |
+| `NEOTOMA_DB_STATEMENT_TIMEOUT_MS`              | Per-statement budget for reads on that reader pool. On expiry the reader is terminated (the synchronous driver cannot be interrupted mid-statement) and respawns on the next read, so one runaway query cannot hold a pool slot indefinitely. Writes and statements inside a transaction are never timed out. `0` disables | `30000`                      |
+| `NEOTOMA_RAW_STORAGE_DIR`                      | Content-addressed source files                                                                                                                                                                                                                                                                                             | `{dataDir}/sources`          |
+| `NEOTOMA_LOGS_DIR` / `NEOTOMA_EVENT_LOG_PATH`  | Log directory and event log file                                                                                                                                                                                                                                                                                           | under `{dataDir}/logs`       |
+| `NEOTOMA_HOST_URL` / `NEOTOMA_PUBLIC_BASE_URL` | Public URL of this instance                                                                                                                                                                                                                                                                                                | auto-discovered or unset     |
 
 ### When to opt into `NEOTOMA_DB_BACKEND=libsql`
 
@@ -49,22 +61,22 @@ Before flipping the variable on an existing database, run `npx tsx scripts/valid
 
 ## Server and ports
 
-| Variable | Purpose | Default |
-| --- | --- | --- |
+| Variable                             | Purpose                    | Default                 |
+| ------------------------------------ | -------------------------- | ----------------------- |
 | `NEOTOMA_HTTP_PORT` (or `HTTP_PORT`) | HTTP API and HTTP MCP port | `3080` dev, `3180` prod |
-| `WS_PORT` | WebSocket MCP bridge port | `8280` |
+| `WS_PORT`                            | WebSocket MCP bridge port  | `8280`                  |
 
 See [Running the Server](running_the_server.md) for transports and processes.
 
 ## Auth and access
 
-| Variable | Purpose |
-| --- | --- |
-| `NEOTOMA_REQUIRE_KEY_FOR_OAUTH` | Require a key for OAuth connections |
-| `NEOTOMA_OAUTH_CLIENT_ID` | MCP OAuth client id (hosted mode) |
+| Variable                              | Purpose                                                                                                                                                                     |
+| ------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `NEOTOMA_REQUIRE_KEY_FOR_OAUTH`       | Require a key for OAuth connections                                                                                                                                         |
+| `NEOTOMA_OAUTH_CLIENT_ID`             | MCP OAuth client id (hosted mode)                                                                                                                                           |
 | `NEOTOMA_OAUTH_TRUSTED_CALLBACK_URLS` | Comma-separated **exact** callback URLs additionally trusted to receive an authorization code when the authorize request arrives via a tunnel. Empty by default. See below. |
-| `NEOTOMA_SANDBOX_MODE` | Opt into the public hosted-sandbox profile |
-| `NEOTOMA_REFUSE_MODE` | `warn` or `enforce` when a no-auth, non-loopback topology is detected |
+| `NEOTOMA_SANDBOX_MODE`                | Opt into the public hosted-sandbox profile                                                                                                                                  |
+| `NEOTOMA_REFUSE_MODE`                 | `warn` or `enforce` when a no-auth, non-loopback topology is detected                                                                                                       |
 
 ### Trusted OAuth callback URLs
 
@@ -86,14 +98,14 @@ Several entries are comma-separated. Rules worth knowing before you set it:
 
 A redirect is accepted when the request's callback and a configured entry agree on **scheme, host, port and path** — all four. In detail:
 
-| Part | How it is compared |
-| --- | --- |
-| Scheme | Must match. `https` does not authorise its `http` twin. |
-| Host | Case-**insensitive**. `APP.EXAMPLE.COM` matches `app.example.com`. |
-| Port | Must match, but a default port is equivalent to none: `https://app.example.com` = `https://app.example.com:443`. |
-| Path | Case-**sensitive**. `/Auth/Callback` does **not** match `/auth/callback`. A trailing slash is insignificant in either direction. Dot segments resolve first, so `/auth/callback/../admin` is compared as `/auth/admin`. An encoded slash (`%2F`) is not a path separator. |
-| Query and fragment | **Ignored on both sides**, so the usual `?code=...&state=...` callback matches. |
-| Userinfo | Any URL carrying `user:password@` is rejected. |
+| Part               | How it is compared                                                                                                                                                                                                                                                        |
+| ------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Scheme             | Must match. `https` does not authorise its `http` twin.                                                                                                                                                                                                                   |
+| Host               | Case-**insensitive**. `APP.EXAMPLE.COM` matches `app.example.com`.                                                                                                                                                                                                        |
+| Port               | Must match, but a default port is equivalent to none: `https://app.example.com` = `https://app.example.com:443`.                                                                                                                                                          |
+| Path               | Case-**sensitive**. `/Auth/Callback` does **not** match `/auth/callback`. A trailing slash is insignificant in either direction. Dot segments resolve first, so `/auth/callback/../admin` is compared as `/auth/admin`. An encoded slash (`%2F`) is not a path separator. |
+| Query and fragment | **Ignored on both sides**, so the usual `?code=...&state=...` callback matches.                                                                                                                                                                                           |
+| Userinfo           | Any URL carrying `user:password@` is rejected.                                                                                                                                                                                                                            |
 
 If a redirect is refused, the `400` response names the callback URL the server actually compared (scheme, host and path — query and fragment are stripped) and says how many entries are configured. That is usually enough to tell a near miss — a trailing slash, a port, a case difference — from a typo or an unset variable, without needing server logs.
 
@@ -111,23 +123,23 @@ See [Deployment Modes](deployment.md) and [Agent Access Control](agent_access_co
 
 ## Encryption
 
-| Variable | Purpose |
-| --- | --- |
-| `NEOTOMA_ENCRYPTION_ENABLED` | Turn on AES-256-GCM at-rest column encryption |
-| `NEOTOMA_KEY_FILE_PATH` | Path to a 32-byte key file |
-| `NEOTOMA_MNEMONIC` / `NEOTOMA_MNEMONIC_PASSPHRASE` | BIP-39 mnemonic key source |
-| `NEOTOMA_LOG_ENCRYPTION_ENABLED` | Encrypt the event log |
-| `NEOTOMA_MCP_TOKEN_ENCRYPTION_KEY` | Encrypt stored MCP OAuth tokens |
+| Variable                                           | Purpose                                       |
+| -------------------------------------------------- | --------------------------------------------- |
+| `NEOTOMA_ENCRYPTION_ENABLED`                       | Turn on AES-256-GCM at-rest column encryption |
+| `NEOTOMA_KEY_FILE_PATH`                            | Path to a 32-byte key file                    |
+| `NEOTOMA_MNEMONIC` / `NEOTOMA_MNEMONIC_PASSPHRASE` | BIP-39 mnemonic key source                    |
+| `NEOTOMA_LOG_ENCRYPTION_ENABLED`                   | Encrypt the event log                         |
+| `NEOTOMA_MCP_TOKEN_ENCRYPTION_KEY`                 | Encrypt stored MCP OAuth tokens               |
 
 See [Encryption and Key Management](encryption.md).
 
 ## Search, inspector, and docs
 
-| Variable | Purpose |
-| --- | --- |
-| `OPENAI_API_KEY` | Enables semantic vector search (embeddings); keyword search works without it |
-| `NEOTOMA_INSPECTOR_DISABLE` / `NEOTOMA_PUBLIC_INSPECTOR_URL` / `NEOTOMA_INSPECTOR_BASE_PATH` | Control the bundled Inspector |
-| `NEOTOMA_DOCS_SHOW_INTERNAL` | Show `visibility: internal` docs in the in-app `/docs` browser |
+| Variable                                                                                     | Purpose                                                                      |
+| -------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------- |
+| `OPENAI_API_KEY`                                                                             | Enables semantic vector search (embeddings); keyword search works without it |
+| `NEOTOMA_INSPECTOR_DISABLE` / `NEOTOMA_PUBLIC_INSPECTOR_URL` / `NEOTOMA_INSPECTOR_BASE_PATH` | Control the bundled Inspector                                                |
+| `NEOTOMA_DOCS_SHOW_INTERNAL`                                                                 | Show `visibility: internal` docs in the in-app `/docs` browser               |
 
 ## Limits and mirror
 
--- a/docs/testing/automated_test_catalog.md
+++ b/docs/testing/automated_test_catalog.md
@@ -61,15 +61,15 @@ flowchart TD
 - Do not hand-edit suite inventory entries in this file. Update the generator or the repository tree, then regenerate.
 
 ## Repo-wide summary
-- Total automated test files: **605**
-- Backend and repo Vitest files: **570**
+- Total automated test files: **606**
+- Backend and repo Vitest files: **571**
 - Frontend Vitest files: **9**
 - Playwright spec files: **26**
 
 ### Suite counts
 | Suite | Files |
 |---|---:|
-| Vitest unit tests | 166 |
+| Vitest unit tests | 167 |
 | Vitest service tests | 44 |
 | Source-adjacent tests | 66 |
 | Vitest integration tests | 172 |
@@ -109,7 +109,7 @@ flowchart TD
 **Runner:** `vitest`
 **Command:** `npm test -- tests/unit`
 **Requirements:** Basic `.env` if required by the module under test.
-**Files (166):**
+**Files (167):**
 - `tests/unit/aauth_admission.test.ts`
 - `tests/unit/aauth_attestation_apple_se.test.ts`
 - `tests/unit/aauth_attestation_revocation.test.ts`
@@ -151,6 +151,7 @@ flowchart TD
 - `tests/unit/compliance_scorecard.test.ts`
 - `tests/unit/config_data_dir_resolution.test.ts`
 - `tests/unit/config_db_backend.test.ts`
+- `tests/unit/config_refuses_user_env_under_test.test.ts`
 - `tests/unit/config_sqlite_path.test.ts`
 - `tests/unit/config_trusted_callbacks.test.ts`
 - `tests/unit/content_field_store_warning.test.ts`
--- a/src/config.ts
+++ b/src/config.ts
@@ -19,17 +19,76 @@ function resolveUserEnvPath(): string | null {
   return join(homeDir, ".config", "neotoma", ".env");
 }
 
+/**
+ * Is this process test-shaped (issue #2387)?
+ *
+ * `VITEST` is set by the vitest runner in workers; `NODE_ENV=test` is set by
+ * `vitest.global_setup.ts` and is the conventional signal a child process
+ * inherits. `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR=1` lets any caller demand the
+ * same strictness without pretending to be a test.
+ */
+function requiresExplicitDataDir(): boolean {
+  if (process.env.NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR?.trim() === "1") return true;
+  if (process.env.VITEST?.trim()) return true;
+  return process.env.NODE_ENV?.trim() === "test";
+}
+
+/** Deliberate opt-in that re-permits the user-level fallback under test (issue #2387). */
+function userEnvFallbackAllowedInTest(): boolean {
+  return process.env.NEOTOMA_ALLOW_USER_ENV_IN_TEST?.trim() === "1";
+}
+
+/**
+ * Refuse the `~/.config/neotoma/.env` data-directory fallback in a test-shaped
+ * process (issue #2387).
+ *
+ * A CLI child spawned by a test does not necessarily inherit `NEOTOMA_DATA_DIR`.
+ * Without this guard such a child hydrated the variable from the user-level
+ * config and then read and wrote the data directory it names — the operator's
+ * real database — silently and with exit code 0. A test that passes against
+ * residue in real data is green for the wrong reason twice over, and the write
+ * itself is invisible because nothing fails.
+ *
+ * So: stop, name the variable to set, and exit non-zero. Non-test processes are
+ * untouched — the interactive CLI and the server keep hydrating exactly as
+ * before — and `NEOTOMA_ALLOW_USER_ENV_IN_TEST=1` re-permits it for the rare
+ * test that genuinely means to exercise the fallback.
+ */
+function refuseUserEnvFallbackUnderTest(userEnvPath: string, configuredDataDir: string): never {
+  const message = [
+    "[neotoma] Refusing to resolve the data directory from the user-level config while running under test.",
+    `  A test-shaped process (VITEST / NODE_ENV=test / NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR=1) has no explicit NEOTOMA_DATA_DIR,`,
+    `  so ${userEnvPath} would have pointed it at a data directory holding real data.`,
+    "",
+    "  Set NEOTOMA_DATA_DIR to a test-scoped directory in the environment of this process",
+    "  (and pass it explicitly to any CLI child process the test spawns).",
+    "  To exercise the user-level fallback on purpose, set NEOTOMA_ALLOW_USER_ENV_IN_TEST=1.",
+  ].join("\n");
+  process.stderr.write(`${message}\n`);
+  // Length, not content: the configured path may name a real operator directory.
+  process.stderr.write(
+    `[neotoma] (refused user-level data directory: ${configuredDataDir.length} chars, not used)\n`
+  );
+  process.exit(1);
+}
+
 function hydrateDataDirFromUserEnvConfig(): void {
   if (process.env.NEOTOMA_DATA_DIR?.trim()) return;
   const userEnvPath = resolveUserEnvPath();
   if (!userEnvPath || !existsSync(userEnvPath)) return;
+  let configuredDataDir: string | undefined;
   try {
     const parsed = dotenv.parse(readFileSync(userEnvPath, "utf-8"));
-    const configuredDataDir = parsed.NEOTOMA_DATA_DIR?.trim();
-    if (configuredDataDir) process.env.NEOTOMA_DATA_DIR = configuredDataDir;
+    configuredDataDir = parsed.NEOTOMA_DATA_DIR?.trim();
   } catch {
     // Ignore invalid user env files and continue with existing fallbacks.
+    return;
   }
+  if (!configuredDataDir) return;
+  if (requiresExplicitDataDir() && !userEnvFallbackAllowedInTest()) {
+    refuseUserEnvFallbackUnderTest(userEnvPath, configuredDataDir);
+  }
+  process.env.NEOTOMA_DATA_DIR = configuredDataDir;
 }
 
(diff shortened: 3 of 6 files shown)


neotoma-agent:
**COMMENT**

## Triage & Gate-Status Board — parent issue #2387

**Gate status** (parent issue #2387)

| Gate | State | Note |
|---|---|---|
| pm | not_evaluated (legacy) | No issue entity existed in Neotoma for #2387 at PR-open time |
| ux | not_applicable | `bug` fast path skips ux |
| arch | not_evaluated (legacy) | No issue entity existed in Neotoma for #2387 at PR-open time |
| impl | not_evaluated (legacy) | Cicada's role; PR itself is the implementation artifact |
| pr_review | pending | Assigning now |
| qa | pending | Phase 4b |
| legal | not_required | No legal_required flag applicable |

**Legacy-issue handling**: retrieved `issue` entity by `github_number=2387` (scoped to `markmhendrickson/neotoma`) via both `retrieve_entity_by_identifier` and `retrieve_entities` (including merged entities) — no match. `gate_status` was genuinely absent, not merely pending, so per the legacy-issue rule this does not hard-block. I initialized the gate keys retroactively on a newly created issue entity (`ent_df006a561020f451d58a5c39`), noted the legacy status in `owner_history`, and treat pre-impl gates (pm/ux/arch) as satisfied for inheritance purposes.

**Owner**: current phase owner → **Vanellus** (pr_review), assigned as reviewer on this PR. Next phase: Phoenicurus (qa).

**Labels**: none applied by Lanius on this PR event (label application is scoped to issue-triage, not PR-open).

📎 Neotoma: issue ent_df006a561020f451d58a5c39

GATE_INHERITANCE: clear

markmhendrickson:
Verified this independently before relaying it, and there is one thing to correct plus one place where I was wrong in the brief that produced it.

**I was wrong to name #1504 and #172 as live siblings.** My dispatch brief asked whether this shared a root cause with both. The agent checked instead of accepting it, and it was right: **#1504 is labelled `resolved-in-release` and `NEOTOMA_SQLITE_PATH` is in fact read at `src/config.ts:272`** — I confirmed both — so the defect it reports is already fixed and the issue is open as bookkeeping. **#172 is CLOSED.** The accurate framing is the one the PR body uses: all three share one shape — config resolution silently picking a surprising directory instead of refusing — and this closes the last open member. Pushing back on a brief's premise with evidence is the right behaviour, and it avoided a PR that claimed to be part of a live family that isn't one.

**A small correction to the disclosure claim.** The refusal message deliberately prints the *length* of the refused data directory rather than its content, with the comment `// Length, not content: the configured path may name a real operator directory.` That is a good call and the reasoning is right — CI stderr is public.

But the message on the line above does print `userEnvPath`, which is built from `$HOME` and therefore contains a username. So the accurate statement is not "no paths in the output" — it is that the **operator-specific data directory** is withheld while the **fixed config path** is shown, because the second is needed for the message to be actionable and the first is not. I think that line is defensible exactly where it is drawn; I am noting it so nobody later reads the comment as a guarantee that the refusal output is free of identifying paths. If CI-log hygiene is meant to be stricter than that, this is the one line to change.

## What I verified

- Six files: `src/config.ts`, three test files, two docs. No `src/actions.ts`, no `src/server.ts`, no change to PR #1942's test file — scope held.
- `Design basis:` present; `Closes #2387` present.
- **Zero occurrences of this machine's absolute paths anywhere in the diff.**
- The new env vars are documented in `docs/operations/configuration.md` alongside the existing ones, not only in a code comment.
- The refusal fires only where the fallback would actually have been taken — test-shaped, no explicit dir, *and* a user config that names one — which is what leaves the no-user-config case resolving to `projectRoot/data` as before.

## The result that matters most

The guard **caught a live instance of its own defect on the first full run**: `tests/cli/cli_init_interactive.test.ts` → "prefers NEOTOMA_DATA_DIR from selected env file" went red, because it builds a synthetic home whose user config names a data directory and then imports the CLI in-process. That is the defect's exact pattern, found mechanically rather than by review. Two such tests exist across 547 test files, and both now opt in explicitly via the escape hatch.

That is the part I would not have got from reading: a guard that finds two real occurrences the moment it is switched on is evidence it binds, in a way a passing test suite is not. Planted failure confirms the other direction — removing the two-line guard turns exactly the three refusal cases red and leaves the four "proceeds" cases, including the non-test regression guard, green.

## Sequencing, restated because it is the reason this was dispatched now

This closes the exposure **prospectively only**. Today the fallback is harmless by accident: PR #1942's unfixed defect means `store()` fails for every `skill` entity, so probes cannot land in real data — I confirmed zero `skill`, `product_feedback`, and `reference_note` rows. **The moment #1942 merges, that accident ends.** So this is worth landing before or alongside it, not after.

It also does not clean anything already present, and I have not asked it to. Whether to audit for residue from earlier runs is the operator's call, and neither this PR nor I will touch real data without explicit direction.

neotoma-agent:
review:pm

**COMMENT**

Design basis: PR body states "No `docs/foundation/` design governs this diff" with a reason (runtime safety guard, not a work-model/gate/principle change) — valid per the mechanical check.

Reviewed against the pm-signed scope of record on #2387 (my own sign-off comment, gate_status.pm → signed_off): in-scope was a fail-loud refusal branch in `config.ts` under a test-mode signal, plus an audit of remaining CLI-invoking tests for the same implicit-inheritance pattern; out-of-scope was residue cleanup, non-test CLI paths, and #1504/#172.

**Scope match — clean:**
- `requiresExplicitDataDir()` / `userEnvFallbackAllowedInTest()` / `refuseUserEnvFallbackUnderTest()` in `src/config.ts` (lines ~203–246) implement exactly the refusal branch scoped, gated on `VITEST` / `NODE_ENV=test` / `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR=1` — matching the three signals the issue and my sign-off named.
- The audit item is discharged: PR body reports a full-suite run (540/547 files) found exactly two tests relying on the implicit fallback (`cli_init_interactive.test.ts`, `config_data_dir_resolution.test.ts`), both fixed with an explicit `NEOTOMA_ALLOW_USER_ENV_IN_TEST=1` opt-in and comments explaining why.
- Non-test paths (`src/actions.ts`, `src/server.ts`) are explicitly untouched, matching the out-of-scope boundary.
- #1504 and #172 are correctly left alone — PR body states neither's code was touched, consistent with "family, not one fix."
- No unrequested scope creep: the `docs/operations/configuration.md` table reformatting is mechanical (markdown-table column alignment, likely from a formatter) riding along with the necessary new-row additions for the two new env vars — not new scope, just noisy diff.

**Quality gate (my standing pm requirement) — met:**
- Effect-verified: PR body shows revert-to-red (guard removed → exactly the 3 refusal-case tests go red, 4 proceed-cases including the regression guard stay green), and the built CLI binary itself was exercised (not just the config module) with real exit codes and stderr captured.
- Cross-surface parity: this capability has one surface (CLI/config module) — no MCP tool, REST route, or SDK method exposes data-dir resolution, so N/A rather than skipped; the PR body's note that the 540-file run "includes the integration and contract suites that boot the server" is the relevant regression evidence for the non-test path staying untouched.

[NON-BLOCKING] ux-checklist-gap: `--help` output does not document `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR` / `NEOTOMA_ALLOW_USER_ENV_IN_TEST`.
The ux lens I normalized onto #2387's swarm spec listed "`--help` output documents `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR`" as an acceptance item. The diff touches `docs/operations/configuration.md` (good — new table rows for both vars) but no CLI help-text source file. This isn't a pm-scope violation (the pm-signed acceptance criteria didn't require it, only the ux checklist did) and it's unverified beyond "not in the diff" — I have no CLI checkout to confirm `--help` doesn't already surface env vars generically. Confirm by running `neotoma --help` (or the equivalent flag-help command) on this branch and checking whether env-var docs appear there at all today; if they don't for any existing var, this is consistent with prior practice and not a gap specific to this PR.

**Verdict: COMMENT.** Scope and acceptance criteria match what pm signed off; user-visible behavior (new stderr refusal message, unchanged non-test resolution, two new documented env vars) matches the issue's ask. No blocking pm-lens findings.

📎 Neotoma: see plan_contribution `ent_d24dc0dbe49e65cf2543552d` (pm sign-off on #2387)

neotoma-agent:
review:ux
**COMMENT**

Reviewing diff-only (no checkout) through the ux lens: discoverable naming, actionable error messages, docs/examples for the new surface. Not a correctness/security pass — that's the GHA baseline.

**User goal**: a developer whose test process just exited 1 needs to know, from terminal output alone, why — and the one action that fixes it — without opening `src/config.ts`.

## Walkthrough of the new surface

1. Test-shaped process reaches the fallback → `refuseUserEnvFallbackUnderTest()` fires two `stderr.write` calls → `process.exit(1)`.
2. Message names all three trigger signals (`VITEST` / `NODE_ENV=test` / `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR=1`), names the fix (`NEOTOMA_DATA_DIR`), and names the deliberate escape hatch (`NEOTOMA_ALLOW_USER_ENV_IN_TEST=1`) in the same message. This is a fail-loud error that carries its own remedy inline — meets the actionable-hint bar; a developer does not need to read source to resolve it.
3. Second `stderr.write` discloses refused-path *length* only, with an inline comment explaining why (`src/config.ts:243`) — reasonable non-leaky disclosure for a CI-log-visible message.
4. Both new env vars (`NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR`, `NEOTOMA_ALLOW_USER_ENV_IN_TEST`) land in `docs/operations/configuration.md`'s existing variable table, in the same place and format as every other `NEOTOMA_*` var, plus a new prose subsection ("Under test, the user-level step is refused") explaining the *why*, not just the *what*. This is the right IA move — it extends the existing reference surface rather than creating a separate doc, satisfying the "extend what already generalizes" pattern this repo asks for.

Naming: `NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR` and `NEOTOMA_ALLOW_USER_ENV_IN_TEST` both read correctly on sight — verb-first, state what they do, consistent with the existing `NEOTOMA_*` convention (`NEOTOMA_SANDBOX_MODE`, `NEOTOMA_REFUSE_MODE`, etc.). No naming complaint.

## Findings

[NON-BLOCKING] docs-completeness: no worked example of the actual failure/fix in `configuration.md`.
The new prose section (`docs/operations/configuration.md` lines 21–19 of the diff) states the remedy in words ("set `NEOTOMA_DATA_DIR` to a test-scoped directory... pass it explicitly to every CLI child process") but doesn't show a copy-pasteable line (e.g. `NEOTOMA_DATA_DIR=$(mktemp -d) npm test`) or the literal stderr text a developer will see. The stated user profile for this repo's audience is a senior engineer who "will abandon a tool if it requires more mental overhead than the alternative" — the message itself is already actionable enough that this is a polish item, not a blocker. Confirm by reading the rendered doc section as a first-time reader unfamiliar with the guard.

[NON-BLOCKING] discoverability-gap (concur with Pavo's pm-lens comment, unverified): `--help` output is not touched by this diff.
If `neotoma --help` or an equivalent CLI help surface enumerates env vars anywhere today, these two new ones should appear there too — that's the more likely place a developer looks first, before the docs file. This is unverified against a checkout (I have diff-only access) and, per Pavo's comment, may be consistent with how no other env var is surfaced in `--help` today, in which case this isn't a gap specific to this PR. Confirm by running `neotoma --help` on this branch and checking whether any `NEOTOMA_*` var appears there.

No blocking ux findings. The refusal message is self-contained and actionable (names the trigger, the fix, and the escape hatch in one place); the docs change extends the existing reference table rather than forking a new one, which is the correct IA move for this surface.

**Verdict: COMMENT.** No ux-blocking defects in the diff. Two non-blocking polish items noted above, both unverified against a live checkout per the evidence bar for this review.

📎 Neotoma: PR https://github.com/markmhendrickson/neotoma/pull/2388 · issue #2387

neotoma-agent:
review:qa

**APPROVE**

Scope: test coverage adequacy only (regression test for the fixed bug, edge cases on the new branch, contract impact). Correctness/security is the Claude GHA's lane.

## Verified (ran locally on the PR branch, `pr-2388` fetched from `refs/pull/2388/head`, built via `npm run build:server`)

- **New suite `tests/unit/config_refuses_user_env_under_test.test.ts` (7 cases) — green as claimed.**
  ```
  npx vitest run tests/unit/config_refuses_user_env_under_test.test.ts
  ✓ tests/unit/config_refuses_user_env_under_test.test.ts (7 tests) 198ms
  ```

- **Revert-to-red confirmed** (`docs/foundation/principles.md` invariant 4 — a test that cannot fail on the thing it watches is decoration). I removed the two-line guard call from `hydrateDataDirFromUserEnvConfig()`, rebuilt, reran:
  ```
  FAIL … refuses and exits non-zero when NODE_ENV=test and no data dir is set
  FAIL … refuses and exits non-zero when VITEST is set and no data dir is set
  FAIL … refuses when NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR=1 outside a test runner
  Test Files  1 failed (1)
       Tests  3 failed | 4 passed (7)
  ```
  Exactly the 3 refusal cases went red; the 4 "proceeds" cases, including the non-test regression guard, stayed green — matching the PR body's own reported revert result exactly. Restored the guard, reconfirmed 7/7 green.

- **The `NEOTOMA_ALLOW_USER_ENV_IN_TEST` opt-in added to `tests/cli/cli_init_interactive.test.ts` is load-bearing, not cargo-culted.** I initially suspected it was unnecessary, since the test calls `resolveInitDataDirDefaults()` — a distinct code path in `src/cli/index.ts` that never calls `requiresExplicitDataDir()`/`refuseUserEnvFallbackUnderTest()` directly. I stripped the opt-in and reran to check:
  ```
  FAIL … prefers NEOTOMA_DATA_DIR from selected env file
  Error: process.exit unexpectedly called with "1"
   ❯ refuseUserEnvFallbackUnderTest src/config.ts:72:11
   ❯ hydrateDataDirFromUserEnvConfig src/config.ts:89:5
   ❯ src/config.ts:153:1
   ❯ src/cli/index.ts:25:1
  ```
  The guard fires as a **module-load side effect** when `src/cli/index.ts` imports `src/config.ts` (line 153: `hydrateDataDirFromUserEnvConfig();` runs unconditionally at import time), independent of which CLI function the test calls. So the opt-in is genuinely required, not defensive boilerplate. Restored the wrapper; suite back to 13/13 green.

- **`tests/unit/config_data_dir_resolution.test.ts`** opt-in via `vi.stubEnv("NEOTOMA_ALLOW_USER_ENV_IN_TEST", "1")` is correctly scoped to the one case that deliberately exercises the fallback; ran green alongside the above.

- **CI wiring confirmed**: `test:unit` (which includes `tests/unit` and `tests/cli`) runs on the PR baseline lane per `.github/workflows/ci_test_lanes.yml:78`, so both the new suite and the two modified suites are exercised in CI, not just locally.

- **Doc accuracy**: `docs/operations/configuration.md`'s new "Under test, the user-level step is refused" section states the three trigger signals, the escape hatch, and the "nothing changes for non-test paths" claim — all of which match the verified code in `src/config.ts`.

## Noise ruled out

Running the broader `tests/unit tests/cli` surface locally showed 35 failures across 6 files (`tests/cli/config_api_discovery.test.ts` and others), all resolving to `https://neotoma.markmhendrickson.com` instead of an expected `http://localhost:PORT`. I checked out the pre-PR base commit (`2fc45a45a`) and reran the same file — **identical failures reproduce there**, confirming this is pre-existing environmental bleed from a real user-level config on this specific checkout, not a regression introduced by this PR. Excluded from the review.

## Edge cases not explicitly covered (non-blocking, no reproduction attempted)

- [NON-BLOCKING] test-coverage: whitespace/casing edge cases on the three trigger env vars (e.g. `NODE_ENV=" test "`, `VITEST=""` vs unset) aren't individually asserted, though `requiresExplicitDataDir()`'s `.trim()` calls suggest they're handled. I did not reproduce a failure here — flagging only as an unexercised corner of the guard's decision surface, matching the existing code's own defensive style elsewhere in the file, so not blocking.

No blocking findings. The regression test is real (proven by revert), the escape-hatch usages are load-bearing (proven by removal), and the new behavior is covered where it fires, including the module-load-time trigger path that a less careful test update would have missed.

---
📎 Neotoma: not applicable — no Neotoma entity created or read for this review; findings are sourced entirely from the PR diff and local test execution.

neotoma-agent:
review:content

**COMMENT**

Reviewing PR #2388 through the content/dogfooding lens only — not re-running correctness/security, that's the GHA baseline.

**Shippable as a content story: yes.** This is exactly the kind of artifact the trust hinge wants — a real defect the operator's own tooling surfaced (a test-shaped process silently hydrating `NEOTOMA_DATA_DIR` from `~/.config/neotoma/.env` and exiting 0 against real data), a guard that's proven to fail red via a planted-failure removal, and a full-suite run backing the claim. Every fact in the PR body traces to something checkable: child-process repro output, the planted-failure table, the 540/7/0 suite result, and the one existing test (`cli_init_interactive`) the guard caught relying on the very fallback being fixed. That's evidence, not marketing — no unverifiable superlatives needed to make it interesting.

**Angle for a post:** "a config default degraded silently until we made silence impossible" — the story isn't the fix, it's that a passing test suite was insufficient proof of correctness because the fallback could succeed for the wrong reason (reading residue in real data). The self-catch of `cli_init_interactive` is the best single beat — it's the pattern catching itself in the act rather than a reviewer finding it.

**Not blocking.** Filed as a task in my queue (`ent_79755008a86380a72e0b7468`, linked to issue #2387) rather than drafting now: per my own discipline I hold on drafting until the PR is merged, so the piece describes shipped state rather than an open PR, and I re-verify the final test counts at that point rather than trusting today's numbers going stale.

One flag for whoever merges: the PR body reports the refused path's length only, not its content, in the error output — worth preserving that framing if any of this language gets pulled into user-facing docs/changelog, since it's a good detail for the "we thought about what NOT to leak" angle without over-claiming a security property this PR doesn't otherwise assert.

📎 Neotoma: issue [neotoma#2387](https://neotoma.markmhendrickson.com/entities/ent_df006a561020f451d58a5c39)

neotoma-agent (approved):

**APPROVE**

Aggregated verdict from panel lenses (pm, ux, qa, content) reviewed against parent issue #2387.

- pm (Pavo): COMMENT — scope matches pm-signed acceptance criteria, quality gate (effect-verified + cross-surface parity, single-surface N/A) satisfied.
- ux (Accipiter): COMMENT — no blocking findings; refusal message is actionable, docs extend existing surface correctly.
- qa (Phoenicurus): APPROVE — independently built and ran the suite; confirmed 7/7 new regression tests green and the exact 3-red/4-green revert-to-red split; empirically verified the escape-hatch usage is load-bearing.
- content (Corvus): COMMENT — non-blocking, filed as a future content task gated on merge.

Blocking: 0

Reviewed commit: 9d5e40e8cb84beeb746344372ede3d52327eb0b4

neotoma-agent:

**APPROVE**

Aggregated review panel verdict for PR #2388, parent issue #2387.

**Per-lens roll-up:**
- **pm** (Pavo): COMMENT — scope matches pm-signed acceptance criteria on #2387 (refusal branch + CLI-test audit; out-of-scope items #1504/#172 correctly untouched). Quality gate (effect-verified via revert-to-red + cross-surface parity, single-surface N/A) satisfied.
- **ux** (Accipiter): COMMENT — no blocking findings. Refusal message names trigger/fix/escape-hatch inline; docs extend the existing `configuration.md` table correctly.
- **qa** (Phoenicurus): APPROVE — independently built and ran the suite (not diff-reading). Verified the new 7-case regression suite passes green and reverts to exactly the claimed 3-red/4-green split on guard removal. Empirically disproved own hypothesis that the escape-hatch usage in one modified test was unnecessary — confirmed it's load-bearing (module-load side effect). Ruled out 35 unrelated local failures as pre-existing environmental noise (reproduces on pre-PR base).
- **content** (Corvus): COMMENT — good build-in-public material (checkable defect story, self-catching test). Filed as a future content task, gated on merge. Not blocking.

**Blocking:** 0. No [BLOCKING] findings from any lens.

**Non-blocking notes carried forward (not gating):**
- `--help` output does not surface the two new env vars (`NEOTOMA_REQUIRE_EXPLICIT_DATA_DIR`, `NEOTOMA_ALLOW_USER_ENV_IN_TEST`) — unverified whether this is a gap specific to this PR or consistent with existing practice (pm + ux concur).
- No worked/copy-pasteable example of the fix command or literal error text in docs (ux).

**Merge-readiness checks (steward):**
- Gate inheritance: pm/arch `not_evaluated_legacy` (Lanius retroactive legacy-init at PR-open time, treated as satisfied for inheritance since no live gate blocked it and PR body documents pm/arch reasoning — independently corroborated by Pavo's pm-lens scope review against #2387); ux `not_applicable`. Cleared.
- Required checks: all CI lanes green, including `security_gates`.
- No formal GitHub review existed pre-aggregation (panel lenses posted as issue comments, not reviews) — `reviewDecision` was `REVIEW_REQUIRED`. Submitted a head-SHA-matched formal APPROVE review (`Reviewed commit: 9d5e40e8cb84beeb746344372ede3d52327eb0b4`) to carry this aggregated verdict as a native review; `reviewDecision` cleared to `APPROVED`, `mergeStateStatus` to `CLEAN`.
- Base is `main` (default branch) — not a stacked PR.

**Verdict: APPROVE.**

**Merged:** squash-merged at `cb55a52cd0877860da878ea16cf44f6966e70f50`. `pr_review` gate signed off on #2387; issue advances to Phase 4b (qa + legal parallel review — legal is `not_required`, so effectively qa-only).

📎 Neotoma: issue [neotoma#2387](https://neotoma.markmhendrickson.com/entities/ent_df006a561020f451d58a5c39)

neotoma-agent (approved):
All storage and relationships complete. Summary for the record:

## PR #2388 — Aggregated verdict: APPROVE, merged

**Panel roll-up:** pm (COMMENT), ux (COMMENT), qa (APPROVE, verified by actually running the suite), content (COMMENT) — zero BLOCKING findings across all four lenses.

**Process defect caught and worked around (not silently absorbed):** The panel lenses had posted their reviews as plain issue comments, not native GitHub reviews. That left `reviewDecision: REVIEW_REQUIRED` and `mergeStateStatus: BLOCKED` despite every CI check being green — a real merge blocker my instructions require me to name and unblock via the documented mechanism (a fresh head-SHA-matched formal review), not merge around silently. I submitted a formal `APPROVE` review carrying the aggregated verdict (`Reviewed commit: 9d5e40e8...`), which cleared it to `APPROVED`/`CLEAN`. This matches the reportable-defect pattern in my brief (a review workflow whose verdicts never reach GitHub's native review state) — it's worth flagging as a systemic gap in the panel dispatcher (lens comments aren't currently submitted as native reviews), but wasn't blocking enough to stop this merge once I applied the workaround, so I completed it and note it here rather than filing a duplicate proposal without checking first.

**Gate inheritance:** pm/arch were `not_evaluated_legacy` (Lanius retroactively legacy-initialized the issue at PR-open time since none existed; the PR body's own pm/arch reasoning, independently corroborated by Pavo's pm-lens scope review, satisfies inheritance intent). ux was `not_applicable`. Treated as cleared.

**Merged:** squash, base `main`, commit `cb55a52cd0877860da878ea16cf44f6966e70f50`.

**Gate handoff:** `pr_review` signed off on issue #2387 (`ent_df006a561020f451d58a5c39`); `current_owner` advanced to `phoenicurus` (qa; legal is `not_required`, so effectively qa-only for Phase 4b).

🧠 Neotoma — [vanellus-pr2388-aggregation](https://neotoma.markmhendrickson.com/inspector/conversations/ent_e824cb6cc8f9a6770620cb7f)
- Updated (1): 🚪 [issue #2387](https://neotoma.markmhendrickson.com/inspector/entities/ent_df006a561020f451d58a5c39) — gate_status.pr_review → signed_off, current_owner → phoenicurus
- Created (1): 📝 [plan_contribution](https://neotoma.markmhendrickson.com/inspector/entities/ent_855621eea39a63bff129a6e1) — vanellus sign-off record
- Retrieved (1): 🚪 issue #2387 (initial gate-status lookup)

`[vanellus] merge_decision: PR #2388 — APPROVE, merged squash @ cb55a52cd0877860da878ea16cf44f6966e70f50. All 4 panel lenses clear (pm/ux/content COMMENT, qa APPROVE via actual test execution), Blocking: 0. Gate inheritance cleared (pm/arch legacy-satisfied, ux not_applicable). Required checks green, base=main, no stacked-PR risk. Submitted missing head-SHA-matched formal GitHub review to clear a pre-existing REVIEW_REQUIRED/BLOCKED state that panel-comment-only reviews had left unresolved. pr_review gate signed off on #2387, owner advanced to phoenicurus.`

